PR #1, opened 2018-06-25
Recommend shuffling the RRSET

bemasc:
Suggested by Shumon Huque: https://www.ietf.org/mail-archive/web/dnsop/current/msg23143.html

### DIFF
--- a/draft-schwartz-httpbis-dns-alt-svc.md
+++ b/draft-schwartz-httpbis-dns-alt-svc.md
@@ -168,18 +168,29 @@ an Alt-Svc Field Value.  Instead, server operators SHOULD encode the
 expiration time in the DNS TTL, and MUST NOT set a TTL longer than the
 intended "max age".
 
-Server operators MAY publish multiple ALTSVC records as an RRSET, with
-semantics equivalent to other mechanisms of providing multiple Alt-Svc
-values to the client.  When publishing an RRSET with multiple ALTSVC
+When receiving an ALTSVC record, clients MAY synthesize a new "ma"
+parameter from the DNS TTL, in order to interoperate with Alt-Svc processing
+subsystems.
+
+## Multiple records and values
+
+Server operators MAY publish multiple ALTSVC records as an RRSET.  When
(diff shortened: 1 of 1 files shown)


bemasc:
OK.  Could we say that the semantics are the same as getting multiple Alt-Svc response headers?

RFC 2616 says "Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)].  It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma."

Alt-Svc says "When multiple values are present, the order of the values reflects the server's preference".

I think this is actually workable.  The server can produce multiple RRs, each of which can contain multiple options in preference order.  The DNS client will scramble them, and then HTTP will glue them together in that order.  Then the client will (probably) use the first one that's compatible.

bemasc:
OK, I added reassurances around multiple alt-values and explicit recommendations for client behavior when receiving multiple records.  Please take a look.

draft-schwartz-httpbis-dns-alt-svc.md
```diff
@@ -168,15 +168,29 @@ an Alt-Svc Field Value.  Instead, server operators SHOULD encode the
...
+
+When receiving an RRSET containing multiple ALTSVC records, clients
+SHOULD apply a random shuffle to the records before using them, to
+ensure randomized load-balancing.  This is consistent with DNS rules for
+interpretation of RRSETs, which are regarded as unordered collections.
+
+After shuffling the RRSET, the client SHOULD concatenate all the values
+in the RRSET, separated by commas.  (This is semantically equivalent to
+receiving multiple Alt-Svc HTTP response headers, according to Section 4.2
+of {{?HTTP=RFC2616}}).
```
MikeBishop: Cite the current HTTP RFC, not the obsolete one; this should be Section 3.2.2 of RFC7230.

bemasc:
Citation fixed.

PR #2, opened 2019-05-16
Add Erik Nygren's HTTPSSVC version of the draft

bemasc:
This is Erik's original draft from my inbox, except for RFC-markdown syntax and linter fixes.

### DIFF
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -393,8 +393,8 @@ track users or hinder their connections after they leave that network.
 
 ## Multiple records and preference ordering {#pri}
 
-Server operators MAY publish multiple HTTPSSVC records as an RRSET.  When
-publishing an RRSET with multiple HTTPSSVC
+Server operators MAY publish multiple SvcRecordType "1" HTTPSSVC 
+records as an RRSET.  When publishing an RRSET with multiple HTTPSSVC
 records, the server operator MUST set the overall TTL to the minimum
 of the "max age" values (following Section 5.2 of {{!RFC2181}}).
 
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -306,7 +306,7 @@ For example, if an operator of https://example.com wanted to
 point HTTPS requests to a service operating at svc.example.net,
 they would publish a record such as:
 
- example.com. 3600 IN HTTPSSVC 1 svc.example.net.
+ example.com. 3600 IN HTTPSSVC 0 svc.example.net.
 
 The SvcDomainName MUST point to a domain name that contains
 another HTTPSSVC record and/or address (AAAA and/or A) records.
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -1,0 +1,789 @@
+---
+title: HTTPSSVC service location and parameter specification via the DNS (DNS HTTPSVC)
+abbrev: HTTPSSVC RR for DNS
+docname: draft-nygren-httpbis-httpssvc-latest
+date: {DATE}
+category: std
+
+ipr: trust200902
+area: General
+workgroup: HTTP Working Group
+keyword: Internet-Draft
+
+stand_alone: yes
+pi: [toc, sortrefs, symrefs]
+
+author:
+ -
+    ins: B. Schwartz
+    name: Ben Schwartz
+    organization: Google
+    email: bemasc@google.com
+ -
+    ins: M. Bishop
+    name: Mike Bishop
+    org: Akamai Technologies
+    email: mbishop@evequefou.be
+ -
+    ins: E. Nygren
+    name: Erik Nygren
+    org: Akamai Technologies
+    email: erik+ietf@nygren.org
+
+normative:
+
+informative:
+
+--- abstract
+
+This document specifies an "HTTPSSVC" DNS resource record type to
+facilitate the lookup of information needed to make connections for
+HTTPS URIs.  The HTTPSSVC DNS RR mechanism allows an HTTPS origin
+hostname to be served from multiple network services, each with
+associated parameters (such as transport protocol and keying material
+for encrypting TLS SNI).  It also provides a solution for the
+inability of the DNS to allow a CNAME to be placed at the apex
+of a domain name.
+
+By allowing this information to be bootstrapped in the DNS,
+it allows for clients to learn of alternative services before their first
+contact with the origin.  This arrangement offers potential benefits to
+both performance and privacy.
+
+--- middle
+
+# Introduction
+
+The HTTPSSVC RR is intended to address a number of challenges
+facing HTTPS clients and services, while also providing an extensible
+model to handle similar use-cases without forcing clients to perform
+additional DNS lookups and without forcing them to first make connections
+to a default service for the origin.
+
+When clients need to make a connection to fetch resources associated
+with an HTTPS URI, they must first resolve A and/or AAAA address resource
+records for the origin hostname.  This is adequate when clients default
+to TCP port 443, do not support Encrypted SNI {{!ESNI=I-D.ietf-tls-esni}},
+and where the origin service operator does not have a desire
+to put an CNAME at a zone apex (such as for "https://example.com").
+Handle situations beyond this within the DNS requires learning additional
+information, and it is highly desirable to minimize the number of round-trip
+and lookups required to learn this additional information.
+
+## Introductory Example
+
+As an introductory example, a set of example HTTPSSVC and associated
+A+AAAA records might be:
+
+ www.example.com.  2H  IN CNAME   svc.example.net.
+ example.com.      2H  IN HTTPSVC 0 svc.example.net.
+ svc.example.net.  2H  IN HTTPSVC 1 svc2.example.net. "hq=\\":8002\\" esnikey=\\"...\\""
+ svc.example.net.  2H  IN HTTPSVC 1 svc3.example.net. "h2=\\":8003\\" esnikey=\\"...\\""
+ svc2.example.net. 300 IN A       192.0.2.2
+ svc2.example.net. 300 IN AAAA    2001:db8::2
+ svc3.example.net. 300 IN A       192.0.2.3
+ svc3.example.net. 300 IN AAAA    2001:db8::3
+
+In the preceeding example, both of the "example.com" and
+"www.example.com" origin names are aliased to use service endpoints
+offered as "svc.example.net" (with "www.example.com" continuing to use
+a CNAME alias).  HTTP/2 is available on a cluster of machines located
+at svc2.example.net with TCP port 8002 and HTTP/3 is available on a
+cluster of machines located at svc3.example.net with UDP port 8003.
+An ESNI key is specified which allows the SNI values of "example.com"
+and "www.example.com" to be encrypted in the handshake with these
+service endpoints.  When connecting, clients will continue to treat
+the authoritative origins as "https://example.com" and
+"https://www.example.com", respectively.
+
+## Goals of the HTTPSSVC RR
+
+The goal of the HTTSSVC RR is to allow clients to resolve a single
+additional DNS RR in a way that:
+
+* Provides service endpoints authoritative for an origin,
+  along with parameters associated with each of these endpoints.
+  In particular:
+    * to support connecting directly to {{!HTTP3=I-D.draft-ietf-quic-http-20}} (QUIC transport)
+      service endpoints
+    * to obtain the {{!ESNI}} keys associated with a service endpoint
+* Does not assume that all service endpoints have the same parameters
+  (such as ESNI keys) or capabilities (such as {{!HTTP3}}) or are even
+  operated by the same entity.  This is important as DNS does not
+  provide any way to tie together multiple RRs for the same name.
+  For example, if www.example.com is a CNAME alias that switches
+  between one of three CDNs or hosting enviroments, records (such as A and AAAA)
+  for that name may have been sourced from different environments.
+* Enables the functional equivalent of a CNAME at a zone apex (such as
+  "example.com") for HTTPS traffic, and generally enables
+  delegation of operational authority for an HTTPS origin
+  within the DNS to an alternate name.  This addresses
+  a set of long-standing issues due to HTTP(S) clients
+  not implementing support for SRV records, as well as
+  due to a limitation that a DNS name can not have
+  both a CNAME record as well as NS RRs
+  (as is the case for zone apex names)
+
+
+## Overview of the HTTPSSVC RR
+
+This subsection briefly describes the HTTPSVC RR in
+a non-normative manner.
+
+The HTTPSSVC RR has three primary fields:
+
+1. SvcRecordType: A numeric flag indicating how to interpret the
+   subsequent fields.  When "0", it indicates that the RR contains an
+   alias.  When "1", it indicates that the RR contains an alternative service
+   definition.
+2. SvcDomainName: The domain name of either the alias target (when
+   SvcRecordType is "0") or the uri-host domain name of the alternative service
+   endpoint (when SvcRecordType is "1").
+3. SvcFieldValue: An Alternative Service field value describing the
+   alternative service endpoint for the domain name specified in
+   SvcDomainName (only when SvcRecordType is "1" and otherwise empty).
+
+DNS recursive resolvers may perform subsequent record resolution (for
+HTTPSSVC, A, and AAAA records) and may return them in the Additional Section
+of the response.  Clients must either use responses included
+in the additional section returned by the recursive resolver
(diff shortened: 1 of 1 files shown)


draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+
+## Omitting Max Age and Persist
+
+When publishing an HTTPSSVC record in DNS, server operators MUST omit the
+"ma" parameter, which encodes the "max age" (i.e. expiration time) of
+an Alt-Svc Field Value.  Instead, server operators SHOULD encode the
+expiration time in the DNS TTL, and MUST NOT set a TTL longer than the
+intended "max age".
+
+When receiving an HTTPSSVC record, clients MAY synthesize a new "ma"
```
MikeBishop: Why are clients only permitted to synthesize a lifetime instead of encouraged?

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+parameter from the DNS TTL, in order to interoperate with Alt-Svc processing
+subsystems.
+
+When publishing an HTTPSSVC record, server operators MUST omit the
+"persist" parameter, which indicates whether the client should use
+this record on other network paths.  When receiving an HTTPSSVC record,
+clients MUST discard any records that contain a "persist" flag.
+Disabling persistence is important to prevent an adversary on one
```
MikeBishop: I’m dubious that omitting persist actually helps.  Once the client has connected to the server, if the server possesses a valid certificate, then it simply supplies the persistent record as an HTTP header.  If the server doesn’t possess a valid certificate, then the persistent entry will fail to connect anyway.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+network from implanting a forged DNS record that allows them to
+track users after they leave that network.
+
+## Multiple records and preference ordering
+
+Server operators MAY publish multiple HTTPSSVC records as an RRSET.  When
```
MikeBishop: Given the requirement to only have one alias (2.4), perhaps it’s worth specifying that this is only applicable to the alt-svc form?

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+in Section 3 of {{!AltSvc}}.
+
+As RRs within an RRSET are explictly unordered collections, a new
+"pri" alt-svc parameter is introduced to indicate priority.  The value
+of the "pri" parameter is a non-negative integer between 0 and 65535.
+Alt-Svc alt-values with a smaller pri value should be given preference
+over alt-values with a larger pri value.
+
+TBD: what should default pri values be for Alt-Svc records
+received via DNS and HTTP when pri is unspecified?
```
MikeBishop: Information coming over an HTTPS connection is authenticated by the certificate, while DNS is potentially unauthenticated.  Records from an authenticated connection are always higher priority than those from DNS.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+received over HTTPS to those received via DNS.
+
+
+# Client behaviors
+
+TODO: Write more details on the recommended resolution algorithm.
+This section does NOT yet cover the SvcRecordType=0 case well yet.
+
+
+## HTTP Strict Transport Security
```
MikeBishop: I’m leery of making this HSTS-equivalent, especially since it would be trivial to permit 2.2 to indicate that the https scheme can be served over HTTPS using existing Alt-Svc mechanisms.  Just point to RFC 8164 and keep going.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+already in cache or because the host is an IP address.
+
+Note that this procedure does not rely on recursive resolvers handling
+the HTTPSSVC record type correctly.  If HTTPSSVC queries receive spurious
+NXDOMAIN responses, or even no response at all, connections will proceed
+as usual without any delay.
+
+## Optimizing for privacy
+
+Clients that are optimizing for privacy SHOULD implement {{!AltSvcSNI}}
```
MikeBishop: This references the draft which defines sni=, but you redefine it in 6.1.  Need to resolve.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+improve user privacy.  Server operators implementing this standard
+SHOULD also implement TLS 1.3 {{!I-D.ietf-tls-tls13}} and OCSP Stapling
+{{!RFC6066}}, both of which confer substantial performance and privacy
+benefits when used in combination with HTTPSSVC records.
+
+To realize the greatest privacy benefits, this proposal is intended for
+use with a privacy-preserving DNS transport (like DNS over TLS
+{{!RFC7858}} or DNS over HTTPS {{!DOH=I-D.ietf-doh-dns-over-https}}),
+and with the "SNI" Alt-Svc Parameter
+{{!AltSvcSNI=I-D.bishop-httpbis-sni-altsvc}}.  However, performance
```
MikeBishop: This references the draft which defines sni=, but you redefine it in 6.1.  Need to resolve.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+
+# Acknowledgements and Related Proposals
+
+There have been a wide range of proposed solutions over the years to
+the "CNAME at the Zone Apex" challenge proposed.  These include
+{{?I-D.draft-bellis-dnsop-http-record-00}},
+{{?I-D.draft-ietf-dnsop-aname-03}}, and others.
+
+
+# Appendix
```
MikeBishop: This is titled "Appendix," but is a top-level section in the body.  I'd suggest making each of these subsections an actual and separate appendix.

draft-nygren-httpbis-httpssvc.md
bemasc: The concern here is about an active attacker on the local network who is also a pervasive passive adversary.  This attacker could inject a fraudulent DNS record that would then permit them to track that user even after they leave the network.  The adversary in this case is not the server operator.

For example, suppose there is a server operator who doesn't use Alt-Svc at all, and who serves the same domain on all IPs of a /64.  An adversary who can inject a long-TTL DNS record can pin a user to one of those IPs, and then reidentify that user across the internet whenever they visit that domain.  This can (and should) be solved by clearing the client's DNS cache when moving between networks.  Since DNS data in this case can be cached in the Alt-Svc cache, we need "persist=0" to get equivalent behavior.

draft-nygren-httpbis-httpssvc.md
bemasc: I agree, how about
> clients SHOULD synthesize a new "ma" parameter from the DNS TTL if they intend to integrate with an Alt-Svc cache.

draft-nygren-httpbis-httpssvc.md
bemasc: I think HSTS support here is important.  It's a rare chance to fix the "bare domain" bootstrap problem, and would win a lot of support for HTTPSSVC from parties who would otherwise not be very interested in this stuff.

I don't view opportunistic security as sufficient, and I think it's considerably more complicated than "HTTPSSVC requires HTTPS".

bemasc:
Thanks for the review @MikeBishop.  I added comments on a few points where I contributed text, but I'll defer to @enygren on the others.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+
+
+## SvcRecordType
+
+The SvcRecordType field is a numeric value defined to be either "0" or
+"1". Within an HTTPSSVC RRSet, all RRs must have the same value for
+SvcRecordType.  Clients and recursive servers MUST ignore HTTPSSVC
+resource records with other SvcRecordType values, as well as
+RRSets where both "0" and "1" values for SvcRecordType are present in
+the same RRSet.
```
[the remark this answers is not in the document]
enygren: Good idea...  At least keeps it deterministic.

draft-nygren-httpbis-httpssvc.md
enygren: I tend to agree wtih Ben here that we should at least try and propose this.

draft-nygren-httpbis-httpssvc.md
enygren: Agreed.

enygren:
No disagreement with any of the remaining comments which are mostly editorial.

draft-nygren-httpbis-httpssvc.md
bemasc: As I've mentioned before, I tend to think we should drop this, based on the chilly reception we've gotten thus far.  The browser folks I've talked to seemed particularly opposed to implementing any protocol that created mismatch between the SNI and the actual validation process, because of the potential for security bugs.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+A+AAAA records might be:
+
+ www.example.com.  2H  IN CNAME   svc.example.net.
+ example.com.      2H  IN HTTPSVC 0 svc.example.net.
+ svc.example.net.  2H  IN HTTPSVC 1 svc2.example.net. "hq=\\":8002\\" esnikey=\\"...\\""
+ svc.example.net.  2H  IN HTTPSVC 1 svc3.example.net. "h2=\\":8003\\" esnikey=\\"...\\""
+ svc2.example.net. 300 IN A       192.0.2.2
+ svc2.example.net. 300 IN AAAA    2001:db8::2
+ svc3.example.net. 300 IN A       192.0.2.3
+ svc3.example.net. 300 IN AAAA    2001:db8::3
```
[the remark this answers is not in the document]
bemasc: Fixed in #12

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+
+The SvcFieldValue of the HTTPSSVC resource record contains an Alt-Svc
+Field Value, exactly as defined in Section 4 of {{!AltSvc}},
+but with the uri-host moved to the SvcDomainName field.
+
+For example, if the operator of https://www.example.com
+intends to include an HTTP response header like
+
+ Alt-Svc: hq="svc.example.net:8003"; ma=3600, \
+          h2="svc.example.net:8002"; ma=3600
```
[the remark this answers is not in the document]
bemasc: Fixed in #12

draft-nygren-httpbis-httpssvc.md
bemasc: Fixed in #12

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+When SvcRecordType is 0, the length of SvcFieldValue SHOULD be 0
+and clients MUST ignore the contents of non-empty SvcFieldValue fields.
+
+
+## RRNames
+
+In the case of the HTTPSSVC RR, an origin is translated into the RRName
+in the following manner:
+
+1. The origin scheme MUST be "https".  If the port is unspecified or "443",
```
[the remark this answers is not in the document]
bemasc: Addressed by #13.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+
+ www.example.com. 3600 IN TYPE??? \\# TBD:WRITEME
+
+This construction is intended to be extensible in two ways.  First,
+any extensions that are made to the Alt-Svc format for transmission
+over HTTPS are also applicable here, unless expressly mentioned
+otherwise.
+
+Second, by defining a way to map non-HTTPS schemes and non-default
+ports we provide a way for the HTTPSSVC to be used for them as needed.
```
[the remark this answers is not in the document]
bemasc: Addressed by #13.

draft-nygren-httpbis-httpssvc.md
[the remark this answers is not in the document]
bemasc: Addressed in #12.

draft-nygren-httpbis-httpssvc.md
bemasc: Addressed in #12

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+portion of the alt-authority component
+moved into the SvcDomainName field of the HTTPSSVC RR.
+Client receiving this information during DNS resolution
+can skip the initial connection and proceed directly to an
+alternative service.
+
+
+## Additional Alt-Svc parameters
+
+This document also defines additional Alt-Svc parameters
```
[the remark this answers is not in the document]
bemasc: Fixed in #12

draft-nygren-httpbis-httpssvc.md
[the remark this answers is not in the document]
bemasc: What is the advantage of preferring version 0?  Intuitively, I would expect a preference for version 1, since it enables better performance and security.  For future-proofing, I could imagine a rule like "prefer the highest type number you support, and ignore anything higher".

draft-nygren-httpbis-httpssvc.md
bemasc: Proposal in #14.

draft-nygren-httpbis-httpssvc.md
bemasc: Clarified in #15.

draft-nygren-httpbis-httpssvc.md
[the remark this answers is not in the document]
MikeBishop: Two reasons, but I could be persuaded otherwise:  There's a requirement that there be only one "0."  If the rule is "consume the first zero, ignore anything else" then you have an easy and deterministic rule.  Also, it makes it easy to roll out "1" by populating the new records prior to removing the old one.

And of course, something is better than nothing -- I'm amenable to other deterministic ways of selecting, but I'm less attracted to "pretend you didn't see it" as a mandated reaction.

draft-nygren-httpbis-httpssvc.md
MikeBishop: I'm amenable to leaving them separate -- my comment is purely around document consistency.  It would be consistent if all references to both were removed, too.  :-)

draft-nygren-httpbis-httpssvc.md
MikeBishop: The DoS issue is solved by Alt-Svc failing back to a direct connection if alternatives fail.  But the tracking argument is valid -- the attacker could just relay the packets to the real server, meaning the TLS handshake will complete (albeit with TLS RTT exceeding TCP's RTT from the client's point of view).

bemasc: Good point, that makes this attack even easier.

draft-nygren-httpbis-httpssvc.md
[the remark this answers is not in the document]
enygren: I think what Mike says makes sense (to use one of the "0"'s, ignore the others if multiple 0's or a mix of 0/1 are present).  These aren't different versions/types, they have different semantics.  (The "0" is a CNAME/ANAME equivalent and the "1" is the ALTSVC record.)  I guess we could swap the 0/1 value definitions and then say "use the highest you understand" or we could also just say to ignore ones you don't understand and then let any future draft that defines "2" if-any tell clients how to prefer or disprefer it.)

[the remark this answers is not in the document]
bemasc: OK, proposed text in #16

draft-nygren-httpbis-httpssvc.md
enygren: I'm fine with dropping this.  It could always be a separate document and reduces complexity here.

enygren: Created PR #17 for this.

draft-nygren-httpbis-httpssvc.md
enygren: Added clarification in #9e5d5a2f60c5410598ca1e9b3c916b75ea452855

draft-nygren-httpbis-httpssvc.md
```diff
@@ -0,0 +1,791 @@
...
+  for that name may have been sourced from different environments.
+* Enables the functional equivalent of a CNAME at a zone apex (such as
+  "example.com") for HTTPS traffic, and generally enables
+  delegation of operational authority for an HTTPS origin
+  within the DNS to an alternate name.  This addresses
+  a set of long-standing issues due to HTTP(S) clients
+  not implementing support for SRV records, as well as
+  due to a limitation that a DNS name can not have
+  both a CNAME record as well as NS RRs
+  (as is the case for zone apex names)
```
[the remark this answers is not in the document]
enygren: Added PR #18 to address this.  (Although later in the doc.)

PR #9, opened 2019-05-30
Defend HSTS behavior against certain attackers

bemasc:
With this change, the behavior might even be worthy of
the name "strict".

### DIFF
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -481,14 +481,18 @@ Because HTTPSSVC is received over an often insecure channel (DNS),
 clients MUST NOT place any more trust in this signal than if they
 had received a 302 redirect over cleartext HTTP.
 
-If the HTTPSSVC query results in a SERVFAIL error, the client SHOULD abandon
-the connection attempt and display an error message.  A SERVFAIL error can
-occur if the domain is DNSSEC-signed, the recursive resolver is
+If the HTTPSSVC query results in a SERVFAIL error, and the connection
+between the client and the recursive resolver is cryptographically protected
+(e.g. using TLS {{!RFC7858}} or HTTPS {{!RFC8484}}), the client SHOULD
+abandon the connection attempt and display an error message.  A SERVFAIL
+error can occur if the domain is DNSSEC-signed, the recursive resolver is
 DNSSEC-validating, and an active attacker between the recursive resolver
-and the authoritative resolver is attempting to prevent the upgrade to
-HTTPS.  Similarly, if the client validates DNSSEC, it SHOULD abandon the
-connection attempt if the HTTPSSVC RR fails to validate.
+and the authoritative DNS server is attempting to prevent the upgrade to
+HTTPS.
 
+Similarly, if the client enforces DNSSEC validation on A/AAAA
+RRs, it SHOULD abandon the connection attempt if the HTTPSSVC RR fails
+to validate.
 
 ## Cache interaction
 
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -481,6 +481,14 @@ Because HTTPSSVC is received over an often insecure channel (DNS),
 clients MUST NOT place any more trust in this signal than if they
 had received a 302 redirect over cleartext HTTP.
 
+If the HTTPSSVC query results in a SERVFAIL error, the client SHOULD abandon
+the connection attempt and display an error message.  A SERVFAIL error can
+occur if the domain is DNSSEC-signed, the recursive resolver is
+DNSSEC-validating, and an active attacker between the recursive resolver
+and the authoritative resolver is attempting to prevent the upgrade to
+HTTPS.  Similarly, if the client validates DNSSEC, it SHOULD abandon the
+connection attempt if the HTTPSSVC RR fails to validate.
+
 
 ## Cache interaction
 


draft-nygren-httpbis-httpssvc.md
```diff
@@ -481,6 +481,14 @@ Because HTTPSSVC is received over an often insecure channel (DNS),
...
 had received a 302 redirect over cleartext HTTP.
 
+If the HTTPSSVC query results in a SERVFAIL error, the client SHOULD abandon
+the connection attempt and display an error message.  A SERVFAIL error can
+occur if the domain is DNSSEC-signed, the recursive resolver is
+DNSSEC-validating, and an active attacker between the recursive resolver
+and the authoritative resolver is attempting to prevent the upgrade to
+HTTPS.  Similarly, if the client validates DNSSEC, it SHOULD abandon the
+connection attempt if the HTTPSSVC RR fails to validate.
+
```
enygren: Presumably this could also happen for servers that are misbehaving and don't understand the HTTPSSVC record type in the query?  I think we'd need more data an implementation experience.  Perhaps we could reword this as a MAY?  (Or put in Security Considerations?)

bemasc: Yeah, it depends on how much faith the client has in their resolver.  I would be OK with conditioning this on a secure transport to the resolver (as a proxy for "decent resolver, won't SERVFAIL on new RRTYPEs").

bemasc: OK, I've updated this PR to only apply to "cryptographically protected connections".  Please review.

bemasc:
> I'm a little leery of assuming one feature supported means a different one isn't broken.

Supporting unknown RRTYPEs has been the standard [since 2003](https://tools.ietf.org/html/rfc3597), and almost all servers do this correctly.  I don't think we need to accommodate servers who've implemented the still-rare standard from 2018 but not the near-universal one from 2003.  AFAIK, no such servers exist today, and once HTTPSSVC is widely deployed, anybody who tried to set up such a server would quickly figure out that it is seriously broken

Additionally, this accommodation is not the only reason to apply this logic.  If you don't have a secure connection to your recursive, then respecting SERVFAIL doesn't actually protect you from a network adversary (specifically an adversary between you and the recursive).  With a validating resolver and a secure connection, respecting SERVFAIL provides reliable protection (on signed domains).

> Perhaps a better approach is to synthesize an HSTS record just like we synthesize an Alt-Svc record and persist that independently of clearing the Alt-Svc records.

This approach is still opportunistic, not reliable against an active attacker.  It would also enable a persistent DoS by a transient local network adversary (as with persist=1).  I think this functionality is better served by the existing `Strict-Transport-Security` header, which can be set securely by the site itself on first access.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -481,6 +481,18 @@ Because HTTPSSVC is received over an often insecure channel (DNS),
...
+(e.g. using TLS {{!RFC7858}} or HTTPS {{!RFC8484}}), the client SHOULD
+abandon the connection attempt and display an error message.  A SERVFAIL
+error can occur if the domain is DNSSEC-signed, the recursive resolver is
+DNSSEC-validating, and an active attacker between the recursive resolver
+and the authoritative DNS server is attempting to prevent the upgrade to
+HTTPS.
+
+Similarly, if the client enforces DNSSEC validation on A/AAAA
+RRs, it SHOULD abandon the connection attempt if the HTTPSSVC RR fails
+to validate.
```
enygren: Is this special in any way?  How is this different/new behavior here?  (Are there clients that always enforce DNSSEC validation on all A/AAAA RRs?) While nice, I'm not sure this needs to be included here.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -481,6 +481,18 @@ Because HTTPSSVC is received over an often insecure channel (DNS),
...
```
enygren: Is this special in any way? How is this different/new behavior here? (Are there clients that always enforce DNSSEC validation on all A/AAAA RRs?) While nice, I'm not sure this needs to be included here.

draft-nygren-httpbis-httpssvc.md
enygren: Looks fine now.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -562,6 +517,39 @@ Authoritative DNS servers SHOULD return A, AAAA, and HTTPSSVC records
...
+and the client does not have address records for any of these values in
+its DNS cache, the client SHOULD perform an additional address query for
+the selected SvcDomainName.
+
+The additional DNS query in this case introduces a delay.  To avoid
+causing a delay for clients using a nonconforming recursive resolver,
+domain owners SHOULD choose the SvcDomainName to be a name in the
+origin hostname's CNAME chain if possible.  This will ensure that the required
+address records are already present in the client's DNS cache.
+
```
enygren: ... "or are received as part of the reponses to the AAAA and/or A queries that were issued-in-parallel."

draft-nygren-httpbis-httpssvc.md
```diff
@@ -562,6 +517,39 @@ Authoritative DNS servers SHOULD return A, AAAA, and HTTPSSVC records
...
+Highly performance-sensitive clients MAY implement the following special-
+case shortcut to avoid increased connection time: if (1) all of the
+HTTPSSVC records returned have SvcRecordType = 0, (2) none of their
+SvcDomainNames are in the DNS cache, and (3) the address queries for the
+origin domain return usable IP addresses, then the client MAY ignore the
+HTTPSSVC records and connect directly to the origin domain.
+
```
enygren: Add:  "When the SvcDomainNames and any needed HTTPSSVC records are available, these client SHOULD make subsequent requests over connections specified by the HTTPSSVC records."

(I suspect that for transitional purposes the A/AAAA record alongside the origin domains in these cases will provide sub-par performance relative to the target of the HTTPSSVC record.)

draft-nygren-httpbis-httpssvc.md
bemasc: The issue is that Alt-Svc and HTTPSSVC are "optional", and connection can always continue without them.  This text is intended to clarify that while their presence is optional (NXDOMAIN isn't a problem), certain failure modes should be treated as fatal, even if the connection actually would complete successfully if the client ignored the error, because this category of errors might represent adversary trying to force a downgrade.

Admittedly, this particular text is mostly academic/optimistic, since very few clients (and AFAIK no commercially deployed clients) actually validate DNSSEC on the client.

draft-nygren-httpbis-httpssvc.md
bemasc: Due to a mixup on my part this PR ended up showing diffs against the wrong branch, including this text.  I've fixed this PR and opened #19 under your name for this change.

draft-nygren-httpbis-httpssvc.md
bemasc: Moved to #19.

PR #10, opened 2019-05-31
Change performance-optimized client behavior

bemasc:
This gives clients an opt-out in the rare cases where
HTTPSSVC would impair performance, and also gives
server admins a clear way to avoid these cases.

### DIFF
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -492,63 +492,6 @@ connection as usual.
 If the client has a cached Alt-Svc entry that is expiring, the
 client MAY perform an HTTPSSVC query to refresh the entry.
 
-## Optimizing for performance
-
-Clients that are optimizing for performance (i.e. minimum connection
-setup time) SHOULD implement the following connection sequence:
-
-1. Issue address (AAAA and/or A) queries, immediately followed by the
-   HTTPSSVC query.
-1. If an HTTPSSVC response is received first, proceed with alternative
-   service connection and ignore the address responses if they are no
-   longer relevant.
-1. Otherwise, initiate connection to the origin server.
-1. As soon as an Alt-Svc field value is received, through the DNS or
-   over HTTP, proceed with alternative service connection.  Do not
-   abort this connection if an Alt-Svc field value is received from the
-   other source later.
-
-If the HTTPSSVC and address queries return approximately simultaneously,
-this process typically saves three roundtrips on a fresh connection
-that uses Alt-Svc: one each for TCP, TLS 1.3, and HTTP.  (On subsequent
-connections, the Alt-Svc information is expected to be cached, so this
-procedure does not apply.)
-
-If a client can cache Alt-Svc entries that were received over both HTTP
-and DNS, the client MAY prefer entries that were received over HTTP.
-These records may be more narrowly targeted for the specific client.
-
-As an additional optimization, when choosing among multiple Alt-Svc
-values, clients MAY prefer those that will not require an address
-query, either because the corresponding address record is
-already in cache or because the host is an IP address.
-
-Note that this procedure does not rely on recursive resolvers handling
-the HTTPSSVC record type correctly.  If HTTPSSVC queries receive spurious
-NXDOMAIN responses, or even no response at all, connections will proceed
-as usual without any delay.
-
-## Optimizing for privacy
-
-Clients that are optimizing for privacy SHOULD implement {{!AltSvcSNI}}
-and DNS over a secure transport (e.g. {{!RFC7858}} or {{!DOH}}).
-Use of a secure transport is important not only for privacy protection,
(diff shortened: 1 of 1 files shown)


draft-nygren-httpbis-httpssvc.md
```diff
@@ -574,6 +517,41 @@ Authoritative DNS servers SHOULD return A, AAAA, and HTTPSSVC records
...
+For optimal performance (i.e. minimum connection setup time), clients
+SHOULD issue address (AAAA and/or A) and HTTPSSVC queries
+simultaneously, and SHOULD implement a client-side DNS cache.
+With these optimizations in place, and conforming DNS servers,
+using HTTPSSVC does not add network latency to connection setup.
+
+Using HTTPSSVC with HTTP/3 saves a roundtrip during connection setup,
+compared to HTTP over TLS 1.3 over TCP.  Performance-sensitive
+server operators SHOULD offer HTTP/3 in their HTTPSSVC records.
+
```
enygren: Does it?  I'd think with the source-address-token-validation it's the same number of RTTs for a cold client as the TCP+TLS 1.3 1RTT flow?  (Perhaps I'm misremembering.)

draft-nygren-httpbis-httpssvc.md
```diff
@@ -574,6 +517,41 @@ Authoritative DNS servers SHOULD return A, AAAA, and HTTPSSVC records
...
+A nonconforming recursive resolver might return an HTTPSSVC response with
+a nonempty SvcDomainName, without the corresponding address queries.  If
+all the HTTPSSVC RRs in the response have nonempty SvcDomainName values,
+and the client does not have address records for any of these values in
+its DNS cache, the client SHOULD perform an additional address query for
+the SvcDomainName.
+
```
enygren: Do we want to mention my hack here?  (That clients SHOULD look at the set of aliases / canonical names in the A/AAAA response chain to see if they have an A/AAAA resolution of the SvcDomainName in this recent response?)  And that operators may choose to have the SvcDomainName match a name that the client would see a resolution for as part of the A/AAAA lookup?

draft-nygren-httpbis-httpssvc.md
```diff
@@ -574,6 +517,41 @@ Authoritative DNS servers SHOULD return A, AAAA, and HTTPSSVC records
...
+The additional DNS query introduces a delay that can be longer or shorter
+than one round trip to the origin.  For simplicity, we assume that these
+two delays will be roughly similar on average, i.e. a 1 RTT delay.
+
```
enygren: DNS RTTs and Origin RTTs often have no relation and may be orders of magnitude different.  (The former may have cache miss impact and server processing time overhead, but there are also plenty of cases where the DNS recursive is close to the client and the origin is far away, or vice-versa.)

draft-nygren-httpbis-httpssvc.md
bemasc: That's true, but I thought most servers were expected not to perform that kind of validation unless an attack is detected.

draft-nygren-httpbis-httpssvc.md
bemasc: I think that's worth mentioning somewhere.  The client behavior might just be "decent DNS cache" behavior.

draft-nygren-httpbis-httpssvc.md
bemasc: Yeah, they're not well-correlated.  The main point here is to find a way to make the average net performance loss negligible (or even a net win), even on legacy recursives (which will surely be prevalent for many years).  My sense is that an acceptable solution:

1. Can't impose a latency-proportional average-case delay on any large category of connections.
2. Must provide security features reliably or not at all.

I'm certainly open to alternative solutions if you don't think this one will be convincing.  I just don't have any other good ideas for how to square this.

draft-nygren-httpbis-httpssvc.md
bemasc: Done

bemasc:
I've substantially changed the proposal here, incorporating your CNAME idea and also making "SvcRecordType = 0" optional for the client (since it doesn't have any security implication).

draft-nygren-httpbis-httpssvc.md
```diff
@@ -562,6 +505,39 @@ Authoritative DNS servers SHOULD return A, AAAA, and HTTPSSVC records
...
+origin hostname's CNAME chain if possible.  This will ensure that the required
+address records are already present in the client's DNS cache.
+
+Highly performance-sensitive clients MAY implement the following special-
+case shortcut to avoid increased connection time: if (1) all of the
+HTTPSSVC records returned have SvcRecordType = 0, (2) none of their
+SvcDomainNames are in the DNS cache, and (3) the address queries for the
+origin domain return usable IP addresses, then the client MAY ignore the
+HTTPSSVC records and connect directly to the origin domain.
+
```
enygren: I'm still a little dubious on this, but we can include and discuss.

PR #11, opened 2019-06-13
Ban fallback when ESNI is enabled

bemasc:


draft-nygren-httpbis-httpssvc.md
```diff
@@ -602,6 +602,14 @@ encoded in {{!base64=RFC4648}}.
...
 headers and HTTP/2 ALTSVC frames.
 
+The Alt-Svc specification states that "the client MAY fall back to using
+the origin" in case of connection failure {{!AltSvc}}.  This behavior is
+not suitable for ESNI, because fallback would negate the privacy benefits of
+ESNI.
+
+Accordingly, any connection attempt that uses ESNI MUST NOT fall back to a
+connection without ESNI.
+
```
enygren: Are there any caveats we'd want?  Such as for clients that don't care about the privacy benefits of ESNI?  (Does the ESNI spec itself allow for any fallbacks?) This otherwise seems generally reasonable.

bemasc: Why would a client that doesn't care about the privacy benefits of ESNI use ESNI?  Maybe the phrasing is unclear, but my goal was to specify that if the client cared enough to encrypt the SNI on the first try, then it shouldn't send un-encrypted SNI on the second try.

MikeBishop: For one thing, to contribute to the overall prevalence of ESNI on the Internet, reducing the odds that it stands out as unusual.  A client that will use ESNI when possible, but doesn't really care whether it's used in any particular instance, seems entirely plausible.

bemasc: An ESNI fallback behavior creates an incentive for a network adversary to block connections with ESNI, to force a retry without ESNI and reveal the name.  (If only some clients have fallback it's generally easy to target them, because most TLS clients have a distinctive ClientHello due to cipher and extension choices and ordering.)

ESNI is not designed to be opportunistic, and I don't think the TLS group would approve of a document that introduced that kind of behavior.  I'd rather have clients behave consistently, either implementing ESNI support or not.

Bottom line: "MUST" is used when the consequences of failure to comply are serious (and perhaps non-obvious).  I think this situation deserves a MUST.

enygren: I think it's reasonable to include this and then discuss with a broader audience?

enygren: Although one reason to *not* have this as a MUST:  at least one use-case I've been interested in getting implemented is to do ESNI for just HTTP/3 QUIC (as there's lots of value in trying to prevent middle boxes from messing with ESNI).  This isn't possible if doing so prevents QUIC fallback to HTTP/2 (which is necessary). 

Would an additional attribute indicating the ESNI is opportunistic and fallback is allowed be a compromise?  "esnioppt=1"?

bemasc: What a mess.  @MikeBishop, what is the exact semantics of unknown parameters in Alt-Svc?  If a client sees an alt-value with "esnikeys", and doesn't know what that parameter is for, does it ignore the parameter or the whole alt-value?

enygren: From rfc7838:  "Unknown parameters MUST be ignored.  That is, the values (alt-value) they appear in MUST be processed as if the unknown parameter was not present."

bemasc: OK, I've added a way to let server operators choose opportunistic ESNI.  Please review.

enygren: Sounds good.

PR #17, opened 2019-06-19
Remove Alt-Svc sni= parameter

enygren:
Finish removing the Alt-Svc "sni=" parameter (other than from the comparisons appendix discussion).


bemasc:
:+1:

PR #19, opened 2019-06-20
Clarify performance optimized behavior

bemasc:


bemasc:
Text is from @enygren.  :+1: from me.

PR #21, opened 2019-06-21
Define the resolution algorithm

enygren:


draft-nygren-httpbis-httpssvc.md
```diff
@@ -455,9 +455,31 @@ received over HTTPS to those received via DNS.
 
 # Client behaviors
 
-TODO: Write more details on the recommended resolution algorithm.
-This section does NOT yet cover the SvcRecordType=0 case well yet.
+## Client resolution {#resolution}
 
+When attempting to resolve a name HOST, clients should follow in-order:
+
+1. Issue parallel AAAA, A, and HTTPSSVC queries for the name HOST.
```
bemasc: Technically if you're v4 or v6 only this is obviously not quite right.  I think "A/AAAA" is the usual shorthand.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -455,9 +455,31 @@ received over HTTPS to those received via DNS.
...
+   need to resolve AAAA and/or A records for SvcDomainName.
+
+4. If only AAAA and/or A records are present for HOST (and no HTTPSSVC), 
+   clients should make a connection to one of the IP addresses
+   contained in these records.
+
+When selecting between AAAA and A records to use, clients may
+use an approach such as {{!HappyEyeballsV2=RFC8305}}
+
+Some possible optimizations are discussed in {{optimizations}}.
```
bemasc: I would add "When these optimizations are employed, this procedure does not generally add latency in comparison to an ordinary A/AAAA lookup.".  (But only if it's true.  I think it is, if recursives really follow SvcRecordType 0 chains.)

draft-nygren-httpbis-httpssvc.md
```diff
@@ -455,9 +455,31 @@ received over HTTPS to those received via DNS.
...
+
+3. If one or more HTTPSSVC record of SvcRecordType "1" is returned for HOST,
+   clients should synthesize equivalent Alt-Svc records based
+   on the SvcDomainName and SvcFieldValue.  If one of these Alt-Svc
+   records is selected to be used in a connection, the client will
+   need to resolve AAAA and/or A records for SvcDomainName.
+
+4. If only AAAA and/or A records are present for HOST (and no HTTPSSVC), 
+   clients should make a connection to one of the IP addresses
+   contained in these records.
```
bemasc: I would add "and proceed normally"

draft-nygren-httpbis-httpssvc.md
```diff
@@ -455,9 +455,31 @@ received over HTTPS to those received via DNS.
...
+## Client resolution {#resolution}
 
+When attempting to resolve a name HOST, clients should follow in-order:
+
+1. Issue parallel AAAA, A, and HTTPSSVC queries for the name HOST.
+   The answers for these may or may not include CNAME pointers
+   before reaching one or more of these records.
+   
+2. If an HTTPSSVC record of SvcRecordType "0" is returned for HOST,
+   clients should loop back to step 1 replacing HOST with SvcDomainName.
```
bemasc: Any loop detection logic?  Could be as simple as "subject to loop detection heuristics".

draft-nygren-httpbis-httpssvc.md
enygren: I added:
"to reduce latency impact in comparison to ordinary AAAA/A lookups."
I think we may want more experimental testing prior to making a stronger
statement as there are enough corner cases here.

PR #22, opened 2019-06-25
Added some suggestions from Ian Swett.

enygren:
Mention the HTTPS bootstrapping in the intro.
Remind the importance of using the origin name for validating
the TLS server cert.

### DIFF
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -43,7 +43,7 @@ hostname to be served from multiple network services, each with
 associated parameters (such as transport protocol and keying material
 for encrypting TLS SNI).  It also provides a solution for the
 inability of the DNS to allow a CNAME to be placed at the apex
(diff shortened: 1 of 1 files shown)


bemasc:
Looks good to me

draft-nygren-httpbis-httpssvc.md
```diff
@@ -43,7 +43,9 @@ hostname to be served from multiple network services, each with
 associated parameters (such as transport protocol and keying material
 for encrypting TLS SNI).  It also provides a solution for the
 inability of the DNS to allow a CNAME to be placed at the apex
-of a domain name.
+of a domain name.  It also provides a way to indicate that hostname
```
MikeBishop: From a pure reading-flow point of view, can we do something other than "It also provides" twice in a row?
```suggestion
of a domain name.  Finally, it provides a way to indicate that the origin
```

PR #27, opened 2019-06-28
Some more cleanups from moving pri to a field, with reorder

enygren:
Reorder SvcFieldPriority to be before SvcDomainName to be consistent with SRV and MX.
Add this in to some other examples.
Remove the IANA instruction for pri.


enygren:
Perhaps leave it in for now but leave that as open discussion?  It may also make sense to just put it back into an Alt-Svc attribute.

Issue #32, opened 2019-07-03
Remove ambiguities/redundancies in encoding


enygren:
From Ilari Liusvaara <ilariliusvaara@welho.com> 

- What if SvcDomainName has length different from its length field?
  DNS wire-form names are self-delimiting (DNS message parsing relies
  on this).
- What does it mean for SvcDomainName to be absent in alternative
  service form? I would guess it means "same as RRNAME".
- Why there is length field for SvcFieldValue? Why not let it run to
  the end of record?
- 2 byte length field can encode values up to 65535, not 65536.
  And the length of SvcFieldValue can not be that big, because
  RRDATA and DNS message length limits (both 65535) would be hit.

enygren:
Using the empty label (".") to specify an absent in alternative service form which means "same as RRName" would address both of the first two items.

bemasc:
But what if the root wants to serve QUIC?  OK, that seems unlikely, but technically it does seem like squatting on their name.  I'm not sure what the formalities of that are.  (I wonder if Let's Encrypt would issue a cert for "."...)

Thinking about the text representation, the clearest way I can think of to indicate an omitted uri-host is to write the SvcDomainName as "*".  I'm not sure what that suggests for the RR serialization though.

enygren:
It seems like some special cases for SvcDomainName to cover for SvcRecordType=1:

* How to indicate it should be equivalent to the RRNAME?
* How to indicate it should be equivalent to the hostname?
* What happens if it has the value "."?  SRV and MX have this
  mean "no service available.

enygren:
Released in -02.

PR #35, opened 2019-07-05
Remove redundancy from the wire format.

enygren:
Address most issues in #32 

We should still land:

## Special cases for SvcDomainName

When SvcRecordType=1, there are some special cases to handle
for SvcDomainName:
* How to indicate it should be equivalent to the RRNAME?
* How to indicate it should be equivalent to the hostname?
* What happens if it has the value "."?  SRV and MX have this
  mean "no service available.

### DIFF
--- a/draft-nygren-httpbis-httpssvc.md
+++ b/draft-nygren-httpbis-httpssvc.md
@@ -438,8 +438,11 @@ load-balancing.
 For a client to construct the equivalent of an Alt-Svc HTTP response header:
 
 1. For each RR, the SvcDomainName MUST be inserted as the uri-host.
-   If SvcDomainName is empty ("") then the RRNAME for the HTTPSSVC
-   record MUST be inserted as the uri-host.
+   If SvcDomainName is empty ("") then the RRNAME for the final HTTPSSVC
+   record MUST be inserted as the uri-host.  (In the case of a CNAME 
+   or a HTTPSSVC SvcRecordType "0" record pointing to an HTTPSSVC record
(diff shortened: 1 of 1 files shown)


draft-nygren-httpbis-httpssvc.md
```diff
@@ -441,6 +438,11 @@ load-balancing.
 For a client to construct the equivalent of an Alt-Svc HTTP response header:
 
 1. For each RR, the SvcDomainName MUST be inserted as the uri-host.
+   If SvcDomainName is empty ("") then the RRNAME for the final HTTPSSVC
```
bemasc: This talks about SvcDomainName being empty, but SvcDomainName can't be empty in this serialization.

draft-nygren-httpbis-httpssvc.md
```diff
@@ -830,3 +830,27 @@ as MX, SRV, and others all have their own variations here.
...
 to priority.  That is excluded here for simplicity.  It could always be
 added as an optional Alt-Svc attribute.
+
+## Special cases for SvcDomainName
+
+When SvcRecordType=1, there are some special cases to handle
+for SvcDomainName:
+* How to indicate it should be equivalent to the RRNAME?
+* How to indicate it should be equivalent to the hostname?
+* What happens if it has the value "."?  SRV and MX have this
```
bemasc: I think we should resolve this before merging.  If SRV and MX special-case ".", I think that we can do the same.

I don't think we need a way to say "equivalent to the hostname".  The hostname has already been delegated to this RRNAME.

PR #51, opened 2019-09-07
Proposed edits to Ben's edits.  Most are minor.

enygren:
Minor typo fixes and edits.  Add proxy behavior.  Add a/aaaa param clarifications.

### DIFF
--- a/draft-nygren-dnsop-svcb-httpssvc.md
+++ b/draft-nygren-dnsop-svcb-httpssvc.md
@@ -42,7 +42,7 @@ origin resources, such as for HTTPS URLs.  SVCB records
 allow an origin to be served from multiple network
 locations, each with associated parameters (such as transport protocol
 configuration and keying material for encrypting TLS SNI).  They also
-aliasing of apex domains, which not possible with CNAME.
+enable aliasing of apex domains, which is not possible with CNAME.
 The HTTPSSVC DNS RR is a variation of SVCB for HTTPS and HTTP origins.
 By providing more information to the client before it attempts to
 establish a connection, SVCB offers potential benefits to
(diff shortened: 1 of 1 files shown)


draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -542,6 +554,19 @@ use an approach such as {{!HappyEyeballsV2=RFC8305}}.
 Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
+## Clients using a Proxy
+
+Clients using a proxy via issuing CONNECT requests
+({{!RFC7231}} Section 4.3.6) should follow the same
+resolution for resolving SVCB records.  If a ServiceForm SVCB
```
bemasc: It sounds like you are proposing to chase the SVCB chain outside the proxy, and then use the final SvcDomainName in the request-target.  That seems like a fine idea to me, but I think the recommendation should be weaker.  I don't want anyone to think that we're telling them to leak DNS queries to an untrusted resolver when using a proxy for privacy purposes.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -639,6 +665,19 @@ The wire format for each parameter is a sequence of IP addresses in network
...
 
+These parameters MAY be repeated multiple times within a record.
+When this happens, the list of addresses SHOULD be concatenated.
+
+When selecting between AAAA and A records to use, clients may use an
+approach such as {{!HappyEyeballsV2=RFC8305}}.  Within an address
+family, clients SHOULD pick addresses to use in a random order.
+
+When only "a" parameters are present, IPv6-only clients MUST either
+resolve an AAAA records from the SvcDomainName or attempt
```
bemasc: I think this requirement already exists, regardless of whether the client is IPv6-only.

PR #52, opened 2019-09-09
Proxy and DNS64 text proposals

bemasc:


draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -557,16 +559,16 @@ to avoid additional latency in comparison to ordinary AAAA/A lookups.
...
-record is found, the client should select the highest priority acceptable
-option.  From this the client should construct the request-target
-used in the CONNECT request.
-
-In cases where no ServiceForm SVCB record is found, the client should
-proceed as normal and use HOST to construct the request-target.
+Clients using a domain-oriented transport proxy like HTTP CONNECT
+({{!RFC7231}} Section 4.3.6) or SOCKS5 ({{!RFC1928}}) SHOULD disable
+SVCB support if performing SVCB queries would violate the
+client's privacy intent.
```
enygren: Mention "unless SVCB queries are also issued via the proxy (eg, via DoH)"?

Part of the need to specify the proxy behavior is to get ESNI to work properly.  Should we spell this out here (to explain why this proxy behavior is important), or perhaps later?  ie, part of getting ESNI to work means the keys and SvcDomainName corresponding to the connect need to match up.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -659,24 +661,21 @@ The SvcParamKey for ESNI is "esnikeys".  Its value is defined in
...
-approach such as {{!HappyEyeballsV2=RFC8305}}.  Within an address
-family, clients SHOULD pick addresses to use in a random order.
-
-When only "a" parameters are present, IPv6-only clients MUST either
-resolve an AAAA records from the SvcDomainName or attempt
-to synthesize IPv6 addresses as specified in {{!RFC7050}}.
-This is necessary as recursive resolvers will not be able to
+approach such as {{!HappyEyeballsV2=RFC8305}}.
+When only "a" parameters are present, IPv6-only clients may synthesize
+IPv6 addresses as specified in {{!RFC7050}}.  Recursive resolvers MUST NOT
```
enygren: I'm still solidly in the "if you are skipping your local DNS resolver's DNS64, you MUST do rfc7050 synthesis" camp.  There are tens (hundreds?) of millions of mobile end-users who will be broken otherwise for IPv4-only sites.  An alternative would be to say that a "aaaa" parameter MUST be included if an "a" parameter is present?  (Which puts the onus on the server-side.)  If we don't have a MUST on the client or the server, recursive resolvers will do DNS64 synthesis here so having a MUST NOT there is unrealistic.

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: OK, I added an explanation of motivations for the proxy section.

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: > I'm still solidly in the "if you are skipping your local DNS resolver's DNS64, you MUST do rfc7050 synthesis" camp.

Me too.

> There are tens (hundreds?) of millions of mobile end-users who will be broken otherwise for IPv4-only sites.

I don't think anyone will be broken, because the alternative is to just ignore address hints you can't use.

> If we don't have a MUST on the client or the server...

We have a MUST on the client: "clients MUST perform A and/or AAAA queries for SvcDomainName as in {{client-behavior}}"

enygren: Good point.  So perhaps:

When only "a" parameters are present, IPv6-only clients may either synthesize IPv6 addresses as specified in {{!RFC7050}} or fall back to resolving SvcDomainName. To avoid this, "aaaa" parameters SHOULD be published whenever "a" parameters are published.

?

draft-nygren-dnsop-svcb-httpssvc.md
enygren: Sounds good.  Do we want to highlight that proxies trying to do content filtering and assuming that what they are connecting to is the Host/SNI could get confused?  (Which are the same proxies that ESNI will make sad.)

draft-nygren-dnsop-svcb-httpssvc.md
MikeBishop: Works for me.

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: I'd rather leave that alone if possible.  Using an "affiliated resolver" as described would cover that use case.

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: OK, I added a version of this.  Note that "fallback" is probably not an accurate depiction of what's going on here: the AAAA query MUST be issued regardless, and SHOULD be used as soon as it's ready, so the only question is whether to wait for it.

PR #56, opened 2019-09-25
Clarify racing behavior for A/AAAA and SVCB parallel queries. Give rationale to A/AAAA answer preference.

chris-wood:
We have text in the ESNI draft which describes what to do when A/AAAA arrives before ESNI (equivalently SVCB). This attempts to port that text over. Please have a look!

@enygren @MikeBishop @bemasc

### DIFF
--- a/draft-nygren-dnsop-svcb-httpssvc.md
+++ b/draft-nygren-dnsop-svcb-httpssvc.md
@@ -302,7 +302,7 @@ The presentation format of the record is:
 
     Name TTL IN SVCB SvcFieldPriority SvcDomainName SvcFieldValue
 
-The SVCB record is defined specifically within 
+The SVCB record is defined specifically within
 the Internet ("IN") Class ({{!RFC1035}}).
 SvcFieldPriority is a number in the range 0-65535,
 SvcDomainName is a domain name,
@@ -542,7 +542,7 @@ following procedure:
    SvcDomainName if they are not already available.  These are the
    preferred SvcFieldValue and IP addresses.  If the connection fails, the
    client MAY try to connect using values from a lower-priority record.
-   If none of the options succeed, the client SHOULD connect to the origin 
+   If none of the options succeed, the client SHOULD connect to the origin
    server directly.
 
 4. If an SVCB record for HOST does not exist, the received AAAA and/or A
@@ -551,7 +551,7 @@ following procedure:
 This procedure does not rely on any recursive or authoritative server to
 comply with this specification or have any awareness of SVCB.
 
-When selecting between AAAA and A records to use, clients may use an approach 
+When selecting between AAAA and A records to use, clients may use an approach
 such as {{!HappyEyeballsV2=RFC8305}}.
 
 Some important optimizations are discussed in {{optimizations}}
@@ -627,18 +627,19 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
-Moreover, if an A or AAAA response arrives before an SVCB response, a client which 
-prefers ESNI SHOULD wait up to CD milliseconds before starting secure connections 
-to either address. (Clients may begin TCP connections in this time. QUIC connections 
-should wait. This allows receipt of latent keying material in a SVCR repsonse.) 
-If an SVCB record with a "esnikeys" value and "ipv4hint" or "ipv6hint" values 
-arrives and one of those hints matches an in-progress connection, the secure 
-connection should proceed using the provided ESNI keying material. However, if 
-neither a "ipv4hint" nor "ipv6hint" is available, the client MUST resolve the 
-SvcDomainName in the SVCB response and connect to the preferred A or AAAA adddress
-according to the procedure in {{client-behavior}}.
+Moreover, if an A or AAAA response arrives before an SVCB response, a client which
+prefers ESNI SHOULD wait up to CD milliseconds before starting secure connections
+to either address. (Clients may begin TCP connections in this time. QUIC connections
+should wait. This allows receipt of latent keying material in a SVCB repsonse.)
+If an SVCB record with a "esnikeys" value and "ipv4hint" or "ipv6hint" values
+arrives and one of those hints matches an in-progress connection, the secure
+connection should proceed using the provided ESNI keying material. If neither a
+"ipv4hint" nor "ipv6hint" is available, clients must then resolve the SvcDomainName
+in the SVCB response to the preferred A or AAAA adddress. If a client's DNS cache has
+valid A or AAAA records for the SvcDomainName, those SHOULD be used. Otherwise,
+clients MUST resolve SvcDomainName to according to the procedure in {{client-behavior}}.
 
-CD (Connection Delay) is a configurable parameter.  The recommended value is 50 
+CD (Connection Delay) is a configurable parameter.  The recommended value is 50
 milliseconds, as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
 
 A nonconforming recursive resolver might not return all the information
--- a/draft-nygren-dnsop-svcb-httpssvc.md
+++ b/draft-nygren-dnsop-svcb-httpssvc.md
@@ -628,7 +628,7 @@ With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
 Moreover, if an A or AAAA response arrives before an SVCB response, the
-client SHOULD wait up to CD milliseconds before connecting as if the
+client MAY wait up to CD milliseconds before connecting as if the
 SVCB query returned NODATA, but MUST NOT transmit any information that
 could be altered by an SVCB response until the SVCB response arrives.
 For example, a TLS ClientHello may be altered by the "esnikeys" value
@@ -637,8 +637,13 @@ is a configurable parameter. The recommended value is 50 milliseconds,
 as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
 
 An SVCB record is consistent with an active or in-progress connection C
-if the A or AAAA address for its SvcDomainName matches the destination
-address of C. If an SVCB record is consistent with an active or in-progress
+if the following conditions are met:
+
+1. The A or AAAA address for its SvcDomainName matches the destination
+address of C.
+2. The destination port of C matches that specified in the SVCB record.
+
+If an SVCB record is consistent with an active or in-progress
 connection C, the client may continue using C with any information provided
 by the SVCB record. For example, if an SVCB record with a "esnikeys" value and
 "ipv4hint" or "ipv6hint" values {{svcparamkeys-iphints}} arrives, and one of
--- a/draft-nygren-dnsop-svcb-httpssvc.md
+++ b/draft-nygren-dnsop-svcb-httpssvc.md
@@ -631,28 +631,30 @@ Moreover, if an A or AAAA response arrives before an SVCB response, the
 client SHOULD wait up to CD milliseconds before connecting as if the
 SVCB query returned NODATA, but MUST NOT transmit any information that
 could be altered by an SVCB response until the SVCB response arrives.
-For example, a TLS connection may be altered by the "esnikeys" value
-of an SVCB response. CD (Connection Delay) is a configurable parameter.
-The recommended value is 50 milliseconds, as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
+For example, a TLS ClientHello may be altered by the "esnikeys" value
+of an SVCB response {{svcparamkeys-esnikeys}}. CD (Connection Delay)
+is a configurable parameter. The recommended value is 50 milliseconds,
+as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
 
-If an SVCB record is consistent with an active or in-progress connection,
-the client may continue using that connection using any information provided
-by the SVCB RR. For example, if an SVCB record with a "esnikeys" value and
-"ipv4hint" or "ipv6hint" values arrives and one of those hints matches the
-address of an in-progress connection, the TLS connection should proceed using
-the provided ESNI keying material.
+An SVCB record is consistent with an active or in-progress connection C
+if the A or AAAA address for its SvcDomainName matches the destination
+address of C. If an SVCB record is consistent with an active or in-progress
+connection C, the client may continue using C with any information provided
+by the SVCB record. For example, if an SVCB record with a "esnikeys" value and
+"ipv4hint" or "ipv6hint" values {{svcparamkeys-iphints}} arrives, and one of
+those hints matches the address of an in-progress connection TCP connection,
+TLS should proceed on that connection using the provided ESNI keying material.
 
 If the SVCB record is not consistent with any active or in-progress connection,
-e.g. there is no "ipv4hint" or "ipv6hint" that matches a connection, clients must
-then resolve the SvcDomainName in the SVCB response to the preferred A or AAAA
-adddress. If a client's DNS cache has valid A or AAAA records for the SvcDomainName,
-those SHOULD be used. Otherwise, clients MUST resolve SvcDomainName to according to
-the procedure in {{client-behavior}}.
+clients must then resolve the SvcDomainName in the SVCB response to the preferred
+A or AAAA adddress. If a client's DNS cache has valid A or AAAA records for the
+SvcDomainName, those SHOULD be used. Otherwise, clients MUST resolve SvcDomainName
+to according to the procedure in {{client-behavior}}.
 
 A nonconforming recursive resolver might not return all the information
 required to use all the records in an SVCB response.  If
-some of the SVCB RRs in the response can be used without requiring
-additional DNS queries, the client MAY prefer those RRs, regardless of
+some of the SVCB records in the response can be used without requiring
+additional DNS queries, the client MAY prefer those records, regardless of
 their priorities.
 
 To avoid a delay for clients using a nonconforming recursive resolver,
@@ -694,7 +696,7 @@ is the corresponding 2 octet numeric value in network byte order.
 The SvcParamKey for ESNI is "esnikeys".  Its value is defined in
 {{esnikeys}}.  It is applicable to most TLS-based protocols.
 
-## "ipv4hint" and "ipv6hint"
+## "ipv4hint" and "ipv6hint" {#svcparamkeys-iphints}
 
 The "ipv4hint" and "ipv6hint" keys represent IP address hints for the service.
 If A and AAAA records for SvcDomainName are locally available, the client SHOULD
--- a/draft-nygren-dnsop-svcb-httpssvc.md
+++ b/draft-nygren-dnsop-svcb-httpssvc.md
@@ -628,7 +628,7 @@ With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
 Moreover, if an A or AAAA response arrives before an SVCB response, the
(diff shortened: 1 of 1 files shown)


draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -237,7 +237,7 @@ Additional Section of the response.  Clients must either use responses
 included in the additional section returned by the recursive resolver
 or perform necessary SVCB, A, and AAAA record resolutions.  DNS
 authoritative servers may attach in-bailiwick SVCB, A, AAAA, and CNAME
-records in the Additional Section to responses for an SVCB query.
+records in the Additional Section to responses for a SVCB query.
```
tfpauly: Since you pronounce this "Ess Vee Cee Bee" it should remain "an"

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,20 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+Moreover, if an A or AAAA response arrives before an SVCB response, a client which 
+prefers ESNI SHOULD wait up to CD milliseconds before starting secure connections
```
bemasc: The current logic just requires the client to wait for the SVCB response.  As you note, the SVCB response contains security-relevant information, so it's important not to miss it.

Is the goal here to handle [DNS non-response](https://tools.ietf.org/html/draft-ietf-dnsop-no-response-issue-13)?  Even the most ancient DNS servers will return an error if there is a problem with the SVCB query.  Total non-response is a serious bug, but it does happen.

As that draft notes, the client [can measure this behavior](https://tools.ietf.org/html/draft-ietf-dnsop-no-response-issue-13#section-3.1.2).  If you think it's worth documenting a solution for this case, I'd prefer that approach, and I'd suggest doing it in a separate "Compatibility" section.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,20 @@ in cache before performing any followup queries.
...
+to either address. (Clients may begin TCP connections in this time. QUIC connections 
+should wait. This allows receipt of latent keying material in a SVCR repsonse.) 
+If an SVCB record with a "esnikeys" value and "ipv4hint" or "ipv6hint" values 
+arrives and one of those hints matches an in-progress connection, the secure 
+connection should proceed using the provided ESNI keying material. However, if 
+neither a "ipv4hint" nor "ipv6hint" is available, the client MUST resolve the 
+SvcDomainName in the SVCB response and connect to the preferred A or AAAA adddress
+according to the procedure in {{client-behavior}}.
+
+CD (Connection Delay) is a configurable parameter.  The recommended value is 50
```
bemasc: I think this would have a very high false-positive rate due to iterative lookups required to walk the SVCB chain.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: > Is the goal here to handle DNS non-response?

Not quite! The goal is to handle answers received out of order. It’s an optimization.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: This is the delay in waiting for the first SVCB response, not the delay in abandoning ESNI altogether. The value is meant to give the SVCB answer time to arrive since, presumably, clients would prefer using that over the A/AAAA answer(s). Maybe I’m misunderstanding your point.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,20 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+Moreover, if an A or AAAA response arrives before an SVCB response, a client which 
+prefers ESNI SHOULD wait up to CD milliseconds before starting secure connections 
+to either address. (Clients may begin TCP connections in this time. QUIC connections 
+should wait. This allows receipt of latent keying material in a SVCR repsonse.) 
+If an SVCB record with a "esnikeys" value and "ipv4hint" or "ipv6hint" values 
+arrives and one of those hints matches an in-progress connection, the secure
```
enygren: This should also reference {#optimizations} and also allow the use of responses in the client DNS cache, not just hints.  In-particular due to this it may often be possible that the hint values aren't needed if the SvcDomainName already has valid A/AAAA records in-cache (eg, due to being part of the CNAME chain from the A/AAAA lookups issued in-parallel):

> To avoid a delay for clients using a nonconforming recursive resolver, domain owners SHOULD use a single SVCB record whose SvcDomainName is in the origin hostname's CNAME chain if possible. This will ensure that the required address records are already present in the client's DNS cache as part of the responses to the address queries that were issued in parallel.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,20 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+Moreover, if an A or AAAA response arrives before an SVCB response, a client which 
+prefers ESNI SHOULD wait up to CD milliseconds before starting secure connections 
+to either address. (Clients may begin TCP connections in this time. QUIC connections 
+should wait. This allows receipt of latent keying material in a SVCR repsonse.)
```
enygren: Typo.  s/SVCR/SVCB/

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: OK.  As your text notes, SVCB covers a variety of transports and use cases, so I think we need to express this with more generality, like:
> If an A or AAAA response arrives before an SVCB response, the client MAY begin connection as if the SVCB query returned NODATA, but MUST NOT transmit any information that could be altered by an SVCB response until the SVCB response arrives. For example, ...(esni)... .

And then separately
> If an SVCB RR is consistent with an active or in-progress connection, the client MAY prefer that RR.

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: OK, got it.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Will fix!

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Yes, great point! I'll add that reference, and specify that cached answers are viable.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Excellent suggestion! I tweaked your proposed text and added it. Please have a look!

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,28 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+Moreover, if an A or AAAA response arrives before an SVCB response, the
+client SHOULD wait up to CD milliseconds before connecting as if the
+SVCB query returned NODATA, but MUST NOT transmit any information that
+could be altered by an SVCB response until the SVCB response arrives.
+For example, a TLS connection may be altered by the "esnikeys" value
```
bemasc: I think "ClientHello" would be clearer than "connection"

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,28 @@ in cache before performing any followup queries.
...
+of an SVCB response. CD (Connection Delay) is a configurable parameter.
+The recommended value is 50 milliseconds, as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
+
+If an SVCB record is consistent with an active or in-progress connection,
+the client may continue using that connection using any information provided
+by the SVCB RR. For example, if an SVCB record with a "esnikeys" value and
```
bemasc: I think changing between "RR" and "record" in this sentence is slightly confusing.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,28 @@ in cache before performing any followup queries.
...
+"ipv4hint" or "ipv6hint" values arrives and one of those hints matches the
+address of an in-progress connection, the TLS connection should proceed using
```
bemasc: Maybe "in-progress TCP connection"?  "proceed on that socket"?  Also might want to mention ports...

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,28 @@ in cache before performing any followup queries.
...
+the provided ESNI keying material.
+
+If the SVCB record is not consistent with any active or in-progress connection,
+e.g. there is no "ipv4hint" or "ipv6hint" that matches a connection, clients must
```
bemasc: I would prefer a definition of "consistent" where the SVCB record is also consistent with a socket if the AAAA for SvcDomainName matches the socket's IP.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,28 @@ in cache before performing any followup queries.
...
+then resolve the SvcDomainName in the SVCB response to the preferred A or AAAA
+adddress. If a client's DNS cache has valid A or AAAA records for the SvcDomainName,
```
bemasc: This requirement is already present in the definition of `ipv*hint`.  Can we find a way to avoid having to repeat it?

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: I didn't want to bubble up too much of TLS, but if you're fine with it, that works for me!

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: I'll stick with record!

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: I think a simplification is, "TLS should proceed on that connection," so we don't need to get into socket or port details. Does that work?

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Hmm... is it? The paragraph here aims to say that if the hints are *not* present, then (1) look in the cache or (2) try to resolve. The definition of `ipv*hint` doesn't seem to cover this case since it assumes they're present, right?

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: :+1:

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Agreed -- I made an attempt at adding one. Please have a look!

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,6 +627,28 @@ in cache before performing any followup queries.
...
+SVCB query returned NODATA, but MUST NOT transmit any information that
+could be altered by an SVCB response until the SVCB response arrives.
+For example, a TLS connection may be altered by the "esnikeys" value
+of an SVCB response. CD (Connection Delay) is a configurable parameter.
+The recommended value is 50 milliseconds, as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
+
+If an SVCB record is consistent with an active or in-progress connection,
+the client may continue using that connection using any information provided
+by the SVCB RR. For example, if an SVCB record with a "esnikeys" value and
+"ipv4hint" or "ipv6hint" values arrives and one of those hints matches the
```
bemasc: Maybe it would be clearer to say
> if an SVCB record arrives with an "esnikeys" value and an IP address from "ipv6hint" or an attached AAAA record that matches an in-progress connection, ...

That doesn't cover every case it but clarifies what is meant by "consistent".

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,10 +627,34 @@ in cache before performing any followup queries.
...
+client SHOULD wait up to CD milliseconds before connecting as if the
+SVCB query returned NODATA, but MUST NOT transmit any information that
+could be altered by an SVCB response until the SVCB response arrives.
+For example, a TLS ClientHello may be altered by the "esnikeys" value
+of an SVCB response {{svcparamkeys-esnikeys}}. CD (Connection Delay)
+is a configurable parameter. The recommended value is 50 milliseconds,
+as per the guidance in {{!HappyEyeballsV2=RFC8305}}.
+
+An SVCB record is consistent with an active or in-progress connection C
+if the A or AAAA address for its SvcDomainName matches the destination
```
bemasc: This definition is good but it leaves out ports and IP hints.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,10 +627,34 @@ in cache before performing any followup queries.
...
+address of C. If an SVCB record is consistent with an active or in-progress
+connection C, the client may continue using C with any information provided
+by the SVCB record. For example, if an SVCB record with a "esnikeys" value and
+"ipv4hint" or "ipv6hint" values {{svcparamkeys-iphints}} arrives, and one of
+those hints matches the address of an in-progress connection TCP connection,
+TLS should proceed on that connection using the provided ESNI keying material.
+
+If the SVCB record is not consistent with any active or in-progress connection,
```
bemasc: I don't think this paragraph is necessary.  The relevant requirements are already laid out in `{client-behavior}` and `{svcparamkeys-iphints}`.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,10 +627,34 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+Moreover, if an A or AAAA response arrives before an SVCB response, the
+client SHOULD wait up to CD milliseconds before connecting as if the
```
bemasc: The normative force here is unclear.  Is optimistic pre-connection RECOMMENDED?  What does it mean to wait for "up to CD milliseconds"?

I would prefer for optimistic pre-connection to be OPTIONAL.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: As mentioned above, I don't think `{svcparamkeys-iphints}` handles this case. (It handles the opposite one, namely, where hints are present.) I think it's important to include this case (paragraph) to be complete.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: I think this is OBE.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: SHOULD and RECOMMENDED are the same per RFC2119, right? I'm fine bumping it down to optional advice, since it truly is optional.

> What does it mean to wait for "up to CD milliseconds"?

Is "wait up to CD milliseconds before connecting" not clear? It means, "don't try and connect until CD ms go by." How can we make this more clear?

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: I can add ports, but I don't think IP hints are part of the definition. (Or, rather, they don't change things.) There's two cases for hints:

1. The hints match C's address: Continue using C's address.
2. The hints do not match C's address: The definition of the hints says that C's address should be used since it corresponds to the A/AAAA of SvcDomainName.

So I left them out and just included the port.

draft-nygren-dnsop-svcb-httpssvc.md
bemasc: "SHOULD wait ... before connecting" is ambiguous.  Is it "SHOULD wait and SHOULD connect" or "MAY connect but SHOULD wait"?  I would prefer the latter.  If indirection (non-empty SvcDomainName) turns out to be the most common reason why SVCB is slow, the success rate for this trick might be too low to justify the complexity.

It sounds like "wait *at least* CD milliseconds" would better capture your meaning.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,10 +627,39 @@ in cache before performing any followup queries.
...
+if the following conditions are met:
+
+1. The A or AAAA address for its SvcDomainName matches the destination
+address of C.
+2. The destination port of C matches that specified in the SVCB record.
+
+If an SVCB record is consistent with an active or in-progress
+connection C, the client may continue using C with any information provided
+by the SVCB record. For example, if an SVCB record with a "esnikeys" value and
+"ipv4hint" or "ipv6hint" values {{svcparamkeys-iphints}} arrives, and one of
```
bemasc: Under the current definition, this record is not consistent with this socket.  The IP hint matches, but consistency is defined in terms of the address record, not the IP hint.

I think we should fix this by changing the definition, e.g.
> An SVCB record is consistent with an active or in-progress connection if the client would attempt an equivalent connection when making use of that record.

draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -627,10 +627,39 @@ in cache before performing any followup queries.
...
+those hints matches the address of an in-progress connection TCP connection,
+TLS should proceed on that connection using the provided ESNI keying material.
+
+If the SVCB record is not consistent with any active or in-progress connection,
+clients must then resolve the SvcDomainName in the SVCB response to the preferred
```
bemasc: Could you help me understand what the new requirement is here?  Or could we simplify this to "If the SVCB record is not consistent with ... clients must proceed as described in step 3 of `{{client-behavior}}`.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: > It sounds like "wait at least CD milliseconds" would better capture your meaning.
 
Indeed it would! I'll use that.

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Simplifying like so is perfectly fine by me! I wanted to keep the blurb to close the dangling "else" case in this section. Good suggestion. :)

draft-nygren-dnsop-svcb-httpssvc.md
chris-wood: Fixed -- thanks!

bemasc:
Thanks for identifying this optimization and writing it up!

draft-nygren-dnsop-svcb-httpssvc.md
MikeBishop: @tfpauly, you can also pronounce it "Service Bee," in which case it's "a".  Ah, the linguistic implications of how we read acronyms....

Unless we're going to specify a pronunciation in the draft?  Would have headed off a whole lot of "giff/jiff" nonsense.

PR #64, opened 2019-09-27
Change the definition of AliasForm

### LINKED ISSUE #55: Clarify RRset preservation with multiple AliasForm records [Bob Harold]
The draft currently says that recursives must ignore additional AliasForm records, because only one is allowed.  (Following all aliases in parallel would allow some terrifying exponential blowup.)  We should clarify that recursives should nevertheless forward the entire RRset intact, in order to preserve DNSSEC validity.

bemasc:
This makes checking the form of an RR faster and simpler,
and ensures that sorting by priority does the right
thing.

Fixes #55

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -361,14 +361,14 @@ should be represented in wire format, using decimal escape codes
 The RDATA for the SVCB RR consists of:
 
 * a 2 octet field for SvcFieldPriority as an integer in network
-  byte order.  For AliasForm, SvcFieldPriority MUST be 0.
+  byte order.
 * the uncompressed SvcDomainName, represented as
   a sequence of length-prefixed labels as in Section 3.1 of {{!RFC1035}}.
 * the SvcFieldValue byte string, consuming the remainder of the record
   (so smaller than 65535 octets and constrained by the RDATA
   and DNS message sizes).
 
-AliasForm is defined by SvcFieldValue being empty.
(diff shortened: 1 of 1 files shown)


draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -363,14 +363,14 @@ should be represented in wire format, using decimal escape codes
...
-  byte order.  For AliasForm, SvcFieldPriority MUST be 0.
+  byte order.
 * the uncompressed SvcDomainName, represented as
   a sequence of length-prefixed labels as in Section 3.1 of {{!RFC1035}}.
 * the SvcFieldValue byte string, consuming the remainder of the record
   (so smaller than 65535 octets and constrained by the RDATA
   and DNS message sizes).
 
-AliasForm is defined by SvcFieldValue being empty.
+AliasForm is defined by SvcFieldPriority being 0.
```
enygren: Do we still need to declare that SvcFieldValue must be empty for AliasForm?  Or that clients must ignore SvcFieldValue in AliasForm?

Generally I worry a little about overloading SvcFieldPriority...

bemasc: My thought is to lean toward "clients must ignore", which is already implicit in the client behavior specification.  That would be forward-compatible if we eventually discover a use for SvcFieldValue in AliasForm.

I admit that this proposal special-cases "0", but I think "0 is special" is an ancient CS tradition.  I particularly like that clients can just sort by priority and then walk down the sorted list in a simple way.

Issue #58, opened 2019-09-26
Consider removing the in-band Alt-Svc precedence guarantee


bemasc:
Currently, in-band Alt-Svc always has precedence over HTTPSSVC, and the presence of a cached Alt-Svc is supposed to disable the HTTPSSVC query altogether.  However, this could be a problem for Chrome.

Chrome currently implements only "local Alt-Svc", i.e. Alt-Svc where the origin hostname is empty.  All other Alt-Svc headers are ignored.

Consider a site hosted in a multi-CDN configuration, using QUIC on both CDNs.  Currently, that site could send a header like `h3=:443`, and Chrome would upgrade to QUIC.

Now suppose that both CDNs support ESNI.  The site could send additional Alt-Svc headers like `h3=cdn1.example:443; esnikeys=ABC...` and `h3=cdn2.example:443; esnikeys=123...`, but Chrome would ignore both of these, so Chrome users would get QUIC but not ESNI.

If some Chrome users can query HTTPSSVC, those users could get the full package for ESNI through the DNS.  However, the presence of `h3=:443` in the Alt-Svc cache would prevent them from even doing the HTTPSSVC query (in the current draft).

A straightforward solution here would be to declare that the precedence of HTTPSSVC vs. Alt-Svc is implementation-defined.  This might reduce the ability of origins to fine-tune suboptimal HTTPSSVC, but it gives clients more flexibility to do what makes sense for them.

davidben:
Another example of the sort of complex interactions between Alt-Svc and HTTPSSVC: HTTPSSVC's http-to-https upgrade does not have an Alt-Svc analog. That means an existing Alt-Svc entry on an http origin would mask off the upgrade behavior and prevent a client from discovering it.

Alt-Svc is mostly an https-only thing, so this mostly doesn't matter, though I believe RFC8164 defines an http use. (Although clients [shouldn't implement](https://www.mozilla.org/en-US/security/advisories/mfsa2015-44/) RFC8164 anyway.)

In general, any property provided by HTTPSSVC but not Alt-Svc becomes a bit of a pickle.

MikeBishop:
@davidben, I think you're incorrect in pointing to that vulnerability as a reason not to implement RFC8164.  Mozilla had a bug, wherein certificate verification could be bypassed.  They fixed it.  That doesn't say anything about the mechanism itself, which continues to require that the certificate be checked.

davidben:
[...]

PR #66, opened 2019-10-11
Reduce the emphasis on conversion to Alt-Svc

### LINKED ISSUE #60: Reduce emphasis on HTTPSSVC <-> Alt-Svc equivalence
Given that conversion between HTTPSSVC and Alt-Svc is now potentially quite lossy (any unmapped or unrecognized keys are dropped) and has unclear security properties, we should probably de-emphasize the idea of converting between them, and focus on their parallel structure instead.

bemasc:
HTTPSSVC is less nearly equivalent to Alt-Svc than
it used to be, and conversion is not a requirement,
nor does it seem to be the most likely path to
implementation.

Fixes #60 and #58.

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -823,13 +823,18 @@ HTTPS, and receiving an HTTPSSVC record is intended to be similar to
 receiving that field value over HTTPS.  However, there are some
 differences in the intended client and server behavior.
 
-### Insecure delivery
+### Untrusted channel
 
-SVCB does not ensure secure delivery of the SvcFieldValue, so any Alt-Svc
-parameters that require secure delivery MUST NOT have a corresponding
-defined SvcParamKey.  For example, there is no SvcParamKey corresponding
-to the Alt-Svc "persist" parameter, because this parameter is not safe to
-accept over an insecure channel.
+SVCB does not require or provide any assurance of authenticity.  (DNSSEC
+signing and verification, which would provide such assurance, are OPTIONAL.)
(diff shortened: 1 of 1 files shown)


draft-nygren-dnsop-svcb-httpssvc.md
```diff
@@ -800,125 +787,54 @@ Note that none of these forms alter the HTTPS origin or authority.
...
-Server operators MAY publish multiple ServiceForm HTTPSSVC
-records as an RRSet.  When converting a collection of alt-values
-into an HTTPSSVC RRSet, the server operator MUST set the
-overall TTL to a value no larger than the minimum
-of the "max age" values (following Section 5.2 of {{!RFC2181}}).
+### Insecure delivery
 
-Each RR corresponds to exactly one alt-value, as described
-in Section 3 of {{!AltSvc}}.
+SVCB does not ensure secure delivery of the SvcFieldValue, so any Alt-Svc
```
enygren: Do we need to define "secure delivery"?  That term seems hand-wavy.

bemasc: True.  (So does "untrusted channels" below.)  I guess we're really talking about authenticity?  But maybe the thing to do is to repeat "untrusted channels" here...

enygren: Perhaps be consistent about "untrusted channels" but then define what that means somewhere, such as its first reference?

bemasc: OK, I've rephrased this to be about "untrusted channels".  Please review.

PR #67, opened 2019-10-15
Explain the need for HTTPSSVC due to wildcards

### LINKED ISSUE #59: Clarify rationale for SVCB/HTTPSSVC split
We've gotten a few questions about why these aren't a single RR type.  The text could be more explicit on the design rationale for this.

bemasc:
Fixes #59


bemasc:
Thanks for the explanation.  I've adjusted the text to include it.

PR #70, opened 2019-10-25
Clarify utility of IP hints

bemasc:


bemasc:
That's true, but I think it's probably offset by (1) the likelihood of hints being a suboptimal IP and (2) the increased response size.

I've adjusted the language to be a bit less absolute.

PR #71, opened 2019-10-28
ESNIKeys is now ESNIConfig

bemasc:
Matches the terminology from https://github.com/tlswg/draft-ietf-tls-esni/pull/187

@chris-wood FYI

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -140,9 +140,9 @@ example HTTPSSVC and associated A+AAAA records might be:
     example.com.      7200  IN HTTPSSVC 0 svc.example.net.
     ; ServiceForm
     svc.example.net.  7200  IN HTTPSSVC 2 svc3.example.net. ( alpn=h3
-                                        port=8003 esnikeys="..." )
+                                        port=8003 esniconfig="..." )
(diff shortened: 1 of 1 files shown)


chris-wood:
LGTM -- thanks, @bemasc!

bemasc:
It looks like that PR hasn't landed yet, so this one should probably wait until it's merged.

chris-wood:
@bemasc the ESNI draft PR has been merged, so this one should be good to go. One thing we might want to consider is the option for the field to represent a set of configurations, so perhaps we can rename this to "esniconfigs" and note that it's the concatenation of one or more ESNIConfig structures?

bemasc:
Thanks for the reviews.

> unless we want to just call it "esni" ?

I would be OK with renaming to "esni", but I do wonder if it could be confusing, because the contents of this field are not the "encrypted SNI".  I would defer to the ESNI authors on preferred terminology.

> perhaps we can rename this to "esniconfigs" and note that it's the concatenation of one or more ESNIConfig structures

I would be OK with this, but it seems like a nontrivial change so I'm not going to attempt it in this PR.  I would also want to think a bit about what the use case is.  It occurs to me that a client with multiple `ESNIConfig`s could potentially have some kind of fallback behavior (but how does that interact with `esni_retry_request`?) or it could even put multiple `encrypted_server_name` extensions in the ClientHello.

PR #72, opened 2019-11-04
Clean up handling of the "http" scheme

bemasc:


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -792,15 +792,19 @@ will implement both, and a partial mapping exists between them
...
-For origins other than https with port 443 and http with port 80,
+For HTTPS origins with ports other than 443,
 the port and scheme continue to be prefixed to the hostname
 as described in {{svcb-names}}.  Following of HTTPSSVC AliasForm and
 CNAME aliases is also unchanged from SVCB.
 
+Clients always convert the "http" scheme to "https" before performing an
+HTTPSSVC query ({{hsts}}), so domain owners MUST NOT publish HTTPSSVC
+records with a prefix of "_http".
+
```
enygren: What if a URL is:
http://www.example.com:80/foo.html
do we drop/ignore the ":80" ?  I'd assume so and this might be good to call out here explicitly.

bemasc: The `{{hsts}}` link points to the section on "HTTP Strict Transport Security", which says "If the "http" URL explicitly specifies port 80, specify port 443.".  Do you think we need to repeat that instruction here?

enygren: Seems safer to include?  (perhaps copy the text and reference back.)

bemasc: I'd like to avoid repeating the URL conversion algorithm, but I've rephrased this to avoid implying that the conversion is a simple replacement of the scheme.

PR #76, opened 2019-11-13
Remove recommendation to shorten the TTL

### LINKED ISSUE #74: Remove TTL shortening recommendation
The current text recommends shortening the TTL to compensate for misbehaving clients.  This is not a good recommendation, since some clients will just ignore the shorter TTL.  Instead, we should just mention that servers cannot rely on prompt expiration.

Credit: @puneetsood

bemasc:
Shortening the TTL doesn't actually solve the problem,
because some caches ignore small TTL values.  Regardless of TTL,
servers have to be able to deal with occasional use of expired
records.  Shortening TTL is also bad for efficiency, so we
shouldn't encourage it.

Fixes #74

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -851,8 +851,8 @@ HTTPSSVC records expiring on time.  Shortening the TTL to compensate
 for incorrect caching is NOT RECOMMENDED, as this practice impairs the
 performance of correctly functioning caches and does not guarantee
 faster expiration from incorrect caches.  Instead, server operators
-SHOULD continue operating the endpoints in expired records until they
-observe that nearly all connections have migrated to the new endpoints.
+SHOULD maintain compatibility with expired records until they observe
+that nearly all connections have migrated to the new configuration.
 
 Sending Alt-Svc over HTTP allows the server to tailor the Alt-Svc
 Field Value specifically to the client.  When using an HTTPSSVC DNS
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -843,9 +843,11 @@ this parameter is not safe to accept over an untrusted channel.
 There is no SvcParamKey corresponding to the Alt-Svc "ma" (max age) parameter.
 Instead, server operators SHOULD encode the expiration time in the DNS TTL.
 
-Some DNS caching systems incorrectly extend the lifetime of DNS
-records beyond the stated TTL.  Server operators MUST NOT rely on
-HTTPSSVC records expiring on time, and MAY shorten the TTL to compensate.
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -843,9 +843,11 @@ this parameter is not safe to accept over an untrusted channel.
 There is no SvcParamKey corresponding to the Alt-Svc "ma" (max age) parameter.
 Instead, server operators SHOULD encode the expiration time in the DNS TTL.
 
-Some DNS caching systems incorrectly extend the lifetime of DNS
-records beyond the stated TTL.  Server operators MUST NOT rely on
-HTTPSSVC records expiring on time, and MAY shorten the TTL to compensate.
+The appropriate TTL value will typically be similar to the "ma" value
+used for Alt-Svc, but may vary depending on the desired efficiency and
+agility.  Some DNS caching systems incorrectly extend the lifetime of DNS
+records beyond the stated TTL, so server operators MUST NOT rely on
```
bemasc: "MUST NOT" seems wrong here.  Maybe this can be non-normative operational guidance.

enygren: Agreed, especially as ESNI key rotations need to assume that clients will at least eventually re-resolve and not extend the lifetime indefinitely.  Perhaps "[...], so server operators should assume that clients will continue to use records for some amount of time past their expiration."   ?

bemasc: I've adjusted the advice here.  PTAL.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -843,9 +843,16 @@ this parameter is not safe to accept over an untrusted channel.
...
+The appropriate TTL value will typically be similar to the "ma" value
+used for Alt-Svc, but may vary depending on the desired efficiency and
+agility.  Some DNS caches incorrectly extend the lifetime of DNS
+records beyond the stated TTL, so server operators cannot rely on
+HTTPSSVC records expiring on time.  Shortening the TTL to compensate
+for incorrect caching is NOT RECOMMENDED, as this practice impairs the
+performance of correctly functioning caches and does not guarantee
+faster expiration from incorrect caches.  Instead, server operators
+SHOULD continue operating the endpoints in expired records until they
+observe that nearly all connections have migrated to the new endpoints.
```
enygren: Do we want to call out ESNI keys as well?

"Instead, server operators SHOULD continue operating the endpoints listed in expired records until they
observe that nearly all existing and new connections have migrated to the new endpoints (and are using new ESNI keys if applicable)."

bemasc: I tried to capture this by changing the text to describe "configurations" instead of "endpoints".

Issue #40, opened 2019-07-09
HTTPSSVC: Comments from Brian Dickson


enygren:
One minor issue is that wherever CNAME is referenced, you probably want to also include a reference to DNAME, including any implied or explicit chaining of CNAMEs (which could be sequences of CNAME and/or DNAME modulo their respective behavior.)

You might also want to explain the motivation for keeping the FQDN separate from the alt-svc parameters (to make it trivial to parse, and thus to do DNS substitutions like CNAME/DNAME). It is there, just not as up-front as it could be.

It might be a little clearer if the list of alt-svc values (h2, h3, etc) that can occur were to be listed in the document. In particular, the association between h3 and QUIC is inferred but not explicitly called out (at least not that I noticed.)

enygren:
These are not yet addressed in the proposed draft-nygren-dnsop-svcb-httpssvc-00.
We should include in the next version.

bemasc:
> It might be a little clearer if the list of alt-svc values (h2, h3, etc) that can occur were to be listed in the document.

The values are in an IANA registry, and I don't think the "alpn" SVCB parameter is intended to be restricted to a subset of those values.  However, it is true that only some of those values make sense in the context of HTTPSSVC, and their meaning is indeed nontrivial.  I don't think we could enumerate them in the text in any permanent way, but I suppose we could explain the present state of affairs more explicitly.

PR #77, opened 2019-11-13
Clarify various recommendations

### LINKED ISSUE #75: Clarity improvements
* Disentangle description of recursive and authoritative server behavior in section 4
* Explain why alias and service forms need to share an RR type
* Clarify interaction between CNAME and AliasForm in Section 4
* State clearly that only A and AAAA records will ever need to be added to the response

bemasc:
Fixes #75 and addresses comments in #40

@puneetsood FYI

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -216,19 +216,21 @@ a non-normative manner.  (As mentioned above, this all
 applies equally to the HTTPSSVC RR which shares
 the same encoding, format, and high-level semantics.)
 
-The SVCB RR has two forms: AliasForm and ServiceForm.
-SVCB RR entries with two non-empty fields are in AliasForm.
-When more fields are present, this indicates that the SVCB RR
-is in ServiceForm.  The fields are:
+The SVCB RR has two forms: AliasForm, which aliases a name to another name,
+and ServiceForm, which provides connection information bound to a service
+endpoint domain.  Placing both forms in a single RR type allows clients to
+fetch the relevant information with a single query.
+
+The SVCB RR has two mandatory fields and one optional.  The fields are:
 
 1. SvcFieldPriority: The priority of this record (relative to others,
-   with lower values preferred).  Applicable for the ServiceForm,
-   and otherwise has value "0".  (Described in {{pri}}.)
+   with lower values preferred).  A value of 0 indicates AliasForm.
+   (Described in {{pri}}.)
 2. SvcDomainName: The domain name of either the alias target (for
    AliasForm) or the alternative service endpoint (for ServiceForm).
-3. SvcFieldValue: A list of key=value pairs
+3. SvcFieldValue (optional): A list of key=value pairs
    describing the alternative service endpoint for the domain name specified in
-   SvcDomainName (only for ServiceForm and otherwise empty).
+   SvcDomainName (only used in ServiceForm and otherwise ignored).
    Described in {{svcfieldvalue}}.
 
 Cooperating DNS recursive resolvers will perform subsequent record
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -616,7 +631,7 @@ on its contents.
 When responding to a query that includes the DNSSEC OK bit ({{!RFC3225}}),
 DNSSEC-capable recursive and authoritative DNS servers MUST accompany
 each RRSet in the Additional section with the same DNSSEC-related records
-that it would send when providing that RRSet as an Answer.
+that it would send when providing that RRSet as an Answer (e.g. RRSIG, NSEC3).
```
puneetsood: RRSIG, NSEC and NSEC3

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -604,10 +610,19 @@ to a stub resolver for an SVCB record query:
...
+   Otherwise, go to step 1, subject to loop detection heuristics.
+
+In steps 2 and 3, "resolve" means the resolver's ordinary iterative
+resolution procedure, as if processing a query for that RRSet.
+This includes following any aliases that the resolver would ordinarily
+follow (e.g. CNAME, {{!DNAME}}).  If a queried record does not exist, the
+resolver SHOULD add to the Additional section any records that it would
+normally include in a negative response (e.g. CNAME, SOA), in order to
+enable negative caching by the stub.
+
```
enygren: Is the negative behavior actually possible?    (Is it possible to add an empty record additionals?  I'm forgetting whether this is valid, but I can't recall seeing it done?  It may be that neither is possible?)    Also, is there a distinction to be made here between NODATA/NOERROR (the RRType doesn't exist) and NXDOMAIN?  Certainly, I don't believe the latter can be added to the Additional section?

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -604,10 +610,19 @@ to a stub resolver for an SVCB record query:
...
 3. Otherwise, select an AliasForm record at random, and resolve A, AAAA,
-   and SVCB records for
-   the SvcDomainName.  If the SVCB record does not exist, add the A and AAAA
-   records to the Additional section.  Otherwise, go to step 1,
-   subject to loop detection heuristics.
+   and SVCB records for the SvcDomainName.  If the SVCB record does not exist,
+   add all records from the A and AAAA responses to the Additional section.
+   Otherwise, go to step 1, subject to loop detection heuristics.
+
+In steps 2 and 3, "resolve" means the resolver's ordinary iterative
```
enygren: "iterative" or "recursive" ?  I typically see the latter used.

bemasc: I think you're right.  Changed to "recursive".

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: > Is the negative behavior actually possible?

I think it might be possible, but it's definitely not straightforward.  It would require the client to recapitulate the recursive resolution procedure, notice the absence of certain RRSets, and check for a corresponding SOA instead.  It's more obviously possible (but even more complicated) for signed zones.

Anyway, I've significantly reworked this section to try to pin down the precise behavior of which records go where, what RCODE to return, etc.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -586,28 +588,51 @@ Providing the proxy with the final SvcDomainName has several benefits:
...
 execute the procedure in {{client-behavior}} without issuing a second
-round of queries, by following this procedure while constructing a response
-to a stub resolver for an SVCB record query:
+round of queries, by incorporating all the necessary information into a
+single response.  For the initial SVCB record query, this is just the normal
+response construction process (i.e. unknown RR type resolution under
+{{!RFC3597}}).  For followup resolutions performed during this procedure,
+we define incorporation as adding all Answer and Additional RRs to the
+Additional section, and all Authority RRs to the Authority section,
+without altering the response code.
```
enygren: Perhaps add:

"Recursive resolvers MAY respond to the SVCB Answer with 
only a subset of Additional RRs when the response would
be larger than is permitted or when responding before fully chasing 
dependencies would improve performance.  Additional information from
higher priority ServiceForm records SHOULD be prioritized over lower priority
ServiceForm records."

bemasc: OK.  I added this to the Performance Optimizations section.

Issue #84, opened 2019-11-25
Add a simple example


bemasc:
After looking at the examples, several readers came away with the impression that an alias is required to use SVCB.  In fact, aliasing is bad for performance and should be avoided when possible, but we've included it in all of our examples to show how all the pieces of the system work.

We should include an optimized example, and make it clear that many common cases can be handled without adding any indirection.

MikeBishop:
This should be resolved now.

PR #85, opened 2019-11-25
Clarify examples

### LINKED ISSUE #83: Grammar for presentation format does not produce parentheses
Examples like this one (from section 1.1) use a notation where the SvcFieldValue is surrounded by parentheses.
```
   ; ServiceForm
   svc.example.net.  7200  IN HTTPSSVC 2 svc3.example.net. ( alpn=h3
                                       port=8003 esniconfig="..." )
```

It looks like the top-level production rule is this:
```
   pair          = display-key "=" value
```

Shouldn't there be a rule just one level higher? Something like this?
```
   svc-field-value = [ "(" 1*pair ")" ]
```

bemasc:
Fixes #83, #84

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -130,44 +130,68 @@ Perhaps this could be included as a svc="baz" parameter
 for protocols that are not the default for the RR type?
 Current inclination is to not do so.
 
-## Introductory Example
+## Example: Protocol enhancements
 
-As an introductory example for an HTTPS origin resource, a set of
-example HTTPSSVC and associated A+AAAA records might be:
+Consider a simple zone of the form
 
-    www.example.com.  7200  IN CNAME    svc.example.net.
-    ; AliasForm
-    example.com.      7200  IN HTTPSSVC 0 svc.example.net.
-    ; ServiceForm
-    svc.example.net.  7200  IN HTTPSSVC 2 svc3.example.net. ( alpn=h3
-                                        port=8003 esniconfig="..." )
-    svc.example.net.  7200  IN HTTPSSVC 3 svc2.example.net. ( alpn=h2
-                                        port=8002 esniconfig="..." )
-    svc2.example.net. 300   IN A        192.0.2.2
-    svc2.example.net. 300   IN AAAA     2001:db8::2
-    svc3.example.net. 300   IN A        192.0.2.3
-    svc3.example.net. 300   IN AAAA     2001:db8::3
-    ; Compatibility records for non-HTTPSSVC-aware clients
-    example.com.      300   IN A        192.0.2.1
-    example.com.      300   IN AAAA     2001:db8::1
-    svc.example.net.  300   IN A        192.0.2.1
-    svc.example.net.  300   IN AAAA     2001:db8::1
+    simple.example. 300 IN A    192.0.2.1
+                           AAAA 2001:db8::1
 
-In the preceding example, both of the "example.com" and
-"www.example.com" origin names are aliased to use alternative service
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -130,51 +130,77 @@ Perhaps this could be included as a svc="baz" parameter
...
-continue to treat the authoritative origins as "https://example.com"
-and "https://www.example.com", respectively.
-
-For services other than HTTPS (as well as for HTTPS origins
-with non-default ports), the SVCB RR and an {{?Attrleaf}} label will be used.
-For example, to reach an example resource of
+    simple.example. 300 IN A    192.0.2.1
+                           AAAA 2001:db8::1
+
+The domain owner could publish new records like
```
enygren: replace "new" with "added" or "additional".  ("new" might imply that they replace the A and AAAA records.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -130,51 +130,77 @@ Perhaps this could be included as a svc="baz" parameter
...
+    pool  300 IN A        192.0.2.2
+                 AAAA     2001:db8::2
+    h3    300 IN A        192.0.2.3
+                 AAAA     2001:db8::3
+
+This configuration is entirely compatible with the "Apex aliasing" example,
+whether the client supports HTTPSSVC or not.  If the client does support
+HTTPSSVC, all connections will be upgraded to HTTPS, and clients will
+use HTTP/3 if they can.  Parameters are "bound" to each server pool, so
+each server pool can have its own protocol, ESNI configuration, etc.
```
enygren: Perhaps explicitly call out that pool.svc.example. may be either the target of a CNAME chain or an alias?
For example:

$ORIGIN example.com.
www    3600 IN CNAME pool.svc.example.
 .          3600 IN HTTPSSVC 0 pool.svc.example.

enygren: It would also be good to clarify that neither form impacts the domain name in the origin.
For example:
------

Similar to CNAME, HTTPSSVC aliases have no impact on the origin name.  When connecting, clients will continue to treat the authoritative origins as "https://www.example.com" and "https://example.com", respectively, and will validate TLS server certificates accordingly.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -130,51 +130,77 @@ Perhaps this could be included as a svc="baz" parameter
...
+Suppose that svc.example's default server pool supports HTTP/2, and
+it has deployed HTTP/3 on a new server pool with a different
+configuration.  This can be expressed in the following form:
+
+    $ORIGIN svc.example. ; A hosting provider.
+    pool 7200 IN HTTPSSVC 1 h3 alpn=h3 esniconfig="123..."
+                 HTTPSSVC 2 .  alpn=h2 esniconfig="abc..."
+    pool  300 IN A        192.0.2.2
+                 AAAA     2001:db8::2
+    h3    300 IN A        192.0.2.3
```
enygren: Might be less confusing if we called it "h3pool" ?

draft-ietf-dnsop-svcb-httpssvc.md

draft-ietf-dnsop-svcb-httpssvc.md

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: > Perhaps explicitly call out that pool.svc.example. may be either the target of a CNAME chain or an alias?

This is already in the preceding example.

> It would also be good to clarify that neither form impacts the domain name in the origin.

Added.

PR #89, opened 2019-12-06
Replace "alpn" with "proto"/"no-proto"

### LINKED ISSUE #73: Redesign ALPN mismatch handling
Currently, if a server publishes a QUIC-only HTTPSSVC RRSet with ESNI, there is no way for a client to fall back to a non-QUIC connection, because doing so would reveal the SNI.  This could increase the likelihood of partial outages for server admins who haven't considered the small fraction of users whose network path does not support QUIC.

Reviewers have reported concerns that this creates an undesirable level of fragility.  We should consider whether there is an alternative design that would be less likely to result in accidental breakage.

bemasc:
This change
1. Makes it difficult for admins to accidentally drop support for HTTP/1.1 and HTTP/2
2. Makes it easy for clients to determine the precise set of a protocols supported by an endpoint before attempting a connection.

Fixes #73

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -693,19 +693,14 @@ The presentation format and wire format of SvcParamValue
 is its registered "Identification Sequence".  This key SHALL NOT
 appear more than once in a SvcFieldValue.
 
-Clients MUST include this value in the ProtocolNameList in their
-ClientHello's `application_layer_protocol_negotiation` extension.
-Clients SHOULD also include any other values that they support and
-could negotiate on that connection with equivalent or better security
-properties.  For example, when using a SvcFieldValue with an "alpn" of
-"h2", the client MAY also include "http/1.1" in the ProtocolNameList.
+The presence of a protocol ID in SvcParamKeys indicates that this
+service endpoint supports the named protocol ID, and is able to negotiate
+it.  For example, an "alpn" value of "h3" ({{HTTP3}} Section 11.1)
+indicates the endpoint supports QUIC, because that is the only transport
+protocol over which HTTP/3 can be negotiated.
 
-Clients MUST ignore SVCB RRs where the "alpn" SvcParamValue
-is unknown or not supported for use with the current scheme.
-
-The value of the "alpn" SvcParamKey can have effects beyond the content
-of the TLS handshake and stream.  For example, an "alpn" value of "h3"
-({{HTTP3}} Section 11.1) indicates the client must use QUIC, not TLS.
+The service endpoint MAY also support protocol IDs that are not listed in
+SvcParamKeys.
 
 ## "port"
 
@@ -790,7 +785,8 @@ All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPSSVC, and the "alpn" SvcParamKey is REQUIRED.  Its value MUST
 be an ALPN that is suitable for use with HTTPS.  For example, the
 value MAY be "http/1.1", "h2", or "h3", but MUST NOT be "h2c" or
-"ftp".
+"ftp".  Regardless of the "alpn" SvcParamValue, all service endpoints
+MUST support HTTP/1.1.
 
 The presence of an HTTPSSVC record for an HTTP or HTTPS service also
 provides an indication that all resources are available over HTTPS, as
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -139,9 +139,9 @@ example HTTPSSVC and associated A+AAAA records might be:
     ; AliasForm
     example.com.      7200  IN HTTPSSVC 0 svc.example.net.
     ; ServiceForm
-    svc.example.net.  7200  IN HTTPSSVC 2 svc3.example.net. ( alpn=h3
+    svc.example.net.  7200  IN HTTPSSVC 2 svc3.example.net. ( proto=h3
                                         port=8003 esniconfig="..." )
-    svc.example.net.  7200  IN HTTPSSVC 3 svc2.example.net. ( alpn=h2
+    svc.example.net.  7200  IN HTTPSSVC 3 svc2.example.net. (
                                         port=8002 esniconfig="..." )
     svc2.example.net. 300   IN A        192.0.2.2
     svc2.example.net. 300   IN AAAA     2001:db8::2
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -790,7 +785,8 @@ All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPSSVC, and the "alpn" SvcParamKey is REQUIRED.  Its value MUST
 be an ALPN that is suitable for use with HTTPS.  For example, the
 value MAY be "http/1.1", "h2", or "h3", but MUST NOT be "h2c" or
-"ftp".
+"ftp".  Regardless of the "alpn" SvcParamValue, all service endpoints
+MUST support HTTP/1.1.
```
enygren: What about DoH that mandates H2 support?  It also is entirely reasonable for service endpoints to support QUIC and not H2 or HTTP/1.1.  There are features of QUIC that make it attractive to have it on separate IPs that are QUIC-only.

Special-casing http/1.1 seems reasonable, but I'm not sure this is the easiest way.  
Perhaps:
A service endpoint supporting "h2" SHOULD also support "http/1.1" unless either no service endpoints support "http/1.1" or a distinct service endpoint is specified for "http/1.1"

?

bemasc: > What about DoH that mandates H2 support?

DoH doesn't recommend refusing connections over HTTP/1.1, so I don't think there's a conflict.

> There are features of QUIC that make it attractive to have it on separate IPs that are QUIC-only.

I'd be really interested to know more about this use case.  I've been hearing concerns from browser people that support for QUIC-only endpoints may be difficult to implement, because it complicates TCP fallback when UDP is failing.  I'm hoping that we can find a middle ground that addresses both needs.  (Feel free to switch to email if it's too long for this comment thread.)

bemasc:
@enygren @MikeBishop I've updated this PR to reflect our conversation today.  PTAL.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1024,8 +1036,8 @@ be populated with the registrations below:
 
 | SvcParamKey | NAME        | Meaning                       | Reference       |
 | ----------- | ------      | ----------------------        | --------------- |
-| 0           | key0        | Reserved                      | (This document) |
-| 1           | alpn        | ALPN for alternative service  | (This document) |
+| 0           | proto       | Supported protocol            | (This document) |
+| 1           | no-proto    | Unsupported protocol          | (This document) |
```
MikeBishop: This could go either way, but since there are existing implementations with supported protocols at codepoint 1, wouldn't it make sense to put `no-proto` at 0 instead?

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1125,11 +1138,14 @@ intends to include an HTTP response header like
...
     www.example.com. 3600 IN HTTPSSVC 2 svc.example.net. (
-                                        alpn=h3 port=8003 foo=123 )
+                                        proto=h3 port=8003 foo=123 )
                              HTTPSSVC 3 svc.example.net. (
-                                        alpn=h2 port=8002 foo=123 )
+                                        port=8002 foo=123 )
 
 Where "foo" is a hypothetical future HTTPSSVC and Alt-Svc parameter.
+(Note that unlike the Alt-Svc values, these records also indicate support
+for HTTP/1.1 and HTTP/2 on both ports.  This indication could be removed
```
MikeBishop: Slight nitpick -- TCP 8003 is not the same point as UDP 8003, so "both ports" might not be applicable.  This RRSET makes assertions about *three* ports.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1175,16 +1187,18 @@ The following:
...
+        proto=h3 port=8003 esniconfig="ABC..." )
     svc.example.net.  7200  IN HTTPSSVC 3 . (
-        alpn=h2 port=8002 esniconfig="123..." )
+        port=8002 esniconfig="123..." )
 
-is equivalent to the Alt-Svc record:
+is similar to the Alt-Svc record:
 
     Alt-Svc: h3="svc3.example.net:8003"; esniconfig="ABC..."; ma=7200, \
              h2="svc.example.net:8002"; esniconfig="123..."; ma=7200
```
MikeBishop: Also:

- h2="svc3.example.net:8003"; esniconfig="ABC..."; ma=7200,
- http/1.1="svc3.example.net:8003"; esniconfig="ABC..."; ma=7200,
- http/1.1="svc.example.net:8002"; esniconfig="123..."; ma=7200

enygren:
I am liking the direction of this, but this is a broad-reaching enough change that we should expand this proposal/discussion to include a larger audience when we're ready.  A bunch of comments and questions:

* Can we name this something easier to write?
  The usability of "additional-transport" seems poor.
  Also, it is bikeshedding, but having a good name that
  these tokens and their registry can be referred
  to with and distinguished from ALPN seems critical.
  (So far "TP" is the best I've come up with as writing
  tp=tls is more readable.)

* I'm not sure all encodings / data types have
  an empty value as meaningful.

* We should talk about the relationship of {additional-transport /
  transport protocol} to ALPN.  If we are going to go this path, this
  seems like a broader conversation to have that we should loop people
  like Martin Thompson and Tommy Pauly into as well.  (There may be a
  question as to whether there is value to also supporting ALPN, but
  that may be unneeded complexity.)  How this relates to Alt-Svc is
  something that should likely have discussion in the HTTP and QUIC
  (and perhaps TAPS?)  working groups as well once there is a concrete
  proposal.  A question may be whether this should get factored into
  its own document as this is a significant enough change to Alt-Svc.

  One way to frame the difference may be that:
  - $TP (whatever we call it) defines a transport layer
    or transport+security layer protocol (eg, TCP+TLS, UDP+QUIC).
    Each of these provide a way to communicate over the network
    in a way that a secure connection can be negotiated and
    initiated.
  - ALPN is an aggreement as part of the $TP negotiation
    to agree on the upper application-layer protocol
    (eg, HTTP-over-HTTP/1.1 vs HTTP-over-HTTP/2).

  Closely related to this is the relation of the "transport protocol
  set" with ALPN values.  (ie, are there cases where the client
  needs to be aware which ALPN protocols within the transport
  protocol an endpoint supports prior to connecting?)

* Does it make sense that the transport registry is HTTPS-specific
  or should a single registry be shared across HTTPS
  and other protocols?  I'd be inclined towards a single registry,
  especially given the uniqueness properites with ALPN.
  Having a bunch of registries all sharing a key space
  sounds like a disaster.

* There are legitimate use-cases for no-default-transport
  outside ot web browsers.  It would be entirely reasonable
  for web services in a closed ecosystem to only support HTTP/3.

* Do we need to update something to indicate that
  the ALPN registry also doesn't collide with this?
  Does IANA have a precedent for this?  Would merging
  these into a single registry with a new column (for if
  it is a $TP or an ALPN) be easier?

* We may want to be clear in our initial registry entries
  that these are "TLS over TCP" and "QUIC over UDP" ?

* We should provide guidance for how servers make
  a transition for Alt-Svc from alpn to these additional-transports.
  (ie, recommend that they include both, and we should probably
  show this in our example.)

* For "tls", what ALPN protocols does a server need to support?
  Could a server be HTTP/2 only or HTTP/1.1 only?  I guess
  this will get figured out during the TLS handshake negotiation?
  (How this relates to Alt-Svc is also important.)

enygren:
(oops, added comment to the wrong issue)

enygren:
Documenting here that I think we're leaning in the direction of transport/no-default-transport and will reject this alternate PR?

PR #93, opened 2019-12-16
Make HTTPSSVC imply fatal certificate errors

### LINKED ISSUE #87: Should the presence of an HTTPS record suppress any certificate warning bypass option?
Elsewhere, HTTPSSVC's redirect was referred to as HSTS and it occurred to me we're missing one of HSTS's properties. It redirects and then it directs the browser to suppress the certificate click through button.

We could do something similar and say that HTTPS connections made off an HTTPSSVC record are assumed to have a competent TLS config and don't get a bypass button.

davidben:
This is the other half of HSTS. Closes #87.

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -919,7 +919,7 @@ it SHOULD abandon the connection attempt if the HTTPSSVC response fails
 to validate.
 
 Finally, when making an "https" scheme request to an origin with an HTTPSSVC
-record, either directly or via the above redirect, client SHOULD terminate the
+record, either directly or via the above redirect, the client SHOULD terminate the
 connection if there are any errors with the underlying secure transport, such as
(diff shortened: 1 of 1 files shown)


davidben:
HSTS uses MUST instead of SHOULD, but I've left it as SHOULD to align with the rest of HTTPSSVC.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -918,6 +918,12 @@ Similarly, if the client enforces DNSSEC validation on A/AAAA responses,
 it SHOULD abandon the connection attempt if the HTTPSSVC response fails
 to validate.
 
+Finally, when making an "https" scheme request to an origin with an HTTPSSVC
+record, either directly or via the above redirect, client SHOULD terminate the
```
bemasc: Nit: "the client"

davidben: Fixed.

PR #94, opened 2019-12-17
Do not allow duplicate ipv4hint and ipv6hint parameters

davidben:
The contents already encode multiple addresses. We shouldn't have multiple ways to say the same thing.

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -750,9 +750,6 @@ The wire format for each parameter is a sequence of IP addresses in network
 byte order.  Like an A or AAAA RRSet, the list of addresses represents an
 unordered collection, and clients SHOULD pick addresses to use in a random order.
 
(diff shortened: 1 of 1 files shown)


davidben:
@dmcardle FYI

bemasc:
@enygren I believe you added this language in https://github.com/MikeBishop/dns-alt-svc/commit/9e1726cd142d32dde32d18d5c86db16f2e60d61a.  Do you think it's still needed?

enygren:
I think if we address issue #95 and just recommend using a multi-value parameter (and have very clear specification that duplicate parameters must ignored in a deterministic manner) that this text isn't needed as-is.

enygren:
Should we replace this with text elsewhere prohibiting the same parameter key from being used multiple times and specifying that one (first? last?) should be deterministically selected?

enygren:
(so pend this on #95 and close/wontfix if we merge that, or approve if we reject #95)

bemasc:
This has been merged in #109.

Issue #101, opened 2020-01-15
Include Alt-Used example


enygren:
It would be good to include an example for Alt-Used (or mention it in a previous example).

The {client-behavior} reference is also broken.

enygren:
WONTFIX (alternative is #107)

PR #102, opened 2020-01-15
Alt-Used example

enygren:
* Add an example for Alt-Used per issue #101  (although exact details may need to change depending on other PRs around alpn vs transport).
* Fix reference to client-behavior

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -851,7 +851,7 @@ For example, with records of:
                  HTTPSSVC 2 . transport=tls
                  
 The "Alt-Used" header for requests to all of https://example.com
-and https://www.example.com and https://pool.svc.example
+and https://www.example.com (and https://pool.svc.example if it exists)
 would be one of:
 
     Alt-Used: h3pool.svc.example:8443
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -838,9 +838,26 @@ The header's value (in ABNF) SHOULD be
 
     uri-host ":" port
 
-where uri-host is the final value of HOST ({client-behavior}) minus
+where uri-host is the final value of HOST ({{client-behavior}}) minus
 the trailing ".", and port is the port number in use.
 
+For example, with records of:
+
+    $ORIGIN example.com.
+    .   7200 IN HTTPSSVC 0 pool.svc.example.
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -838,9 +838,26 @@ The header's value (in ABNF) SHOULD be
...
+
+    $ORIGIN example.com.
+    .   7200 IN HTTPSSVC 0 pool.svc.example.
+    www 7200 IN CNAME pool.svc.example.
+    $ORIGIN svc.example. ; A hosting provider.
+    pool 7200 IN HTTPSSVC 1 h3pool transport=quic port=8443
+                 HTTPSSVC 2 . transport=tls
+                 
+The "Alt-Used" header for requests to all of https://example.com
+and https://www.example.com and https://pool.svc.example
```
bemasc: I would suggest "(and https://pool.svc.example, if it exists)", since there's no particular need for that origin to exist.

davidben:
Sending this header has some interesting tracking consequences. Your DNS resolver can unilaterally send a common cross-site identifier to every site by faking an `Alt-Svc` name. The DNS resolver would answer for that name too, so it doesn't matter if the name is fake. More generally, this is feeding an insecure value into a header in a secure context, which is odd.

(Chrome doesn't send `Alt-Used` for Alt-Svc either, though I don't know what the reasoning was at the time.)

MikeBishop:
In Alt-Svc days, the argument was that the issuer of Alt-Svc could produce tracking anyway -- they give you a name that resolves to an IPv6 address with the host portion being the identifier they want to use, then serve over the whole prefix.  The argument was that Alt-Used didn't give them any additional information versus that structure.  (However, it permits it to work over IPv4, which seems like a sufficient increase in capability to warrant some caution.)

However, that requires control of the actual IP address.  The recipient of Alt-Used need not be the issuer of the Alt-Svc record if the alternative can come from DNS.  Seems like another incarnation of these records not being authenticated as being from the affected domain.

bemasc:
Per #107, it looks like we're going to try a different direction for this. @enygren, would you like to close this PR?

enygren:
Abandoning this for now, pending discussion of issue 107.

bagder:
Not exactly the right place for this but note that `Alt-Used` is slightly flawed in that it doesn't mention protocol/alpn so when we switch *nothing but* h2 to h3 to the same host name and port number, Cloudflare for example will wrongly think it is a loop.

PR #104, opened 2020-01-16
Generalize SERVFAIL handling for security

### LINKED ISSUE #103: Expand SERVFAIL handling to prevent ESNI DoS
The HSTS section currently says
> If the HTTPSSVC query results in a SERVFAIL error, and the connection between the client and the recursive resolver is cryptographically protected (e.g. using TLS {{!RFC7858}} or HTTPS {{!RFC8484}}), the client SHOULD abandon the connection attempt and display an error message.

This is meant to defend against DoS on the recursive-authoritative leg, when the zone is DNSSEC-signed and the recursive is validating.

We should generalize this defense to include timeouts and transport errors, not just explicit SERVFAIL, and move it out of the HSTS section, because (1) this DoS also applies to ESNI, not just HTTPS-upgrade, and (2) an attacker might be able to mount this attack on the client-recursive leg by selectively dropping DNS query or response packets (e.g. only dropping the largest responses, so the client gets the A record but not the HTTPSSVC record).

bemasc:
Fixes #103

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -595,9 +595,9 @@ to avoid additional latency in comparison to ordinary AAAA/A lookups.
 If an SVCB query results in a SERVFAIL error, transport error, or timeout,
 and DNS exchanges between the client and the recursive resolver are
 cryptographically protected (e.g. using TLS {{!RFC7858}} or HTTPS
-{{!RFC8484}}), the client MUST abandon the connection attempt and display
-an error message.  This ensures that an active attacker cannot mount a
-downgrade attack by deny the user access to the SVCB information.
+{{!RFC8484}}), the client MUST NOT fall back to non-SVCB connection
+establishment.  This ensures that an active attacker cannot mount a
+downgrade attack by denying the user access to the SVCB information.
 
 A SERVFAIL error can occur if the domain is DNSSEC-signed, the recursive
 resolver is DNSSEC-validating, and the attacker is between the recursive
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -590,6 +590,26 @@ such as {{!HappyEyeballsV2=RFC8305}}.
 Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
+## Handling resolution failures
+
+If an SVCB query results in a SERVFAIL error, transport error, or timeout,
+and DNS exchanges between the client and the recursive resolver are
+cryptographically protected (e.g. using TLS {{!RFC7858}} or HTTPS
+{{!RFC8484}}), the client MUST abandon the connection attempt and display
```
enygren: Can the client use a previous stale answer if it has one in-cache still?

enygren: The client may also wish to retry the query against another equally-trusted resolver?  (eg, if it has multiple resolver IPs.)

bemasc: Good point, we don't necessarily want to rule out those options.  I've changed the text from "MUST abandon" to "MUST NOT fall back".

davidben:
This is to address the attack we discussed where the attacker drops QUIC packets to time out the HTTPSSVC response, right? I think the rationale here should be in the draft, in case this fix fails. E.g., if many authoritative servers time out unexpected record types, HTTPSSVC would break sites and thus this would not be viable.

Having the reasoning documented means we won't forget to explore alternate fixes. (Perhaps we make some kind of "combined query" to the DoH resolver so it knows to pack the responses together?)

bemasc:
The rationale is there:
> A transport error or timeout can
> occur if an active attacker between the client and the recursive resolver is
> selectively dropping SVCB queries or responses, based on their size or
> other observable patterns.

davidben:
Ah oops! Apparently I can't read. :-)

(We were discussing how we want to handle this and the risks associated.)

Issue #105, opened 2020-01-16
ESNI lifetime implications from Alt-Svc vs HTTPSSVC precedence


davidben:
The draft originally said (prior to #66) said the Alt-Svc cache overrides HTTPSSVC. This had problems (issue #58 and #60), so #66 downgraded it to a MAY. It seems this still has problems.

**HTTPSSVC vs Alt-Svc**

Consider a server which uses both QUIC and ESNI. It configures both in HTTPSSVC. It also cares about HTTPSSVC-less clients (older client or legacy DNS resolver), so it configures QUIC in Alt-Svc. Is it required to configure ESNI in Alt-Svc, or can it leave things alone (with the understanding that ESNI will be limited to clients that support HTTPSSVC)?

The spec currently says a client (that would otherwise support HTTPSSVC) MAY skip an HTTPSSVC lookup given an Alt-Svc cache entry. That means, for ESNI to work, the server MUST configure it in both. This is not obvious and should be written down. More importantly, it has deployment consequences.

HTTPSSVC records apply to the *current* HTTP request. If the client has no cached DNS record, it still queries DNS and gets HTTPSSVC. That means HTTPSSVC TTLs may be set more-or-less freely depending on the site's performance vs. flexibility needs. Let's say it's O(1 hour).

Alt-Svc headers apply to *subsequent* HTTP requests. If the client has no Alt-Svc entry cached, it will send the HTTP request without Alt-Svc. That means Alt-Svc TTLs must cover the time to the next HTTP request for Alt-Svc to be used at all. For reference, I see google.com currently uses 30 days.

**Commitments**

For the duration of the HTTPSSVC or Alt-Svc lifetime, the server operator has made a commitment to the client. ESNI is a soft commitment that the server understands this ESNI key and a hard commitment that the server is colocated with the public name. The first lower-bounds key lifetime and rotation on the server. There is a recovery mechanism, but it is expensive, so this is a soft commitment. The second is roughly a commitment to use a particular hosting provider. ESNI's retry mechanism requires the public name, so this is a hard commitment. Breaking this will knock out your site.

(Note Alt-Svc without ESNI was *not* a hosting provider commitment. A provider-specific Alt-Svc may fail if the site changes providers, but the client could still connect without Alt-Svc. ESNI must take this fallback away to prevent network downgrade.)

[...]

PR #106, opened 2020-01-21
Fully specify HTTPSSVC/Alt-Svc/ESNI interaction

bemasc:
Addresses #105

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -948,6 +948,9 @@ client connection procedure ({{client-behavior}}).
 For clients that implement support for ESNI, the interaction between
 HTTPSSVC and Alt-Svc is described in {{esni-client-behavior}}.
 
+This specification does not alter the DNS queries performed when connecting
+to an Alt-Svc hostname (typically A and/or AAAA only).
+
 ## HTTP Strict Transport Security {#hsts}
 
 By publishing an HTTPSSVC record, the server
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -940,14 +940,13 @@ require single-client granularity.
 
 ## Interaction with Alt-Svc
 
-If the client has an Alt-Svc cache, and a usable Alt-Svc value is
-present in that cache, then the client MAY skip the HTTPSSVC query.
-If Alt-Svc connection fails, the client SHOULD fall back to the HTTPSSVC
+Clients that do not implement support for ESNI MAY skip the HTTPSSVC query
+if a usable Alt-Svc value is available in the local cache.
+If Alt-Svc connection fails, these clients SHOULD fall back to the HTTPSSVC
 client connection procedure ({{client-behavior}}).
 
-Clients that implement ESNI are subject to additional requirements
-(see {{esni-client-behavior}}).
-
+For clients that implement support for ESNI, the interaction between
+HTTPSSVC and Alt-Svc is described in {{esni-client-behavior}}.
 
 ## HTTP Strict Transport Security {#hsts}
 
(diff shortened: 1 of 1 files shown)


bemasc:
This proposal has been substantially simplified now that Alt-Svc changes are excluded from this draft.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -938,8 +938,15 @@ record, groups of clients will necessarily receive the same
 SvcFieldValue.  Therefore, HTTPSSVC is not suitable for uses that
 require single-client granularity.
 
+## Interaction with Alt-Svc
+
 If the client has an Alt-Svc cache, and a usable Alt-Svc value is
 present in that cache, then the client MAY skip the HTTPSSVC query.
+If Alt-Svc connection fails, the client SHOULD fall back to the HTTPSSVC
+client connection procedure ({{client-behavior}}).
```
davidben: This text still allows clients to short-circuit HTTPSSVC if an Alt-Svc cache entry exists, which means all the issues from #105 apply. (Is this just a mistake? The client behavior below seems to suggest HTTPSSVC goes first.)

Or is the intent that the client behavior overrides this sentence? It's probably worth writing this a bit more explicitly. The combination is also weird because that text isn't additional requirements on top of the base algorithm. It completely changes it. (The base algorithm, while not explicitly spelled out, implies that the Alt-Svc integration point is before HTTPSSVC while the ESNI algorithm says it's after.)

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -993,41 +1000,41 @@ of {{HSTS}}.
...
 
-Other clients and services implementing SVCB or HTTPSSVC with esniconfig
-are encouraged to take a similar approach.
+## Deployment considerations
 
+An HTTPSSVC RRSet containing some RRs with esniconfig and some without is
+vulnerable to a downgrade attack.  This configuration is NOT RECOMMENDED.
+Zone owners who do use such a mixed configuration SHOULD mark the RRs with
+esniconfig as more preferred (i.e. smaller SvcFieldPriority) than those
+without.
```
davidben: Marking it more preferred means ESNI is still not downgrade-protected, right? I.e. the prioirty thing is just to be opportunistically okay? It's probably worth clarifying. This confused me at first.

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: > Or is the intent that the ESNI client behavior overrides this paragraph? It's probably worth writing this a bit more explicitly.

Yes.  I've pushed a change to clarify this.

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: OK, I've added a clarification on this point.

draft-ietf-dnsop-svcb-httpssvc.md
enygren: It seems like we should be explicit around whether DNS lookups for the Alt-Svc host are done only for A/AAAA records or also for SVCB/HTTPSSVC records?
Approaches:
1) Be explicit that the host in Alt-Svc is just resolved for A/AAAA records/
2) Have an Alt-Svc attribute of "svcb=1" which would indicate that the Alt-Svc name is to a HTTPSSVC record?  In the behavior ordering, we could include then allow these to go before HTTPSSVC DNS lookups.  This might make sense in a separate AltSvc-bis/rfc7838bis draft that updates this and 7838 (along with other changes).

bemasc: I think your option 1 is our only choice at this point.  I've uploaded a sentence to clarify that.

Some kind of interaction between Alt-Svc and SVCB could be very interesting in the future, but as it stands, I don't see how to make that delegation work.  For example, there could be very complex interactions between Alt-Svc and HTTPSSVC values that can disagree (e.g. ALPN/"transport", "persist", "esniconfig").

I think Alt-Svc-bis is beginning to sound like a good idea, but first we should get this draft finalized.

PR #108, opened 2020-01-24
Remove normative references to Alt-Svc

bemasc:
This simplifies the draft, and allows any changes
to Alt-Svc to happen in a separate draft.  Substantive
text relating HTTPSSVC to Alt-Svc has been moved into
a new file of notes for posterity.

### DIFF
--- a/alt-svc-notes.md
+++ b/alt-svc-notes.md
@@ -1,0 +1,152 @@
+# Alt-Svc ideas related to HTTPSSVC and ESNI
+
+This file contains notes related to Alt-Svc that were removed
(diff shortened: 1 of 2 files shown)


MikeBishop:
It's a good start.  I still think we should define precedence when clients implement both Alt-Svc and HTTPSSVC, though we can potentially handle that separately.

bemasc:
> I still think we should define precedence when clients implement both Alt-Svc and HTTPSSVC, though we can potentially handle that separately.

Agreed.  I'm hoping to handle that in a followup PR.

Issue #95, opened 2019-12-17
Why allow multiple values for the same parameter at all?


davidben:
It seems the only parameters which allow this is ipv4hint and ipv6hint, which is not necessary because the value serialization supports multiple addresses (https://github.com/MikeBishop/dns-alt-svc/pull/94).

Why do we need this? I think we should only allow one value per parameter.

This means duplicate values can be rejected generically in the parser. Right now the spec says individual parameters silently ignore all but the first value. Silently ignoring things isn't great. Other protocols like Roughtime further require keys be sorted, which makes it easy for the parser to reject duplicates generically.

This also avoids questions when designing a new parameter as to whether to use multiple parameter values or to encode multiple structures in the value. We should have one way to do things, so let's use the latter.

enygren:
The other argument would be to always do the former (multiple parameter values) in the wireformat and to strongly discourage things from having values that encode multiple structures in the value.  Presentation format multi-value parameters could be syntactic sugar and convert to to multiple parameters on the wire

enygren:
Agreed that we should be consistent here.

Presumably ESNIConfig is another use-case?

davidben:
I believe HTTPSSVC's ESNI story, as currently specified, does not allow for multiple ESNIConfigs. It doesn't say anything and thus gets the default. This is indeed a mistake, but we're fixing it by doing the multiple values at the TLS level, which lets us handle it consistently at all sources of ESNI config. (See #88.)

It's also generally more compact to handle multiplicity within the value. Each HTTPSSVC value costs four bytes of framing per value. If the value is self-delimiting (true for addresses and ESNIConfig), you have a one-time four byte cost from HTTPSSVC, then each additional value costs no extra framing. If the value is not self-delimiting, you need length prefixes internally, costing two bytes per value. That's two bytes extra at one value, breaking even at two values, and a win beyond that.

davidben:
@dmcardle FYI

enygren:
Your argument seems reasonable.  We should make this clear and deterministic especially at the wireformat layer (eg, must only use the first instance of a parameter, provisioning systems should warn on duplicates, etc).

Allowing syntactic sugar at the presentation layer (eg, concatenation of parameters that are duplicated?) might be helpful usability-wise or could create many more problems and instead we should recommend that provisioning systems prevent this (or at least warn).

bemasc:
[...]

PR #109, opened 2020-02-18
Only allow each SvcParamKey to appear once

bemasc:
This includes some adjustments to make repeated keys
unnecessary.

Incorporates #94 by @davidben
Addresses #95

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -418,9 +418,11 @@ SvcParamKeys SHALL appear in increasing numeric order.
 
 Clients MUST consider an RR malformed if
 * the parser reaches the end of the RDATA while parsing a SvcFieldValue.
-* SvcParamKeys are not in increasing numeric order.
-* a single SvcParamKey appears twice.
+* SvcParamKeys are not in strictly increasing numeric order.
 * a SvcParamValue for a known SvcParamKey does not have the expected format.
+
+Note that the second condition implies that there are no duplicate
+SvcParamKeys.
 
 If any RRs are malformed, the client MUST reject the entire RRSet and
 fall back to non-SVCB connection establishment.
@@ -689,9 +691,6 @@ resolution procedure, as if processing a query for that RRSet.
 This includes following any aliases that the resolver would ordinarily
 follow (e.g. CNAME, {{!DNAME}}).
 
-Recursive resolvers SHOULD NOT attempt to parse the contents of the
-SvcFieldValue.
-
 ## General requirements
 
 All DNS servers SHOULD treat the SvcParam portion of the SVCB RR
@@ -779,14 +778,23 @@ HTTPS, and most are applicable to other protocols as well.
 
 The "transport" and "no-default-transport" SvcParamKeys together
 indicate the set of transport protocols supported by this service endpoint.
-A transport protocol is identified by a protocol-id with the following form:
+Each transport protocol is identified by a protocol-id, which is a sequence
+of 1-255 octets.
 
-    protocol-id = 1*(ALPHA_LC / DIGIT / "-" / "_" / ".")
+    protocol-id = 1*255(OCTET)
 
-The presentation and wire format of "transport" are the same: a comma (0x2c)
-separated list of one or more `protocol-id`s:
+The presentation value of "transport" is a comma-separated list of one or
+more `protocol-id`s.  Any commas present in the protocol-id are escaped
+by a backslash:
 
-   transport-value = protocol-id *("," protocol-id)
+   escaped-octet = %x00-2b / "\," / %x2d-5b / "\\" / %5d-%FF
+   escaped-id = 1*255(escaped-octet)
+   transport-value = escaped-id *("," escaped-id)
+
+In the wire format for "transport", each protocol-id is prefixed by its
+length as a single octet, and these length-value pairs are concatenated
+to form the SvcParamValue.  These pairs MUST exactly fill the SvcParamValue;
+otherwise, the SvcParamValue is malformed.
 
 For "no-default-transport", the presentation and wire format values MUST be
 empty.
@@ -1159,14 +1167,15 @@ that can be used with HTTPS.
 A registration MUST include the following fields:
 
 * Protocol: Typical short name for the transport protocol.
-* Identification Sequence: The precise set of octet values that identifies the protocol, subject to the requirements in {{transport-key}}.
+* Identification Sequence: The precise sequence of 1 to 255 octet values that identifies the protocol.
 * Reference: Pointer to text specifying use of this transport with HTTPS.
 
 Values to be added to this name space require Expert Review (see
 {{!RFC5226}}, Section 4.1).  Identification Sequences MUST NOT collide with
 values in the "TLS Application-Layer Protocol Negotiation (ALPN) Protocol IDs"
-registry, and MUST contain only octets from the ranges 0x61-7a (a-z) and
-0x30-0x39 (0-9).
+registry, and SHOULD contain only id-characters according to this definition:
+
+    id-character = ALPHA_LC / DIGIT / "-" / "." / "_"
 
 ### Initial contents
 
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -416,11 +416,14 @@ SvcParamKey=SvcParamValue pairs, represented as:
 
 SvcParamKeys SHALL appear in increasing numeric order.
 
-Clients MUST reject the RR as malformed if
+Clients MUST consider an RR malformed if
 * the parser reaches the end of the RDATA while parsing a SvcFieldValue.
 * SvcParamKeys are not in increasing numeric order.
 * a single SvcParamKey appears twice.
 * a SvcParamValue for a known SvcParamKey does not have the expected format.
+
+If any RRs are malformed, the client MUST reject the entire RRSet and
+fall back to non-SVCB connection establishment.
 
 TODO: decide if we want special handling for any SvcParamKey ranges?
 For example: range for greasing; experimental range;
@@ -778,7 +781,7 @@ A transport protocol is identified by a protocol-id with the following form:
     protocol-id = 1*(ALPHA_LC / DIGIT / "-" / "_" / ".")
 
 The presentation and wire format of "transport" are the same: a comma (0x2c)
-separated list of one or more`protocol-id`s:
+separated list of one or more `protocol-id`s:
 
    transport-value = protocol-id *("," protocol-id)
 
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -689,6 +689,9 @@ resolution procedure, as if processing a query for that RRSet.
 This includes following any aliases that the resolver would ordinarily
 follow (e.g. CNAME, {{!DNAME}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -415,8 +414,13 @@ each of which contains:
   (but constrained by the RDATA and DNS message sizes).
 * an octet string of the length defined by the previous field.
 
-If the parser reaches the end of the RDATA while parsing a SvcFieldValue,
-the RR is invalid and MUST be discarded.
+SvcParamKeys SHALL appear in increasing numeric order.
+
+Clients MUST reject the RR as malformed if
```
enygren: Should we mention that clients SHOULD instead use other non-malformed RRs in the RRSET if possible?  (We want to avoid forward compat issues where a client implementation bug or some change causes entire rrsets to get rejected.)

Would it be safer for some of these to have deterministic behavior rather than "reject the RR"?

marka63: If a RR is malformed the entire RRset should be rejected.  One cannot safely use partial RRsets.   Operators should not depend on RRsets getting through a recursive server if a RR is malformed.

STD 13 says to reject the entire zone if a single RR is malformed when loading a zone. 
Recursive servers that are aware of the structure of these records should be rejecting answers that contain malformed RRs (they are not permitted to filter out individual RRs).

enygren: That would seem to argue in-favor of having deterministic handling of certain types of errors rather than rejecting the RR? 
For example, in the case of a parameter appearing multiple times, clients MUST use only the first instance.
Overly strict rejection (rather than deterministic handling) could make extensibility much harder in the future?

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -846,17 +853,14 @@ byte order.  Like an A or AAAA RRSet, the list of addresses represents an
 unordered collection, and clients SHOULD pick addresses to use in a random order.
 An empty list of addresses is invalid.
 
-These parameters MAY be repeated multiple times within a record.
-When receiving such a record, clients SHOULD combine the sets of addresses.
-
 When selecting between IPv4 and IPv6 addresses to use, clients may use an
 approach such as {{!HappyEyeballsV2=RFC8305}}.
-When only "ipv4hint" parameters are present, IPv6-only clients may synthesize
+When only "ipv4hint" is present, IPv6-only clients may synthesize
```
tfpauly: Does this restrict us to only one address of each family?

dmcardle: Both ipv4hint and ipv6hint are meant to contain an unordered set of addresses.

Looks like it was not included in the diff context, but this appears earlier in the section:

> The wire format for each parameter is a sequence of IP addresses in network
byte order.  Like an A or AAAA RRSet, the list of addresses represents an
unordered collection, and clients SHOULD pick addresses to use in a random order.
An empty list of addresses is invalid.

draft-ietf-dnsop-svcb-httpssvc.md
marka63: There are exactly 4 conditions listed here.

1 the last TLV being malformed (incomplete tag, length or value)
2 TLVs not being sorted on T
3 multiple T with the same value
4 L being incorrect for a given T

Named already implements 1.
Named already implements 4 for some T.

e.g. for port L == 2, for hint4 L != 0 && L % 4 == 0.

This checks are done as named parses a record learn off the wire or after loading with unknown record format.   All the tools which lookup records also perform these checks. 

None of these currently implemented checks actually prevent new Ts being deployed.

ericorth: I would have been worried about invalid SVCBs breaking everything until I came to accept the suggestion from a couple weeks ago that the client would handle an SVCB timeout by disallowing fallback to A/AAAA.  So we're already into the territory that a resolver doing something bad or a bad config can completely break resolving the name, and not just the SVCB response.

Specific to the case of repeating keys, as long as we have good standardized semantics for list values, I don't see any reason we'd ever need to switch to allowing repeated keys.  So I don't see any reason to ever treat it as anything other than malformed and thus reject the RRSet.  Better for completely invalid things to get noticed by brokenness, so people fix them.

But in general, anywhere that we have good reason to want future extensibility, yes, deterministic handling of things that are "valid but unknown" is often reasonable.  The obvious example is the "clients MUST ignore SvcFieldValue parameters that they do not recognize" rule.

bemasc: > If a RR is malformed the entire RRset should be rejected. One cannot safely use partial RRsets.

I think this is reasonable.  I've updated the PR with that logic.

Technically, I don't think it's a security issue, but this situation should never happen so it makes sense to be cautious.  Who knows what else might have gotten corrupted if one RR is malformed?  Rejecting the whole RRSet seems like a clearer signal to the zone owner that something is wrong with (e.g.) their authoritative server.

> Operators should not depend on RRsets getting through a recursive server if a RR is malformed.
> Recursive servers that are aware of the structure of these records should be rejecting answers that contain malformed RRs (they are not permitted to filter out individual RRs).

I disagree with this.  I've updated the PR to clarify that we don't want recursive resolvers even thinking about the TLVs.  If recursive resolvers are dropping whole RRSets, it will be harder for clients to debug this class of problems (SERVFAIL can mean almost anything).  It's also bad for consistency between resolvers, bad for performance, and could encourage laziness on the part of clients.

> That would seem to argue in-favor of having deterministic handling of certain types of errors rather than rejecting the RR?

I agree that determinism is important, and the text was unclear on what was supposed to happen.  I've updated the PR to clarify the intended client behavior.

> For example, in the case of a parameter appearing multiple times, clients MUST use only the first instance.
> Overly strict rejection (rather than deterministic handling) could make extensibility much harder in the future?

I'm open to this kind of change, but I think if we impose the single-value requirement now, the chances of us wanting to remove it later are very slim.  If we did do something like what you describe, I think we would also need to GREASE it, because clients could easily implement the "first value wins" logic incorrectly (e.g. by rejecting the RR, or using the last value instead of the first) without visibly different behavior in any normal case.

As @marka63 notes, extension by adding new types should continue to work as expected, and seems to cover all of our anticipated needs.

enygren: yeah, it's really the support for new SvcParamTypes that is most important.  (Do we want to encourage greasing of them and allocating a range for greased ones?)

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -765,16 +775,27 @@ responses to the address queries that were issued in parallel.
...
+## "transport" and "no-default-transport" {#transport-key}
 
 The "transport" and "no-default-transport" SvcParamKeys together
 indicate the set of transport protocols supported by this service endpoint.
+A transport protocol is identified by a protocol-id with the following form:
+
+    protocol-id = 1*(ALPHA_LC / DIGIT / "-" / "_" / ".")
+
+The presentation and wire format of "transport" are the same: a comma (0x2c)
+separated list of one or more `protocol-id`s:
```
davidben: Replace this with fixed-width length prefixes. A client parsing DNS (and TLS for that matter) already has tons of routines for such formats. Nothing the client parses in this spec may use separators.

Then revert the change to `protocol-id`. It should be an 8-bit-clean opaque byte string, like ALPN, so we do not have random implementation variations around people remembering to and forgetting to check, getting the check subtle wrong, and weird bugs in the corners.

draft-ietf-dnsop-svcb-httpssvc.md
tfpauly: Great, thanks!

draft-ietf-dnsop-svcb-httpssvc.md
marka63: > > If a RR is malformed the entire RRset should be rejected. One cannot safely use partial RRsets.
> 
> I think this is reasonable. I've updated the PR with that logic.
> 
> Technically, I don't think it's a security issue, but this situation should never happen so it makes sense to be cautious. Who knows what else might have gotten corrupted if one RR is malformed? Rejecting the whole RRSet seems like a clearer signal to the zone owner that something is wrong with (e.g.) their authoritative server.
> 
> > Operators should not depend on RRsets getting through a recursive server if a RR is malformed.
> > Recursive servers that are aware of the structure of these records should be rejecting answers that contain malformed RRs (they are not permitted to filter out individual RRs).
> 
> I disagree with this. I've updated the PR to clarify that we don't want recursive resolvers even thinking about the TLVs. If recursive resolvers are dropping whole RRSets, it will be harder for clients to debug this class of problems (SERVFAIL can mean almost anything). It's also bad for consistency between resolvers, bad for performance, and could encourage laziness on the part of clients.

Yes, SERVFAIL can mean almost anything which is why there are tools like DiG that are designed to follow delegations and talk directly to authoritative servers.  They can even dump malformed responses as hex blobs.

Named has been rejecting malformed RR content for over a decade now and I guarantee you that it has been a net plus.  I think there has been 1 bug report in that entire time where someone put garbage into a record in a zone and expected named to serve it rather than rejecting it or expected it to get through a resolver. 

I also think that you are worrying way too much about malformed records being there in the first place.  People will use tools to generate the records.  The tools will get it right from the start or will be very quickly fixed to get it right.  SERVFAIL from the recursive server is a great motivator to fix tools.

As for lazy clients, whether the resolver rejects malformed records or not it won't make a difference.

I also don't want to have to do the validation of the record as I am printing it out.  Validation on data entry is much safer.

marka63: Also if it isn't the recursive server checking the content it will end up being the firewall and once it is validation the record content debugging becomes just about impossible.

enygren:
> This is over specification on ungrounded fears that people will generate bad RRs that need to be debugged through resolvers. This change is not driven by consensus.

@marka63 : which particular part of this are you concerned about?  (There are lots of items in this thread now.)

draft-ietf-dnsop-svcb-httpssvc.md
enygren: Makes sense to use fixed-width length prefixes to be consistent with everything else.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -415,8 +414,16 @@ each of which contains:
...
 * an octet string of the length defined by the previous field.
 
-If the parser reaches the end of the RDATA while parsing a SvcFieldValue,
-the RR is invalid and MUST be discarded.
+SvcParamKeys SHALL appear in increasing numeric order.
+
+Clients MUST consider an RR malformed if
+* the parser reaches the end of the RDATA while parsing a SvcFieldValue.
+* SvcParamKeys are not in increasing numeric order.
+* a single SvcParamKey appears twice.
```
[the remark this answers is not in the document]

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: Done.  I made them 1-byte prefix, so the max size is 255.

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: @marka63 Thanks for sharing your experience with malformed records in practice.  I've removed the recommendation regarding recursive server behavior.

I think that this draft is negotiating a culture clash between the flexibility of DNSOP and the paranoia of TLS, but I'm optimistic that we'll be able to find a middle ground on the question of specificity.

enygren:
Did we have any remaining issues here before we merge?

bemasc:
(I'm ready to merge after approval from one of the authors.)

PR #110, opened 2020-02-19
Add reference to the GitHub

enygren:
Warren Kumari suggests on dnsop:
=======================

Could I suggest adding something like:
[ RFC Editor - please remove before publication. This document is being
collaborated on in Github at:
https://github.com/MikeBishop/dns-alt-svc.  The most
recent version of the document, open issues, etc should all be
available here.  The authors (gratefully) accept pull requests.]

to the abstract - that way, people reading the draft know where submit PRs, etc.


bemasc:
:+1:

Issue #115, opened 2020-03-04
Consider switching back to ALPN


enygren:
Based on some feedback from @martinthomson we should consider switching back to ALPN rather than introducing "transport".  As part of this we'd want to clarify that:
1) ALPN defines the whole stack  (eg, "H2-over-TLS-over-TCP" )
2) The alpn thus implicitly specifies a transport protocol but also allows filtering of higher-level protocols (ie, to allow doh and dot on different endpoints).
3) Secure in-band negotiation takes precedence, so if the alpn negotiated is different from one indicated in the record then the client MUST use it.

We'd then switch from having a default transport protocol to a default alpn (eg, "http/1.1" for HTTPSSVC) and have no-default-alpn as a param.

MikeBishop:
Done in #116.

PR #116, opened 2020-03-05
Switch from transport= back to alpn=

enygren:
Per discussion in issue #115, switch back to using ALPN but with clarifications.

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -143,11 +143,12 @@ Consider a simple zone of the form
 
 The domain owner could add records like
 
-    simple.example. 7200 IN HTTPSSVC 1 . transport=quic ...
+    simple.example. 7200 IN HTTPSSVC 1 . alpn=h3 ...
 
 The presence of this record indicates to clients that simple.example
 supports HTTPS, and the key=value pairs indicate that it supports QUIC
-in addition to HTTPS over TLS.  The records can also include other information
+in addition to HTTPS over TLS (an implicit default).
+The records can also include other information
 (e.g. non-standard ports, ESNI configuration).
 
 ## Example: Apex aliasing
@@ -183,7 +184,7 @@ it has deployed HTTP/3 on a new server pool with a different
 configuration.  This can be expressed in the following form:
 
     $ORIGIN svc.example. ; A hosting provider.
-    pool  7200 IN HTTPSSVC 1 h3pool transport=quic esniconfig="123..."
+    pool  7200 IN HTTPSSVC 1 h3pool alpn=h3 esniconfig="123..."
                   HTTPSSVC 2 .      esniconfig="abc..."
     pool   300 IN A        192.0.2.2
                   AAAA     2001:db8::2
@@ -460,12 +461,12 @@ As an example:
 
     _8443._foo.api.example.com. 7200 IN SVCB 0 svc4.example.net.
     svc4.example.net.  7200  IN SVCB 3 svc4.example.net. (
-        transport="bar" port="8004" esniconfig="..." )
+        alpn="bar" port="8004" esniconfig="..." )
 
 would indicate that "foo://api.example.com:8443" is aliased
 to the service endpoints offered at "svc4.example.net" on port 8004,
-which support the transport protocol "bar" in addition to the default
-transport for "foo://".
+which support the transport protocol associated with "bar" in 
+addition to the default transport protocol for "foo://".
 
 (Parentheses are used to ignore a line break ({{RFC1035}} Section 5.1).)
 
@@ -778,51 +779,75 @@ responses to the address queries that were issued in parallel.
 A few initial SvcParamKeys are defined here.  These keys are useful for
 HTTPS, and most are applicable to other protocols as well.
 
-## "transport" and "no-default-transport" {#transport-key}
+## "alpn" and "no-default-alpn" {#alpn-key}
 
-The "transport" and "no-default-transport" SvcParamKeys together
-indicate the set of transport protocols supported by this service endpoint.
-Each transport protocol is identified by a protocol-id, which is a sequence
-of 1-255 octets.
+The "alpn" and "no-default-alpn" SvcParamKeys together
+indicate the set of Application Layer Protocol Negotation (ALPN)
+protocol names (per {{ALPN=?RFC7301}}) 
+and associated transport protocols supported by this service endpoint.  
 
-    protocol-id = 1*255(OCTET)
+As with {{AltSvc}}, the ALPN protocol name is used to 
+identify the application protocol and associated suite 
+of protocols used by the SVCB record (the "protocol suite").
+Clients use the set of ALPN protocol
+to filter SVCB records to the set of protocol suites they support,
+and then this informs the underlying transport protocol used (such
+as QUIC-over-UDP or TLS-over-TCP).
 
-The presentation value of "transport" is a comma-separated list of one or
-more `protocol-id`s.  Any commas present in the protocol-id are escaped
+The actual Application Layer Protocol negotiated during 
+the secure handshake MAY differ from the ALPN specified 
+in the SVCB record.
+
+ALPNs are identified by their registered "Identification Sequence"
+(alpn-id), which is a sequence of 1-255 octets.
+
+    alpn-id = 1*255(OCTET)
+
+The presentation value of "alpn" is a comma-separated list of one or
+more `alpn-id`s.  Any commas present in the protocol-id are escaped
 by a backslash:
 
    escaped-octet = %x00-2b / "\," / %x2d-5b / "\\" / %5d-%FF
    escaped-id = 1*255(escaped-octet)
-   transport-value = escaped-id *("," escaped-id)
+   alpn-value = escaped-id *("," escaped-id)
 
-In the wire format for "transport", each protocol-id is prefixed by its
+In the wire format for "alpn", each alpn-id is prefixed by its
 length as a single octet, and these length-value pairs are concatenated
 to form the SvcParamValue.  These pairs MUST exactly fill the SvcParamValue;
 otherwise, the SvcParamValue is malformed.
 
-For "no-default-transport", the presentation and wire format values MUST be
+For "no-default-alpn", the presentation and wire format values MUST be
 empty.
 
 Each scheme that is mapped to SVCB defines a set or registry of allowed
-protocol IDs, and a "default set" of supported IDs, which SHOULD NOT
-be empty.  To determine the set of transport protocols supported by an
-endpoint (the "transport set"), the client parses the "transport" value's
-protocol IDs, and then adds the default set unless the
-"no-default-transport" SvcParamKey is present.  The presence of a value in
-the transport set indicates that this service endpoint, described by
+ALPNs, and a "default set" of supported ALPNs, which SHOULD NOT
+be empty.  To determine the set of protocol suites supported by an
+endpoint (the "alpn set"), the client parses the "alpn" values'
+ALPN IDs, and then adds the default set unless the
+"no-default-alpn" SvcParamKey is present.  The presence of a value in
+the alpn set indicates that this service endpoint, described by
 SvcDomainName and the other parameters (e.g. "port") offers service with
-that transport.
+the protocol suite associated with the ALPN ID.
 
 Clients SHOULD NOT attempt connection to a service endpoint whose
-transport set does not contain any compatible transport protocols.  To ensure
+alpn set does not contain any compatible protocols suites.  To ensure
 consistency of behavior, clients MAY reject the entire SVCB RRSet and fall
 back to basic connection establishment if all of the RRs indicate
-"no-default-transport", even if connection could have succeeded using a
-non-default transport.
+"no-default-alpn", even if connection could have succeeded using a
+non-default alpn.
 
 For compatibility with clients that require default transports,
 zone operators SHOULD ensure that at least one RR in each RRSet supports the
 default transports.
+
+As long as {{ALPN}} is securely negotiated as part of a cryptographic 
+handshake, clients MUST be willing to use "alpn" protocols negotiated
+during a handshake even if they do not match an alpn value in 
+the alpn set.
+
+In cases where the alpn set is a list of multiple values,
+clients SHOULD prefer values earlier in the list, followed 
+by the default alpn if-present.
 
 ## "port"
 
@@ -909,8 +934,9 @@ semantics apply equally to HTTPSSVC unless specified otherwise.
 All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPSSVC.
 
-The allowed transport protocols are the protocols in the HTTPS Transport
-Registry ({{transportregistry}}).  The only default transport value is "tls".
+The allowed protocols suites are the protocols in the ALPN
+Registry defined for HTTPS.  The only default alpn value 
+is "http/1.1".
 
 The presence of an HTTPSSVC record for an origin also indicates
 that all HTTP resources are available over HTTPS, as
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+non-default alpn.
 
 For compatibility with clients that require default transports,
 zone operators SHOULD ensure that at least one RR in each RRSet supports the
 default transports.
 
+As long as {{ALPN}} is securely negotiated as part of a cryptographic 
+handshake, clients MUST be willing to use "alpn" protocols negotiated
+during a handshake even if they do not match an alpn value in 
+the alpn set.
```
martinthomson:
```suggestion
handshake, clients MUST be willing to use protocols negotiated
during a handshake even if they do not match an ALPN value from the
set of values indicated in the "alpn" parameter.
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+
+In cases where the alpn set is a list of multiple values,
```
martinthomson:
```suggestion
In cases where the ALPN set is a list of multiple values,
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+clients SHOULD prefer values earlier in the list, followed 
+by the default alpn if-present.
```
martinthomson:
```suggestion
by the default ALPN identifier if-present.
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -909,8 +934,9 @@ semantics apply equally to HTTPSSVC unless specified otherwise.
 All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPSSVC.
 
-The allowed transport protocols are the protocols in the HTTPS Transport
-Registry ({{transportregistry}}).  The only default transport value is "tls".
+The allowed protocols suites are the protocols in the ALPN
+Registry defined for HTTPS.  The only default alpn value
```
martinthomson:
```suggestion
Registry defined for HTTPS.  The only default ALPN value 
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -960,6 +986,9 @@ HTTPS, and receiving an HTTPSSVC record is intended to be similar to
 receiving that field value over HTTPS.  However, there are some
 differences in the intended client and server behavior.
 
+Unlike Alt-Svc, the ALPN value in a SVCB record is not
+guaranteed to match the ALPN value negotiated in the TLS handshake.
```
martinthomson: I would prefer to correct Alt-Svc on this point.  But that requires more work.  Perhaps you could leave a marker here indicating a desire to fix that aspect of 7838.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
-protocol IDs, and a "default set" of supported IDs, which SHOULD NOT
-be empty.  To determine the set of transport protocols supported by an
-endpoint (the "transport set"), the client parses the "transport" value's
-protocol IDs, and then adds the default set unless the
-"no-default-transport" SvcParamKey is present.  The presence of a value in
-the transport set indicates that this service endpoint, described by
+ALPNs, and a "default set" of supported ALPNs, which SHOULD NOT
+be empty.  To determine the set of protocol suites supported by an
+endpoint (the "alpn set"), the client parses the "alpn" values'
+ALPN IDs, and then adds the default set unless the
```
martinthomson:
```suggestion
endpoint (the "ALPN set"), the client parses the set of ALPN identifiers in
the "alpn" parameter, and then adds the default set unless the
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
 
-In the wire format for "transport", each protocol-id is prefixed by its
+In the wire format for "alpn", each alpn-id is prefixed by its
 length as a single octet, and these length-value pairs are concatenated
 to form the SvcParamValue.  These pairs MUST exactly fill the SvcParamValue;
 otherwise, the SvcParamValue is malformed.
 
-For "no-default-transport", the presentation and wire format values MUST be
+For "no-default-alpn", the presentation and wire format values MUST be
 empty.
```
martinthomson: Separately, I think that `no-default-alpn` is the wrong default.  If a server wants to support HTTP/1.1 in this way, the bits it spends on saying that (I count 9 bytes) isn't that much.  I suspect that we'll want to prioritize space for servers that support other protocols in future, and burdening them with an additional attribute is not great.

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+more `alpn-id`s.  Any commas present in the protocol-id are escaped
 by a backslash:
 
    escaped-octet = %x00-2b / "\," / %x2d-5b / "\\" / %5d-%FF
    escaped-id = 1*255(escaped-octet)
-   transport-value = escaped-id *("," escaped-id)
+   alpn-value = escaped-id *("," escaped-id)
 
-In the wire format for "transport", each protocol-id is prefixed by its
+In the wire format for "alpn", each alpn-id is prefixed by its
```
martinthomson:
```suggestion
In the wire format for "alpn", each alpn-value is prefixed by its
```

```suggestion
In the wire format for "alpn", each ALPN identifier (alpn-value) is prefixed by its
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+identify the application protocol and associated suite 
+of protocols used by the SVCB record (the "protocol suite").
+Clients use the set of ALPN protocol
+to filter SVCB records to the set of protocol suites they support,
+and then this informs the underlying transport protocol used (such
+as QUIC-over-UDP or TLS-over-TCP).
 
-The presentation value of "transport" is a comma-separated list of one or
-more `protocol-id`s.  Any commas present in the protocol-id are escaped
+The actual Application Layer Protocol negotiated during
```
martinthomson:
```suggestion
The actual application-layer protocol negotiated during 
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+The "alpn" and "no-default-alpn" SvcParamKeys together
+indicate the set of Application Layer Protocol Negotation (ALPN)
+protocol names (per {{ALPN=?RFC7301}}) 
+and associated transport protocols supported by this service endpoint.  
 
-    protocol-id = 1*255(OCTET)
+As with {{AltSvc}}, the ALPN protocol name is used to 
+identify the application protocol and associated suite 
+of protocols used by the SVCB record (the "protocol suite").
+Clients use the set of ALPN protocol
```
martinthomson:
```suggestion
Clients use the set of ALPN identifiers
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
-## "transport" and "no-default-transport" {#transport-key}
+## "alpn" and "no-default-alpn" {#alpn-key}
 
-The "transport" and "no-default-transport" SvcParamKeys together
-indicate the set of transport protocols supported by this service endpoint.
-Each transport protocol is identified by a protocol-id, which is a sequence
-of 1-255 octets.
+The "alpn" and "no-default-alpn" SvcParamKeys together
+indicate the set of Application Layer Protocol Negotation (ALPN)
+protocol names (per {{ALPN=?RFC7301}})
```
martinthomson:
```suggestion
protocol identifiers {{?ALPN=RFC7301}} 
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
+ALPN IDs, and then adds the default set unless the
+"no-default-alpn" SvcParamKey is present.  The presence of a value in
+the alpn set indicates that this service endpoint, described by
 SvcDomainName and the other parameters (e.g. "port") offers service with
-that transport.
+the protocol suite associated with the ALPN ID.
 
 Clients SHOULD NOT attempt connection to a service endpoint whose
-transport set does not contain any compatible transport protocols.  To ensure
+alpn set does not contain any compatible protocols suites.  To ensure
```
bemasc:
```suggestion
alpn set does not contain any compatible protocol suites.  To ensure
```

draft-ietf-dnsop-svcb-httpssvc.md
bemasc:
```suggestion
In cases where the alpn set contains multiple values,
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
 zone operators SHOULD ensure that at least one RR in each RRSet supports the
 default transports.
 
+As long as {{ALPN}} is securely negotiated as part of a cryptographic 
+handshake, clients MUST be willing to use "alpn" protocols negotiated
+during a handshake even if they do not match an alpn value in 
+the alpn set.
+
+In cases where the alpn set is a list of multiple values,
+clients SHOULD prefer values earlier in the list, followed
```
bemasc: If the order matters then this isn't a set, and we need to define, where the default ALPN is inserted, for example.  However, I don't understand what value we are gaining for this complexity.  Why would a server need to tell the client which supported transport should be preferred?

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +779,76 @@ responses to the address queries that were issued in parallel.
...
-non-default transport.
+"no-default-alpn", even if connection could have succeeded using a
+non-default alpn.
 
 For compatibility with clients that require default transports,
 zone operators SHOULD ensure that at least one RR in each RRSet supports the
 default transports.
 
+As long as {{ALPN}} is securely negotiated as part of a cryptographic 
+handshake, clients MUST be willing to use "alpn" protocols negotiated
```
bemasc: I think this paragraph should be replaced by the text we previously had on this point: https://github.com/MikeBishop/dns-alt-svc/commit/9c133e81c8932d4017a9cff040577c09273186c7#diff-6a78b06198e2a4dd4bb7dc07a558a5adR696.  That text essentially states this requirement in the opposite way, which I think is clearer.  It could also be combined with the paragraph at Line 797 to reduce repetition.

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: This structure isn't intended to increase efficiency.  The additional attribute is there to make it hard to disable TLS accidentally.  Otherwise, a zone owner could easily write "alpn=h3", observe that everything works in their testing, and deploy without realizing that they've created an outage for all their non-QUIC-capable clients.

bemasc:
I think the text needs to address the question of ALPNs indicating a transport, e.g.
```suggestion
This specification requires that each ALPN value unambiguously indicates a single protocol suite.  If an ALPN value is specified for use with multiple protocol suites (e.g. both TLS and DTLS), it MUST NOT be included in the SvcParamValue.
```
Alternatively, this restriction could be documented as a requirement specific to HTTPSSVC.

Ideally, I would define a new "ALPN Context" registry with values "TLS", "DTLS", and "QUIC", and add a column to the ALPN registry indicating which context each ALPN is allowed in.  However, I worry that will create delay if we add it to this draft, and it runs into a problem with the "stun" ALPNs (which are allowed in both TLS and DTLS).

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: The text currently allows a scheme to have _multiple_ default ALPNs, so this needs to be a plural, and instead of a "default set" we would presumably need a "default list" in a specified order.  (As above, I don't think this is worthwhile.)

chris-wood: Agreed. I'd drop this sentence.

enygren: Sounds good to me to drop this.

draft-ietf-dnsop-svcb-httpssvc.md
martinthomson: I think that it's fine to say that this is an expression of the server's preferences.

Ordering relative to the default is why I would prefer to see no default value.

draft-ietf-dnsop-svcb-httpssvc.md
martinthomson: One caution there: "For example, when using a SvcFieldValue with an "alpn" of "h2", the client MAY also" has a "MAY" in a "For example".  The word "could" is better.

draft-ietf-dnsop-svcb-httpssvc.md
bemasc: OK, I added a marker.

draft-ietf-dnsop-svcb-httpssvc.md

draft-ietf-dnsop-svcb-httpssvc.md

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +780,78 @@ responses to the address queries that were issued in parallel.
...
-that transport.
+the protocol suite associated with the ALPN ID.
+
+ALPN IDs that do not uniquely identify a protocol suite (e.g. an ID that
+can be used with both TLS and DTLS) are not compatible with this
+SvcParamKey and MUST NOT be included in the ALPN set.
 
 Clients SHOULD NOT attempt connection to a service endpoint whose
-transport set does not contain any compatible transport protocols.  To ensure
+alpn set does not contain any compatible protocol suites.  To ensure
```
bemasc:
```suggestion
ALPN set does not contain any compatible protocol suites.  To ensure
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -778,52 +780,78 @@ responses to the address queries that were issued in parallel.
...
 empty.
 
-Each scheme that is mapped to SVCB defines a set or registry of allowed
-protocol IDs, and a "default set" of supported IDs, which SHOULD NOT
-be empty.  To determine the set of transport protocols supported by an
-endpoint (the "transport set"), the client parses the "transport" value's
-protocol IDs, and then adds the default set unless the
-"no-default-transport" SvcParamKey is present.  The presence of a value in
-the transport set indicates that this service endpoint, described by
+Each scheme that is mapped to SVCB defines a
```
bemasc:
```suggestion
Each scheme that uses this SvcParamKey defines a
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1146,53 +1182,20 @@ MUST NOT start with "key".
...
-| 3           | port                 | Port for alternative service     | (This document) |
-| 4           | ipv4hint             | IPv4 address hints               | (This document) |
-| 5           | esniconfig           | Encrypted SNI configuration      | (This document) |
-| 6           | ipv6hint             | IPv6 address hints               | (This document) |
-| 65280-65534 | keyNNNNN             | Private Use                      | (This document) |
-| 65535       | key65535             | Reserved                         | (This document) |
+| SvcParamKey | NAME            | Meaning                         | Reference       |
+| ----------- | ------          | ----------------------          | --------------- |
+| 0           | (no name)       | Reserved for internal use       | (This document) |
+| 1           | alpn            | Additional supported protocol   | (This document) |
```
bemasc:
```suggestion
| 1           | alpn            | Additional supported protocols  | (This document) |
```

PR #127, opened 2020-03-11
Change "ESNI" to "ECHO" to match tls-esni-06

bemasc:


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -462,7 +462,7 @@ As an example:
 
     _8443._foo.api.example.com. 7200 IN SVCB 0 svc4.example.net.
     svc4.example.net.  7200  IN SVCB 3 svc4.example.net. (
-        alpn="bar" port="8004" esniconfig="..." )
+        alpn="bar" port="8004" echo="..." )
```
chris-wood: ECHO uses `ECHOConfig` to carry this data -- should we do the same here?

davidben: Given "echo" has other meetings, that's probably a good plan. "echoconfig" is a hair longer but a more unique string.

bemasc:
> Did we intend to remove "## Differences from the proposed ESNI record "?

I removed it because the ESNI record proposal no longer appears in the ESNI (now ECHO) draft, so there's no longer anything to compare against.  Ultimately, I hope to get rid of all of these comparisons before RFC.

> Would we be better off with calling this "tlsecho" instead of "echoconfig" ?
(I think either is fine.)

I prefer "echoconfig", since (a) it's closer to the ECHO draft terminology, and (b) it avoids possible confusion around TLS vs QUIC.

PR #129, opened 2020-04-07
Improve non-HTTP example text

### LINKED ISSUE #124: Clarify non-HTTPS example
This example currently fuses two different zone files.  We can make it clearer that they are separate by adding a comment line above each RR.

> """
>  As an example:
> 
>    _8443._foo.api.example.com. 7200 IN SVCB 0 svc4.example.net.
>    svc4..example.net.  7200  IN SVCB 3 ( svc4.example.net. alpn="bar"
>                                       port="8004" esniconfig="..." )
> """
> 
> While many examples are useful for the understanding, they often provide the
> impression RRsets are in the same zone file. I am wondering if that would not
> be clarifying to show and mention zone files are different. Note this is just a
> feed back and your ar efree to ignore it.

bemasc:
Fixes #124

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -141,12 +141,11 @@ Consider a simple zone of the form
     simple.example. 300 IN A    192.0.2.1
                            AAAA 2001:db8::1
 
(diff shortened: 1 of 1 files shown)


bemasc:
Done

PR #139, opened 2020-04-21
Fix up ABNF to comply with RFC5234

moonshiner:
in 5234, the rulename says it must contain a "combination of alphabetics, digits, and hyphens (dashes)"

updated ALPHA_LC

also, %5d-%FF was slightly incorrect.


bemasc:
:+1:

PR #140, opened 2020-04-23
Clarify serialization of values of unknown SvcParamKeys

DesWurstes:
I've added the following:

```
When decoding values of unrecognized keys in the presentation format:

* a character other than "\" represents its ASCII value in wire format.
* the character "\" followed by three decimal digits, up to 255, represents
  an octet in the wire format.
* the character "\" followed by any allowed character, except a decimal digit,
  represents the subsequent character's ASCII value.

```
and some other changes as discussed with bemasc in https://github.com/MikeBishop/dns-alt-svc/pull/140/commits/c69812c542fcdce466717f2f55e6ec0a5966e740

Possible todo:
TODO remove outside ascii characters from alpn, inherit escaping from values of private keys

---
Resolved questions:

And escaped alpn values in the presentation format can contain characters outside the ASCII range, non-visible characters or even the null character. Should I change that too?

Moreover, some alpn values are in the gray area: They fit in 255 octets in wire format, but they are longer than 255 characters in the presentation format. (For example `,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,`) We should ban those alpn values too?

Similarly, *The presentation value of "alpn" is a comma-separated list of one or more "alpn-id"s.*, It should say at least one alpn value for the wire format as well?

And since there's no requirement about the order of the pairs in the presentation format, duplicate pairs are allowed?

---
and I would like to add [my implementation](https://github.com/miekg/dns/pull/1067) and [this implementation](https://github.com/rthalley/dnspython/pull/452) to the list. That should be a separate pull request?

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -393,6 +393,14 @@ In presentation format, values of unrecognized keys
 SHALL be represented in wire format, using decimal escape codes
 (e.g. \255) when necessary.
 
+When decoding values of unrecognized keys in the presentation format:
+
+* a character other than "\" represents its ASCII value in wire format.
+* the character "\" followed by three decimal digits, up to 255, represents
+  an octet in the wire format.
+* the character "\" followed by any allowed character, except a decimal digit,
+  represents the subsequent character's ASCII value.
+
 Elements in presentation format MAY appear in any order.
 
 ## SVCB RDATA Wire Format
@@ -425,7 +433,7 @@ Clients MUST consider an RR malformed if
 
 * the parser reaches the end of the RDATA while parsing a SvcFieldValue.
 * SvcParamKeys are not in strictly increasing numeric order.
-* a SvcParamValue for a known SvcParamKey does not have the expected format.
+* the SvcParamValue for an SvcParamKey does not have the expected format.
 
 Note that the second condition implies that there are no duplicate
 SvcParamKeys.
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -401,7 +401,8 @@ When decoding values of unrecognized keys in the presentation format:
 * the character "\" followed by any allowed character, except a decimal digit,
   represents the subsequent character's ASCII value.
 
-Elements in presentation format MAY appear in any order.
+Elements in presentation format MAY appear in any order, but keys MUST NOT be
+repeated.
 
 ## SVCB RDATA Wire Format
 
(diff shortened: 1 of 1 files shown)


bemasc:
> And escaped alpn values in the presentation format can contain characters outside the ASCII range, non-visible characters or even the null character. Should I change that too?

I guess so.  It's unlikely ever to be an issue (putting non-ASCII values in the ALPN registry would be very weird).

> Moreover, some alpn values are in the gray area: They fit in 255 octets in wire format, but they are longer than 255 characters in the presentation format. (For example ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,) We should ban those alpn values too?

No, we shouldn't have a length limit in the presentation format.

> Similarly, The presentation value of "alpn" is a comma-separated list of one or more "alpn-id"s., It should say at least one alpn value for the wire format as well?

If you would prefer to make that a wire format syntax error, I'm fine with that.  I don't think it does any harm, but I suppose it's one less weird variation that could go unnoticed.

> And since there's no requirement about the order of the pairs in the presentation format, duplicate pairs are allowed?

No, duplicate pairs are not allowed in the presentation format.  If the text is unclear on that point, please suggest a clarification.

bemasc:
:+1:

DesWurstes:
@bemasc Thanks, but I've just found an undocumented feature/bug in markdown. This [webpage which I found from Github help pages says](https://daringfireball.net/projects/markdown/syntax#backslash), that the \ escapes certain characters, which don't include double quotes. However the backslash in the output isn't rendered, it only escapes the next ". So **could you please replace the \ in [those lines](https://github.com/MikeBishop/dns-alt-svc/blame/master/draft-ietf-dnsop-svcb-httpssvc.md#L396-L403) with \\**? Or is it easier for you if I open a pull request?

bemasc:
It would be easier if you could get it rendering the way you want and send a PR.  Thanks!

PR #141, opened 2020-04-30
Clarifications to the key-value definition text

### LINKED ISSUE #136: Clarify presentation format section
The wording about key=value syntax is confusing.  It could be clearer, e.g.
>    The presentation format for SvcFieldValue is a whitespace-separated list of
>    key=value pairs (e.g. "key123=value1 keys456=value2"). When the value, or
>    both the value and the "=" are omitted, the value should be interpreted as
>    being empty.

bemasc:
Fixes #136

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -367,11 +367,11 @@ Registered key names should only contain characters from the ranges
 Values are in a format specific to the SvcParamKey.
 Their definition should specify both their presentation format
 and wire encoding (e.g., domain names, binary data, or numeric values).
+The initial keys and formats are defined in {{keys}}.
 
 The presentation format for SvcFieldValue is a whitespace-separated
-list of elements representing a key-value pair, with an absent value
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -367,11 +367,11 @@ Registered key names should only contain characters from the ranges
...
 Their definition should specify both their presentation format
 and wire encoding (e.g., domain names, binary data, or numeric values).
+The initial keys and formats are defined in {{keys}}.
 
 The presentation format for SvcFieldValue is a whitespace-separated
-list of elements representing a key-value pair, with an absent value
-or "=" indicating an empty value.  Each element is presented in the
-following form:
+list of key=value pairs.  When the value, or both the value and the "=", are
+omitted, the presentation value is the empty string.
```
enygren, line 374: The number of commas here is confusing.  Is there a way to reword this?
"When the value is omitted, or both the value and the "=" are omittied", then the presentation value is the empty string."

PR #143, opened 2020-04-30
Add warning text about non-default ports

### LINKED ISSUE #134: Add warning about using non-default ports
Paul Vixie recommends we add a warning that operators should avoid using non-default ports for general Internet use.  Non-default ports may be firewalled in client networks, so may appear to work in testing but may not work for some clients/users.

bemasc:
Fixes #134


bemasc:
OK, I've adjusted the wording similar to your suggestion

PR #145, opened 2020-05-02
Add some text clarifying echoconfig parameter usage.

chris-wood:
cc @bemasc

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -1121,6 +1121,14 @@ Zone owners who do use such a mixed configuration SHOULD mark the RRs with
 without, in order to maximize the likelihood that ECHO will be used in the
 absence of an active adversary.
 
+## Relation to ECHOConfig Extensions
+
+The TLS ECHOConfig structure {{!ECHO}} supports adding arbitrary parameters via
+extensions. This extension mechanism SHOULD NOT be used for any parameter that
+might influence a client's connection properties. In particular, any SVCB parameter
+other than "echoconfig" which affects the TLS ClientHello are directed at the
+encrypted inner ClientHello when ECHO is in use.
+
 # Interaction with other standards
 
 This standard is intended to reduce connection latency and
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -1126,7 +1126,7 @@ absence of an active adversary.
 The TLS ECHOConfig structure {{!ECHO}} supports adding arbitrary parameters via
 extensions. That extension mechanism SHOULD NOT be used for any parameter that
 might influence a client's connection properties. In particular, any SVCB parameter
-other than "echoconfig" which affects the TLS ClientHello are directed at the
+other than "echoconfig" which affects the TLS ClientHello is directed at the
 encrypted inner ClientHello when ECHO is in use.
 
 # Interaction with other standards
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -1124,7 +1124,7 @@ absence of an active adversary.
 ## Relation to ECHOConfig Extensions
 
 The TLS ECHOConfig structure {{!ECHO}} supports adding arbitrary parameters via
-extensions. This extension mechanism SHOULD NOT be used for any parameter that
+extensions. That extension mechanism SHOULD NOT be used for any parameter that
 might influence a client's connection properties. In particular, any SVCB parameter
 other than "echoconfig" which affects the TLS ClientHello are directed at the
 encrypted inner ClientHello when ECHO is in use.
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -184,8 +184,8 @@ it has deployed HTTP/3 on a new server pool with a different
 configuration.  This can be expressed in the following form:
 
     $ORIGIN svc.example. ; A hosting provider.
-    pool  7200 IN HTTPSSVC 1 h3pool alpn=h2,h3 echo="123..."
-                  HTTPSSVC 2 .      alpn=h2 echo="abc..."
+    pool  7200 IN HTTPSSVC 1 h3pool alpn=h2,h3 echoconfig="123..."
+                  HTTPSSVC 2 .      alpn=h2 echoconfig="abc..."
     pool   300 IN A        192.0.2.2
                   AAAA     2001:db8::2
     h3pool 300 IN A        192.0.2.3
@@ -292,7 +292,7 @@ intended destination to all entities along the network path.
 ## Parameter for Encrypted ClientHello
 
 This document also defines a parameter for Encrypted ClientHello {{!ECHO}}
-keys. See {{echo}}.
+keys. See {{echoconfig}}.
 
 ## Terminology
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1121,6 +1121,14 @@ Zone owners who do use such a mixed configuration SHOULD mark the RRs with
 without, in order to maximize the likelihood that ECHO will be used in the
 absence of an active adversary.
 
+## Relation to ECHOConfig Extensions
+
+The TLS ECHOConfig structure {{!ECHO}} supports adding arbitrary parameters via
+extensions. This extension mechanism SHOULD NOT be used for any parameter that
```
bemasc, line 1127:
```suggestion
extensions. That extension mechanism SHOULD NOT be used for any parameter that
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1121,6 +1121,14 @@ Zone owners who do use such a mixed configuration SHOULD mark the RRs with
...
+might influence a client's connection properties. In particular, any SVCB parameter
+other than "echoconfig" which affects the TLS ClientHello are directed at the
```
bemasc, line 1129:
```suggestion
other than "echoconfig" which affects the TLS ClientHello is directed at the
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1121,6 +1121,14 @@ Zone owners who do use such a mixed configuration SHOULD mark the RRs with
 without, in order to maximize the likelihood that ECHO will be used in the
 absence of an active adversary.
 
+## Relation to ECHOConfig Extensions
+
+The TLS ECHOConfig structure {{!ECHO}} supports adding arbitrary parameters via
+extensions. This extension mechanism SHOULD NOT be used for any parameter that
+might influence a client's connection properties. In particular, any SVCB parameter
```
bemasc, line 1128: I think "connection properties" is not clear enough.  To me, the only way an ECHOConfig extension could not affect "connection properties" would be if it were ignored.

draft-ietf-dnsop-svcb-httpssvc.md
bemasc, line 1129:
```suggestion
other than "echoconfig" that affects the TLS ClientHello is directed at the
```

draft-ietf-dnsop-svcb-httpssvc.md
chris-wood: That may be true. How might you rephrase this?

chris-wood: I took a stab at refactoring this. What do you think?

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1091,6 +1091,11 @@ In wire format, the value of the parameter is an ECHOConfigs vector
 {{!ECHO}}, including the redundant length prefix.  In presentation format,
 the value is encoded in {{!base64=RFC4648}}.
 
+The value of this parameter and any corresponding ECHOConfig extensions are directed
```
bemasc, line 1094:
```suggestion
The value of this parameter and any included ECHOConfig extensions are directed
```

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1091,6 +1091,11 @@ In wire format, the value of the parameter is an ECHOConfigs vector
...
+at the outer TLS ClientHello when ECHO is in use. This parameter does not affect
+parameters of the inner TLS ClientHello. Thus, changes targeting the inner TLS
+ClientHello when ECHO is in use SHOULD use other SVCB parameters.
```
bemasc, line 1097:
```suggestion
ClientHello when ECHO is in use SHOULD use other SVCB parameters, and any other
relevant SVCB parameters SHALL be applied to the inner ClientHello.
```

chris-wood: I'd prefer to keep these separate sentences.

bemasc: That's fine with me, but I do think we ought to mention the second requirement somewhere in this draft.   (Currently it's only in the ECHO draft.)

chris-wood: 👍 Do you want to merge and then make that change? (This PR targets your ESNI->ECHO branch.)

bemasc: I think it's time for another round of review with the other coauthors.

chris-wood: Calling @MikeBishop and @enygren!

MikeBishop: I feel like this discussion is about which way to overspecify.

The outer ClientHello is an artifact of ECHO: essentially fake and deliberately public.  Therefore, it's entirely legitimate to populate it with public information, e.g. information anyone can obtain over DNS anyway.  It's also legitimate to populate it with junk or lie.  Those are implementation choices; we don't need to specify them.

The contents of the SVCB record are a server's declarations about its own capabilities; they are useful in the inner ClientHello and might be usable in the outer ClientHello if the implementation so chooses, since they're public anyway.  This particular parameter is about how to conceal the "true" ClientHello using ECHO.  The point of this paragraph is that you don't need this parameter to construct the inner ClientHello; you would construct the inner one as if this parameter weren't there, then encapsulate it if you have an EchoConfig.

chris-wood: @MikeBishop, I don't disagree with what you said, but... are you looking for a change? Suggestions welcome. :-)

draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1091,6 +1091,11 @@ In wire format, the value of the parameter is an ECHOConfigs vector
 {{!ECHO}}, including the redundant length prefix.  In presentation format,
 the value is encoded in {{!base64=RFC4648}}.
 
+The value of this parameter and any included ECHOConfig extensions are directed
+at the outer TLS ClientHello when ECHO is in use. This parameter does not affect
+parameters of the inner TLS ClientHello. Thus, changes targeting the inner TLS
+ClientHello when ECHO is in use SHOULD use other SVCB parameters.
```
MikeBishop, line 1097:
```suggestion
construction of the inner TLS ClientHello.
```
I'm suggesting we can do without this sentence entirely.

chris-wood: Lovely -- thanks!

draft-ietf-dnsop-svcb-httpssvc.md
enygren: The wording "are directed at" tripped me up a few times when first reading this.  I wonder if "influence the construction of" or similar might be more readable? 

Would giving an illustrative example help for people not as familiar with ECHO?  ("For example, the echoconfig will specify the SNI and ALPN values to use in ClientHelloOuter but will have no impact on the SNI and ALPN values used in ClientHelloInner.")

chris-wood: > The wording "are directed at" tripped me up a few times when first reading this. I wonder if "influence the construction of" or similar might be more readable?

That's a great suggestion. We should incorporate it.

> Would giving an illustrative example help for people not as familiar with ECHO? 

Probably! @bemasc, do you want to take a crack at that? (I won't have time to get back to this PR for quite a while, I'm afraid.)

bemasc: OK, I gave it a try: https://github.com/MikeBishop/dns-alt-svc/pull/148

bemasc:
This clarification has been made in #148.

PR #148, opened 2020-05-13
Explain inner vs. outer ClientHello

bemasc:
As requested here: https://github.com/MikeBishop/dns-alt-svc/pull/145#discussion_r424102426

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -1112,7 +1112,8 @@ detail of ECHO, and its contents are controlled by the ECHOConfig in accordance
 with {{ECHO}}.  The inner ClientHello is used for establishing a connection to the
 service, so its contents may be influenced by other SVCB parameters.  For example,
 the requirements on the ProtocolNameList in {{alpn-key}} apply only to the inner
-ClientHello.
+ClientHello.  Similarly, it is the inner ClientHello whose Server Name Indication
+identifies the origin.
 
 ## Client behavior {#echo-client-behavior}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1100,6 +1100,14 @@ In wire format, the value of the parameter is an ECHOConfigs vector
...
 the value is encoded in {{!base64=RFC4648}}.
 
+When ECHO is in use, the TLS ClientHello is divided into an unencrypted "outer"
+and an encrypted "inner" ClientHello.  The outer ClientHello is an implementation
+detail of ECHO, and its contents are controlled by the ECHOConfig in accordance
+with {{ECHO}}.  The inner ClientHello is used for establishing a connection to the
+service, so its contents may be influenced by other SVCB parameters.  For example,
+the requirements on the ProtocolNameList in {{alpn-key}} apply only to the inner
+ClientHello.
+
```
enygren, line 1110: Do we want to add to the example that the ECHOConfig could specify a dummy ALPN TLS extension for the outer client hello (for middlebox compatibility) but it is ignored for connection establishment?  Similarly, explaining how SNI fits in here might be helpful to readers.  (ie, that the "outer" contains a dummy SNI from the ECHOConfig and the "inner" contains the SNI from the Origin.)

bemasc: > Do we want to add to the example that the ECHOConfig could specify a dummy ALPN TLS extension for the outer client hello (for middlebox compatibility) but it is ignored for connection establishment?

I don't think this is a good place to speculate on possible future extensions to a different draft.  (I also think that particular extension is unlikely to happen any time soon.)

> Similarly, explaining how SNI fits in here might be helpful to readers. (ie, that the "outer" contains a dummy SNI from the ECHOConfig and the "inner" contains the SNI from the Origin.)

Added.

enygren: Maybe I misread the ECHO draft, but I thought this was allowed? It's given as an example. But may be more detail than is needed here.

PR #150, opened 2020-05-21
Rename ECHO to ECH.

chris-wood:
See https://github.com/tlswg/draft-ietf-tls-esni/pull/236/files for the related ECHO change.


enygren:
(Looks good to me)

PR #155, opened 2020-05-30
Clarify interaction with WebSocket

### LINKED ISSUE #90: WebSockets should use HTTPSSVC
Despite having their own scheme (ws://, wss://), WebSockets should actually use HTTPSSVC.  This has a firm basis on the Fetch spec, which says that [each WebSocket has an equivalent http(s) URL](https://fetch.spec.whatwg.org/#websocket-opening-handshake), but it has come up as a point of confusion so I think it deserves a sentence in the draft.

bemasc:
This also covers the general case of WebSocket-like protocols.

Fixes #90

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -33,7 +33,10 @@ author:
 normative:
 
 informative:
-
+  FETCH:
+    title: "Fetch Living Standard"
+    date: May 2020
+    target: "https://fetch.spec.whatwg.org/"
 --- abstract
 
 This document specifies the "SVCB" and "HTTPSSVC" DNS resource record types to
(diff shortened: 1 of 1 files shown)


davidben:
Coincidentally, we were looking at WebSockets recently for cookies. It's a mess. :-( Whether this definition of "HTTP-based" works is... unclear. RFC 6455 does *not* map WebSockets to its corresponding http/https scheme. It believes WebSockets is an entirely independent protocol with an HTTP-looking handshake for convenience.

>   The WebSocket Protocol is an independent TCP-based protocol.  Its
>   only relationship to HTTP is that its handshake is interpreted by
>   HTTP servers as an Upgrade request.

But this is an unhelpful way to specify it and was probably a mistake. It doesn't tell you how WebSockets interacts with cookies, HTTP auth, or redirects, and led to a mess of interop issues.

Fetch, accordingly, [overrides this](https://fetch.spec.whatwg.org/#websocket-protocol) and [maps the schemes to HTTP](https://fetch.spec.whatwg.org/#websocket-opening-handshake). This is partially a no-op because cookies don't check schemes, but we should [fix that](https://tools.ietf.org/html/draft-west-cookie-incrementalism-01). At that point we'll be in a weird situation where, without the WHATWG override of the IETF spec, WebSockets' interactions with cookies will be somewhere between incorrect and insufficiently specified. :-(

How that translates to what this specification should do, I don't know. It may make sense to just special-case WebSockets and declare the RFC 6455 specification strategy was a mistake and we shouldn't do it again. Then again, that means new "webby" protocols will require separate SVCB records, which also seems poor. Every extra thing we stick in the query (scheme, port, etc.) weakens the security benefits (HTTPS upgrade, ECH).

bemasc:
@davidben Thanks for flagging the discrepancy.  I've tweaked the text to try to be more precise (without explaining the whole situation).

Issue #79, opened 2019-11-13
Clarify behavior when alpn=h2 but origin only supports http/1.1 (and vice-versa)


enygren:
We should clarify the behavior for how alpn=h2 and alpn=http/1.1 interact with ALPN negotiation over TLS with an origin, including in cases where only one is specified in the DNS but where only the other can get negotiated via TLA.

One option would be to call this out and recommend that "h2" can negotiate to http/1.1 and/or h2 if only one is specified in HTTPSSVC records.

bemasc:
I could see something like "client MUST include the specified ALPN and SHOULD include any others that it would also support".

> in cases where only one is specified in the DNS but where only the other can get negotiated

Why would this happen?

bemasc:
@MikeBishop What do you think is supposed to happen with ALPN in Alt-Svc?  If the Alt-Svc says "h2=...", can the client also offer "http/1.1" in the ClientHello?  RFC 7838 doesn't seem very clear on this.

enygren:
Synchronization between DNS configuration and HTTP/TLS server configuration seems challenging.  I suspect it is highly likely that mismatches here will happen regularly, especially as the h2 vs http/1.1 variations may be a little more subtle than QUIC.  It would be nice if things here had good usability and just worked as long as behaviors are safe and robust.

bemasc:
OK, I added text for this to #78.

MikeBishop:
As best I recall and without rechecking 7838:  With Alt-Svc, the client can offer anything, but if the connection winds up not having the properties the Alt-Svc entry told it to expect, it's not usable.  We might choose to depart from that, however.

bemasc:
Thanks for the explanation.  Please take a look at #78 for the proposed change.

MikeBishop:
Yes, RFC7838 says:
> If the connection to the alternative service does not negotiate the expected protocol (for example, ALPN fails to negotiate h2, or an Upgrade request to h2c is not accepted), the connection to the alternative service MUST be considered to have failed.

The softer stance of "SHOULD include others with equivalent security" seems reasonable, but we might get some pushback on the divergence.

bemasc:
Merged #78

PR #156, opened 2020-05-30
Clarify conditions for enabling HSTS

### LINKED ISSUE #91: Does HSTS apply if you fall back to A?
We should clarify whether the HSTS implication applies to the fallback IPs, and also clarify that it _does_ apply to AliasForm.

bemasc:
Fixes #91

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -1008,7 +1008,7 @@ all "http" scheme requests for that origin SHOULD logically be redirected
 to "https".
 
 Prior to making an "http" scheme request, the client SHOULD perform a lookup
-to determine if an HTTPSSVC record is available for that origin.  To do so,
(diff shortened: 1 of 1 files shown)


davidben:
This PR seems reasonable, though @enygren's comment has now confused me. What does fallback mean here? (Did you mean another comment of mine? Doesn't look like I've commented on #79.)

enygren:
Oops, I meant #91   (no idea why the typo)

davidben:
Ah. No, a fallback wouldn't make sense, both semantically and for security. Semantically, by the time you've done an HSTS-like upgrade, you've had to pop up many layers to trigger something like a redirect. That no longer looks like the HTTPSSVC vs. A/AAAA alternate routes. (The client needs to actually change the URL here.) And, in terms of security, if the network can undo the upgrade by blocking the HTTPS upgrade, what was the point? Maybe as well do a plain 302 redirect at that point.

PR #158, opened 2020-05-31
Exclude parentheses from basic-visible

### LINKED ISSUE #120: basic-visible vs line continuation in master files
basic-visible fails to exclude 0x28  '('   and 0x29  ')' which are line continuation characters in master files.

foo=(1)  will parse as "foo=", "1" in master files.   For it to be a single token it needs to be foo="(1)" or foo=\\(1\\)

bemasc:
Fixes #120


bemasc:
@marka63 Could you confirm that this addresses the ABNF problem in #120?

marka63:
It appears to.

Issue #99, opened 2020-01-08
Clarify Alias wire format and motivation for binary SvcFieldValue encoding


bemasc:
@miekg has asked for two related clarifications:
1. How is it that the AliasForm and ServiceForm encodings are compatible?
2. Why not reuse the TXT/SPF format?

miekg:
If this is previously discussed I am happy to read those emails.

Re 1.: the text in the 01 draft is unclear to me that these two encoding (as proposed) are in fact identical.

Re 2.: Easiest for us is to just re-use the TXT encoding, the draft talks about limit length (255 octets) per txt segment, which is true, but the total size can still be more than that. Downside is that parsing of the key, values must be done by the client, instead of the DNS client. But things like \DDD are supported within TXT as well.

See https://github.com/miekg/dns/issues/1064 for our bug to add this.

enygren:
This has been extensively discussed in the WG.  Can we close this?

Issue #82, opened 2019-11-18
Adjust CamelCase names


bemasc:
Some of the CamelCase names could be improved.  For example, SvcFieldPriority is now used in AliasForm, so maybe it should just be SvcPriority or SvcIndex.

moonshiner:
These are the CamelCase names I found

SvcDomainName
SvcFieldPriority
SvcFieldValue
SvcParamKey
SvcParamValue
SvcRecordType
AliasForm
ServiceForm
ClientHello

marka63:
Personally I'm happy with SvcFieldPriority.  If it has to change then SvcPriority.

enygren:
Of these I think only SvcFieldPriority might want to change to SvcPriority but I don't strongly care either way.

moonshiner:
SvcPriority makes more sense as it is the priority for the RR itself,

tfpauly:
Agreed that SvcFieldPriority can simply be SvcPriority.

Is there a reason we need "field" in SvcFieldValue? Is SvcValue too value? Could that be SvcParams, to indicate that its the parameter list?

moonshiner:
We should do that, but we will also have to expand the ABNF to ref;ect ot

SVCParams = SVCParam *(SVCParam)

bemasc:
I like "SvcPriority" and "SvcParams".

I wrote up a more radical proposal in #160 that gets rid of all the camel case names.  From that analysis, some comments:

* AliasForm/ServiceForm have generated some confusion because they are in fact a single form as far as parsing is concerned.  I think "mode" is probably more accurate than "form".
* "SvcDomainName" is potentially confusing, given that it appears in both ServiceForm and AliasForm, and can itself contain either ServiceForm or AliasForm records.  Also, "DomainName" seems redundant.
* SvcRecordType is vestigial.  It's now computed implicitly from SvcFieldPriority.  I would rephrase to remove it entirely.

enygren:
Perhaps "TargetName" or "TargetDomain" instead of "SvcDomainName" ?
SvcParams and SvcPriority sound good.
AliasMode/ServiceMode also seem fine if "Form" causes confusion.

moonshiner:
I like "TargetName" - it is a fully qualified domain name and not just a domain name correct ?  
That should be more explicit

tfpauly:
One thing in favor of keeping SvcDomainName is that it is the Alt-Svc name, so having "Svc" is nice.

[...]

PR #160, opened 2020-06-01
Adjust CamelCase names

bemasc:
This (purely editorial) change replaces the unambiguous but
idiosyncratic CamelCase terms with capitalized standard terms.
It also removes some duplicated language and changes "form" to "mode",
since there is really only one SVCB wire format.

Relevant issues: #99, #82

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -114,6 +114,12 @@ and introductory examples, much of this document refers only to the SVCB RR,
 but those references should be taken to apply to SVCB, HTTPS,
 and any future SVCB-compatible RR types.
 
+The SVCB RR has two modes: 1) "AliasMode" simply delegates operational
+control for a resource; 2) "ServiceMode" binds together
+configuration information for a service endpoint.
+ServiceMode provides additional key=value parameters
+within each RDATA set.
+
 TO BE REMOVED: If we use this for providing configuration for DNS
 authorities, it is likely we'd specify a distinct "NS2" RR type that is
 an instantiation of SVCB for authoritative nameserver delegation and
@@ -259,7 +265,8 @@ the Internet ("IN") Class ({{!RFC1035}}).
 
 SvcPriority is a number in the range 0-65535,
 TargetName is a domain name,
-and the SvcParams are SvcParamKey=SvcParamValue pairs separated by whitespace.
+and the SvcParams are a whitespace-separated list, with each SvcParam
+consisting of a SvcParamKey=SvcParamValue pair or a standalone SvcParamKey.
 SvcParamKeys are subject to IANA control ({{svcparamregistry}}).
 
 Each SvcParamKey SHALL appear at most once in the SvcParams.
@@ -436,7 +443,7 @@ the target of the SVCB record might offer better performance, and
 therefore would be preferable for clients implementing this specification
 to use.
 
-SVCB and SVCB-compatible record types use AliasMode only within the type.
+AliasMode records only apply to queries for the specific RR type.
 For example, a SVCB record cannot alias to an HTTPS record,
 nor vice-versa.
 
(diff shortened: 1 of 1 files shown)


bemasc:
This PR is firmly in bikeshed territory, and has no particular deadline, so it's a lower priority than any substantive PRs.

enygren:
I'm not sold yet.  It's an interesting idea, but I think there is value in using the camelcase names for unambiguous references to fields and forms.  Common terms have more chances for confusion and ambiguity.  This may be a style thing.  I'd be more in-favor of just simplifying some of the camelcase terms that are overly long.

tfpauly:
I agree with @enygren that the more common names seem a bit too generic, and liable for confusion when referenced out of context. I'd prefer to keep with the existing style.

bemasc:
I've changed this PR to match the names discussed in #82, so the initial description comment is no longer accurate.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -166,22 +160,21 @@ a non-normative manner.  (As mentioned above, this all
...
-   AliasForm) or the alternative service endpoint (for ServiceForm).
-3. SvcFieldValue (optional): A list of key=value pairs
-   describing the alternative service endpoint for the domain name specified in
-   SvcDomainName (only used in ServiceForm and otherwise ignored).
-   Described in {{svcfieldvalue}}.
+2. TargetName: The domain name of either the alias target (for
+   AliasMode) or the alternative service endpoint (for ServiceMode).
+3. SvcParams (optional): A list of key=value pairs
+   describing the alternative service endpoint at
+   TargetName (only used in ServiceMode and otherwise ignored).
```
enygren, line 177:
```suggestion
3. SvcParams (optional): A list of key=value pairs
   describing the alternative endpoint at
   TargetName (only used in ServiceMode and otherwise ignored). 
   Described in {{svcparams}}.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -967,7 +940,7 @@ origin hostname also allows the targets of existing CNAME chains
 requiring origin domains to configure and maintain an additional
 delegation.
 
-Following of HTTPS RR AliasForm and CNAME aliases is unchanged from SVCB.
+Following of HTTPS RR aliases and CNAME aliases is unchanged from SVCB.
```
enygren, line 943:
```suggestion
Following of HTTPS AliasMode RRs and CNAME aliases is unchanged from SVCB.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -114,12 +114,6 @@ and introductory examples, much of this document refers only to the SVCB RR,
 but those references should be taken to apply to SVCB, HTTPS,
 and any future SVCB-compatible RR types.
 
-The SVCB RR has two forms: 1) the "Alias Form" simply delegates operational
-control for a resource; 2) the "Service Form" binds together
-configuration information for a service endpoint.
-The Service Form provides additional key=value parameters
-within each RDATA set.
-
```
enygren, line 122: While repetitive with what is below, this does provide important background context.  Should we see if we can include this even if we condense this down some?  Or just leave this as-is?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1341,16 +1313,16 @@ ACTION: create and include a reference to this registry.
 
 A registration MUST include the following fields:
 
+* Number: Service parameter key numeric identifier (range 0-65535)
```
enygren, line 1316:
```suggestion
* Number: SvcParamKey wire format numeric identifier (range 0-65535)
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1341,16 +1313,16 @@ ACTION: create and include a reference to this registry.
...
 * Name: Service parameter key name
-* SvcParamKey: Service parameter key numeric identifier (range 0-65535)
 * Meaning: a short description
 * Pointer to specification text
 
 SvcParamKey values to be added to this namespace
```
enygren, line 1321:
```suggestion
SvcParamKey entries to be added to this namespace
```

PR #161, opened 2020-06-11
Clarify need for SNI

### LINKED ISSUE #92: Consider requiring support for SNI
Clients that do HTTPSSVC should be modern enough to do SNI.

enygren:
Fixes #92 

(text derived from rfc7838 and rfc7540 which have a similar requirement)

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -1008,9 +1008,8 @@ client supports TLS Server Name Indication (SNI) and
 indicate the origin name when negotiating TLS.
 This supports the conservation of IP addresses.
 
-Note that the SNI information provided in TLS by the client will be
-that of the origin, not the SvcDomainName (as will the Host HTTP header
-field value).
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-httpssvc.md
```diff
@@ -1001,6 +1001,17 @@ HTTPSSVC and Alt-Svc is described in {{ech-client-behavior}}.
...
+## Requiring Server Name Indication
+
+Clients MUST NOT use an HTTPSSVC response unless the
+client supports TLS Server Name Indication (SNI) and 
+indicate the origin name when negotiating TLS.
+This supports the conservation of IP addresses.
+
+Note that the SNI information provided in TLS by the client will be
+that of the origin, not the SvcDomainName (as will the Host HTTP header
+field value).
```
bemasc, line 1013:
```suggestion
Note that the TLS SNI (and also the HTTP "Host" or ":authority") will indicate 
the origin, not the SvcDomainName.
```

PR #163, opened 2020-06-11
SVCB-Used HTTP header

### LINKED ISSUE #107: Consider SVCB-Used header
If decoupling from Alt-Svc, having an alternative to Alt-Used would be valuable.  This should take lessons from challenges with Alt-Used adoption and should minimize the privacy impact.
Some options include:
1) Send SvcDomainName 
2) Send SvcFieldPriority
3) Send an (8? 16?) bit value specified in a new parameter
4) Send a variable number of bits from a value specified in a new parameter.  (ie, client can zero-out as many MSB as it would like)

I'm leaning towards \#3 above as this bounds the amount of additional entropy to be significantly less than what could be done already by using an alternate port number or IP(v4/v6) address but still allows some level of signalling without requiring servers to have to go through the complexity of needing to use distinct ports/IPs (or ESNI key IDs), all of which are possible but which leak more to passive adversaries.

enygren:
Fixes #107 
May want working group discussion in dnsop and/or httpbis

### DIFF
--- a/draft-ietf-dnsop-svcb-httpssvc.md
+++ b/draft-ietf-dnsop-svcb-httpssvc.md
@@ -987,6 +987,17 @@ record, groups of clients will necessarily receive the same
 SvcFieldValue.  Therefore, HTTPSSVC is not suitable for uses that
 require single-client granularity.
 
+## SVCB-Used header
+
+When making HTTPS requests to a server discovered through 
+HTTPSSVC records, clients SHOULD send a "SVCB-Used" HTTP 
+request header.  This is intended to help server operators
+with operational tasks and diagnostics.  The value of the
+"SVCB-Used" header SHOULD be the last SvcDomainName.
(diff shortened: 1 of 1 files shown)


bemasc:
I would choose one option for the behavior, and leave it at MAY.  I also think we probably don't want to do this unless we can convince people that it's not a privacy risk at all.

bemasc:
After giving this some more thought, I think there is enough complexity here that we shouldn't delay this draft while we figure it out.  For example, it might make sense to say that the record ID is a 16-bit value indicated by a SvcParamKey (`record-id=12345`), and that `record-id` and `port` must not appear together.  That would (arguably) make this unambiguously privacy-positive: it reduces the exposed entropy without increasing the total entropy.

I'm not sure that formulation would fly, but we should take our time and get this right.

enygren:
Proposing to defer this to a future draft.  It may make sense to do this as part of Alt-Svc-bis

PR #173, opened 2020-06-16
s/master/primary/

### LINKED ISSUE #172: We should change master to primary in Appendix A.3
FYI, I was just thinking we should be ahead of this.

enygren:
Fixes #172


moonshiner:
Thank you! I was on my to doing a pull request but you are too good to me

PR #197, opened 2020-06-23
added RRType requests in IANA Considerations

moonshiner:
Make specific IANA requests in the IANA Considerations section


moonshiner:
I updated this with the IANA Values that have been assigned.

bemasc:
This looks good but maybe we should hold off on it until after WGLC, just in case there are any last-minute changes?

tfpauly:
@bemasc the allocations are already made, so we may as well mention the values, right?

moonshiner:
We should mention the values as they are set by IANA.

PR #199, opened 2020-06-24
Incorporate comments from implementer review

bemasc:
We got some tweaks from a server implementer review
* Everything additional should go in Additional
* Authoritatives should include DNSSEC for the in-bailiwick followup records
* Harmonize #recursive-behavior with #incomplete-responses

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -596,8 +596,7 @@ Providing the proxy with the final SvcDomainName has several benefits:
 ## Authoritative servers
 
 When replying to a SVCB query, authoritative DNS servers SHOULD return
-A, AAAA, and SVCB records (as well as following any relevant CNAME or
-{{!DNAME=RFC6672}} records) in the Additional Section for any
+A, AAAA, and SVCB records in the Additional Section for any
 in-bailiwick SvcDomainNames.  If the zone is signed, the server SHOULD also
 include positive or negative DNSSEC responses for these records in the Additional
 section.
@@ -605,8 +604,8 @@ section.
 ## Recursive resolvers {#recursive-behavior}
 
 Recursive resolvers that are aware of SVCB SHOULD help the client to
-execute the procedure in {{client-behavior}} without issuing a second
-round of queries, by incorporating additional useful information into the
+execute the procedure in {{client-behavior}} with minimum overall
+latency, by incorporating additional useful information into the
 response.  For the initial SVCB record query, this is just the normal
 response construction process (i.e. unknown RR type resolution under
 {{!RFC3597}}).  For followup resolutions performed during this procedure,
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -596,20 +596,22 @@ Providing the proxy with the final SvcDomainName has several benefits:
 ## Authoritative servers
 
 When replying to a SVCB query, authoritative DNS servers SHOULD return
-A, AAAA, and SVCB records (as well as any relevant CNAME or
+A, AAAA, and SVCB records (as well as following any relevant CNAME or
 {{!DNAME=RFC6672}} records) in the Additional Section for any
```
wtoorop, line 600: Including useful records in the additional section, is called additional section processing in RFC1035.
When comparing SVCB with MX and SRV, note that MX and SRV may not refer to an alias. This is documented in https://tools.ietf.org/html/rfc2181#section-10.3 and in https://tools.ietf.org/html/rfc2782 on page 4, quoting: "The name MUST NOT be an alias".

wtoorop: I suggest to remove the text in parenthesis.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -596,20 +596,22 @@ Providing the proxy with the final SvcDomainName has several benefits:
...
 
-Recursive resolvers that are aware of SVCB SHOULD ensure that the client can
+Recursive resolvers that are aware of SVCB SHOULD help the client to
 execute the procedure in {{client-behavior}} without issuing a second
-round of queries, by incorporating all the necessary information into a
-single response.  For the initial SVCB record query, this is just the normal
+round of queries, by incorporating additional useful information into the
+response.  For the initial SVCB record query, this is just the normal
 response construction process (i.e. unknown RR type resolution under
 {{!RFC3597}}).  For followup resolutions performed during this procedure,
```
wtoorop, line 612: So, comparing this to SRV serving by Unbound: Unbound will include useful RR's in the additional section (i.e. address records for the SRV targets) when they can be found in cache.
The current text states that ``"resolve" means the resolver's ordinary recursive resolution procedure'' rather firmly (verbatim on line 637 actually), but this ignores the modus operandi of providing helpful records from cache (for which you can be certain that it will help improve performance; opposed to doing additional recursive resolution).
Getting the RRs for the additional from cache also plays along nicely with the authoritative providing the A, AAAA and SVCB records which it has in-bailiwick in the additional section.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, removed.  CNAMEs within the bailiwick would be a strange configuration anyway.

draft-ietf-dnsop-svcb-https.md
bemasc: That behavior is compatible with the text as-is, as discussed in "Generating and using incomplete responses".  However, I have adjusted the text here in an attempt to reduce confusion about the motivation and requirements.

Issue #194, opened 2020-06-23
Advice regarding construction of ALPN extension in TLS


martinthomson:
>    Clients MUST include an "application_layer_protocol_negotiation" extension in their ClientHello with a ProtocolNameList that includes at least one ID from the ALPN set.  Clients SHOULD also include any other values that they support and could negotiate on that connection with equivalent or better security properties.  For example, if the ALPN set only contains "http/1.1", the client could include "http/1.1" and "h2" in the ProtocolNameList.

This could be read to imply that a client conditions its choice of ALPN parameter in a TLS ClientHello based on what it learns from SVCB.  This is dangerous as it exposes the connection to potential downgrade attacks.  Though it tries to recommend that "better" options be chosen, this is a determination that the server makes in ALPN, not the client.  This document should instead mandate that clients construct an ALPN extension containing all protocols that they are willing to use.

MikeBishop:
I think that depends on how you define "downgrade."  The existing text says that the client includes:
- At least one ALPN token from the SVCB record(s)
- Any protocol the client supports which it considers "better" than the least-preferable selected token, even if not in the SVCB record

But the point that it leads to selection involving the client's definition of "better" versus the server's is certainly a fair point.

davidben:
Wait, I remember this "better" formulation already came up and was rejected, for precisely this reason. The network does not get to influence the selected ALPN protocol. This notion of "better" breaks any formal notion of downgrade protection. It also has practical consequences beyond downgrades: even if client and server agree that h2 is better than http/1.1, the current draft text makes h2 SVCB break if, say, the server needed to rollback h2.

I agree with Martin that this text needs to change change. Both http/1.1 and h2 SVCB records must result in the *exact* same ALPN advertisement, namely all the client's accepted protocols. The only acceptable filtering is that inherent to collections of compatible ALPNs. Beyond that, SVCB stays out of the TLS ALPN negotation.

enygren:
Merged in #201.  Good enough to close?

PR #201, opened 2020-06-26
Adjust advice on populating the client's ALPN set

bemasc:
Addresses #194

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -783,11 +783,14 @@ to connect using TLS over TCP with a ProtocolNameList of \["http/1.1", "h2"\],
 and could also attempt a connection using QUIC, with a ProtocolNameList of
 \["h3"\].
 
-This procedure prevents downgrade attacks within a transport and ensures that
+Once the client has constructed a ClientHello, protocol negotiation in that
+handshake proceeds as specified in {{!ALPN}}, without regard to the SVCB ALPN
+set.
+
+With this procedure in place, an attacker who can modify DNS and network
+traffic can prevent a successful transport connection, but cannot otherwise
+interfere with ALPN ID selection.  This procedure also ensures that
 each ProtocolNameList includes at least one ID that the server supports.
-Once the client has formulated the ClientHello, protocol negotiation
-on that connection proceeds as specified in {{!ALPN}}, without regard to the
-SVCB ALPN set.
 
 Clients SHOULD NOT attempt connection to a service endpoint whose
 ALPN set does not contain any supported IDs.  To ensure
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -736,20 +736,20 @@ support, and this informs the underlying transport protocol used (such
 as QUIC-over-UDP or TLS-over-TCP).
 
 ALPNs are identified by their registered "Identification Sequence"
-(alpn-id), which is a sequence of 1-255 octets.
+(`alpn-id`), which is a sequence of 1-255 octets.
 
     alpn-id = 1*255(OCTET)
 
 The presentation value of "alpn" is a comma-separated list of one or
-more `alpn-id`s.  Any commas present in the protocol-id are escaped
+more `alpn-id`s.  Any commas present in the Identification Sequence are escaped
 by a backslash:
 
     escaped-octet = %x00-2b / "\," / %x2d-5b / "\\" / %x5D-FF
     escaped-id = 1*(escaped-octet)
     alpn-value = escaped-id *("," escaped-id)
 
-The wire format value for "alpn" consists of at least one ALPN identifier
-(`alpn-id`) prefixed by its length as a single octet, and these length-value
+The wire format value for "alpn" consists of at least one
+`alpn-id` prefixed by its length as a single octet, and these length-value
 pairs are concatenated to form the SvcParamValue.  These pairs MUST exactly
 fill the SvcParamValue; otherwise, the SvcParamValue is malformed.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -784,14 +784,14 @@ default transports.
...
-security properties.  For example, if the ALPN set only contains "http/1.1",
+are willing to negotiate on that connection.
+For example, if the ALPN set only contains "http/1.1",
 the client could include "http/1.1" and "h2" in the ProtocolNameList.
 
 Once the client has formulated the ClientHello, protocol negotiation
 on that connection proceeds as specified in {{!ALPN}}, without regard to the
 SVCB ALPN set.  To preserve the security guarantees of this process, clients
-MUST consolidate all compatible ALPN IDs into a single ProtocolNameList.
+MUST consolidate all compatible ALPN IDs into a single ClientHello.
```
davidben, line 794: I'm not quite following this sentence (old and new version). Is it saying the same thing as the SHOULD above or something else?

bemasc: It's kind of the inverse.  There are really two requirements:
* If you use an ALPN from the list, you SHOULD accompany it with any other ALPN you would like to do on that transport, whether or not it's in the list.
* You MUST only form one ClientHello per transport in total, so if you like two ALPNs from the list, they have to go in the same ClientHello (if that is possible).

The first requirement defends against attackers on the DNS path, and the second defends against attackers on the network path.

That said, these requirements are effectively kind of redundant, so we could possibly simplify this text.  Suggestions welcome.

davidben: Ah. I guess my interpretation was that, if `ClientHelloFromSVCB(proto=h2)` and `ClientHelloFromSVCB(proto=http/1.1)` both result in the same ClientHello (based on the rule that you include all of them), that's fine whether I connect to them serially or in parallel or whatever.

What does consolidating mean if, say, I see SVCB records for h2 and http/1.1 but other properties, like ECH keys or addresses, are different? I figured the model would be that you pick which one(s) you want, and send out ClientHellos for them. If you decided you wanted to race the http/1.1 and h2 records, they'll end up sending the same ALPN. If you decided you wanted to race an h3 one in there, it will be different. When picking them, you probably do need to be aware of the correspondence, but I think the impact is less security and more making your connection strategy coherent.

bemasc: OK, evidently the previous text was not clear enough.  I've reorganized it; hopefully its less ambiguous now.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -770,8 +770,27 @@ ALPN IDs that do not uniquely identify a protocol suite (e.g. an ID that
...
+3. For each transport in this set, construct a ProtocolNameList containing all
+the client's supported IDs for that transport, without regard to the ALPN set
+from SVCB.
+
+For example, if the ALPN set is \["http/1.1", "h3"\], the client could attempt
+to connect using TLS over TCP with a ProtocolNameList of \["http/1.1", "h2"\],
+and could also attempt a connection using QUIC, with a ProtocolNameList of
+\["h3"\].
+
+This procedure prevents downgrade attacks within a transport and ensures that
```
chris-wood, line 786: FWIW, I find "downgrade attacks within a transport" somewhat confusing. Could we add an example?

bemasc: OK, I replaced that language with a more explicit definition.

chris-wood: Much clearer -- thanks!

draft-ietf-dnsop-svcb-https.md
```diff
@@ -770,8 +770,30 @@ ALPN IDs that do not uniquely identify a protocol suite (e.g. an ID that
 can be used with both TLS and DTLS) are not compatible with this
 SvcParamKey and MUST NOT be included in the ALPN set.
 
+To establish a connection to the endpoint, clients MUST
+
+1. Identify all supported IDs in the ALPN set.
```
davidben, line 775: Nitpick: Not sure I've seen these called "IDs" before. RFC7301 says "protocol" or "protocol name". RFC8446 says "ALPN protocol". RFC7838 says "ALPN protocol name" with a few instances of "ALPN protocol".

draft-ietf-dnsop-svcb-https.md
```diff
@@ -770,8 +770,30 @@ ALPN IDs that do not uniquely identify a protocol suite (e.g. an ID that
...
+2. Collect the set of transports (e.g. TLS, DTLS, QUIC) implied by those IDs.
+3. For each transport in this set, construct a ProtocolNameList containing all
+the client's supported IDs for that transport, without regard to the ALPN set
+from SVCB.
+
+For example, if the ALPN set is \["http/1.1", "h3"\], the client could attempt
```
davidben, line 781: Might want to mention in this example that this client supports ["http/1.1", "h2", "h3"]. If the client supports "h4" which also goes over QUIC, the example would be different.

draft-ietf-dnsop-svcb-https.md
bemasc, line 775: "Supported IDs" seems to be ambiguous here.  Needs clarification that this is "supported by the client".

bemasc: I've reworked the terminology to be more precise, and use "Identification Sequence", "protocol name", and "protocol" as appropriate.

draft-ietf-dnsop-svcb-https.md

draft-ietf-dnsop-svcb-https.md

draft-ietf-dnsop-svcb-https.md
```diff
@@ -759,19 +759,44 @@ empty.
...
+
+To establish a connection to the endpoint, clients MUST
+
+1. Let SVCB-ALPN-Intersection be the set of protocols in the SVCB ALPN set
+that the client supports.
+2. Let Intersection-Transports be the set of transports (e.g. TLS, DTLS, QUIC)
+implied by the protocols in SVCB-ALPN-Intersection.
+3. For each transport in Intersection-Transports, construct a ProtocolNameList
+containing the Identification Sequences of all the client's supported ALPN
+protocols for that transport, without regard to the SVCB ALPN set.
```
tfpauly, line 782: FWIW, these steps are exactly what I'm implementing in our client already, so this seems to be the most practical approach.

davidben:
I think we should say *something*. Otherwise the analogy to Alt-Svc, which prescribed the wrong thing, may cause implementers to do the wrong thing. (The wrong thing being considered compliant by confused clients results in deployment implications for a server trying to use SVCB.) But, given the point is to disavow that and get these Alt-Svc-like mechanisms out of the business of messing with ALPN as much as possible, I'm good with something shorter and lighter on details.

Though that suggestion might be a little *too* light? It's arguably not *quite* a hint because an `h3` SVCB record will (at least for now) result in different ALPN advertisement than an `h2` record due to different sets of ALPNs being "compatible". If we don't want to go down the rabbithole of defining that just yet, maybe something like:

> Note a client may not be willing to negotiate all protocols over all connections, and would advertise ALPN protocols accordingly. For example, a client using an `h3` record for a QUIC connection would not advertise `http/1.1` or `h2`. However, a client using an `h2` record for a TCP connection may both `http/1.1` and `h2`, and accept either protocol from the server.

WDYT?

enygren:
Are we comfortable with what is here and ready to merge it, or are changes still desired?
(It is looking fairly reasonable to me.)

bemasc:
I'm going to merge this as-is.  It may be overly verbose, but we seem to have rough consensus that it is correct, and perhaps we can shorten it later.

I want to be cautious about removing text here because I think there is a real risk of implementors getting it wrong in various ways, and previous drafts have resulted in differing interpretations.

Issue #192, opened 2020-06-23
Presentation of ALPN


martinthomson:
The presentation format for ALPN is not consistent with the generic presentation format.  It also includes unprintable characters.

marka63:
Apart from having to recognise "\," vs "," and having to remember which commas where escaped and which where not, APLN are still a optionally quoted string.  e.g. ALPN="x\,y,z" or ALPN=x\,y,z

martinthomson:
It is unnecessarily different.  And it's not clear that the same escaping rules apply.  In the presentation format I might include \000 to include a literal NUL, but it appears to require the inclusion of a literal NUL preceded by a backslash in the ALPN format.

If the section said "just like the presentation format, but you have to escape comma" then it might be understandable, but the differences are inexplicable.

enygren:
Larry Campbell (@lcampbel) also found this unclear, so it does seem like we should clarify.

moonshiner:
How about a couple of examples?

enygren:
Fixed in #202.  Good enough now?

PR #202, opened 2020-06-26
Rework ABNF for clarity and simpler implementation

### LINKED ISSUE #185: ABNF for values
The definition of values allows for quoting, but the ABNF doesn't capture that.

```
   escaped-char  = "\" (VCHAR / WSP)
```

This implies that "\3" is valid, where it is not.  Capturing this is not so difficult.

```
   basic-visible = %x21 / %x23-27 / %x2A-3A / %x3C-5B / %x5D-7E
   unescaped = %x21 / %x23-27 / %x2A-2F / %x3C-5B / %x5D-7E
   escaped-number = (%x30-31 2DIGIT) / (%x32 (%x30-34 DIGIT) / (%x35 %x30-35))
   escaped-char  = "\" (unescaped / WSP / escaped-number)
```

bemasc:
Fixes #185, #192

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -266,8 +266,8 @@ The SvcFieldValue is empty for the AliasForm.
 
 ### Presentation format for SvcFieldValue {#svcfieldvalue}
 
-In ServiceForm, the SvcFieldValue is an associative array of keys
-and values.
+In presentation format, the SvcFieldValue is represented as zero or more
+key=value pairs separated by whitespace.
 
 Keys are IANA-registered SvcParamKeys ({{svcparamregistry}})
 with both a lower-case string representation and
@@ -275,26 +275,26 @@ a numeric representation in the range 0-65535.
 Registered key names should only contain characters from the ranges
 "a"-"z", "0"-"9", and "-".  In ABNF {{!RFC5234}},
 
-    alpha-lc    = %x61-7A   ;  a-z
-    key         = 1*(alpha-lc / DIGIT / "-")
+    alpha-lc = %x61-7A   ;  a-z
+    key      = 1*(alpha-lc / DIGIT / "-")
+    pair     = key ["=" char-string]
+    value    = *OCTET
 
-The presentation format for the SvcFieldValue is a whitespace-separated
-list of `key=value-list` pairs.  When the "=" is omitted, the `value-list`
-is interpreted as empty.
+The definition of each key indicates that it is either single-valued or
+multi-valued.  To parse a single-valued key, the parser applies the
+character-string decoding algorithm ({{decoding}}), producing a `value`,
+and then performs key-specific processing to validate the input and produce
+the wire-format encoding.  To parse a multi-valued key, the parser applies
+the value-list decoding algorithm to the `char-string` ({{value-list}}),
+splitting on commas to produce a list of zero or more values.
 
-    pair  = key ["=" value-list]
-
-The `value-list` is a comma-separated list of values, represented in the zone
-file with appropriate escaping ({{decoding}}).  To decode the `value-list`, the
-implementation applies the value-list decoding algorithm , and then performs
-key-specific processing to validate the input and produce the wire-format
-encoding.  The initial keys and their encodings are defined in {{keys}}.
+When the "=" is omitted, the `value` or value list is interpreted as empty.
 
 Unrecognized keys are represented in presentation
 format as "keyNNNNN" where NNNNN is the numeric
 value of the key type without leading zeros.
-The `value-list` corresponding to a key in this form
-SHALL contain a single value, which is the wire format encoded value.
+Keys in this form are always treated as single-valued, and
+the decoded `value` SHALL be used as its wire format encoding.
 
 Pairs in presentation format MAY appear in any order, but keys MUST NOT be
 repeated.
@@ -720,7 +720,8 @@ ALPNs are identified by their registered "Identification Sequence"
 
     alpn-id = 1*255OCTET
 
-Each decoded value in the "alpn" value list SHALL be an `alpn-id`.
+"alpn" is a multi-valued key.  Each decoded value in the "alpn" value list
+SHALL be an `alpn-id`.  The value list MUST NOT be empty.
 
 The wire format value for "alpn" consists of at least one ALPN identifier
 (`alpn-id`) prefixed by its length as a single octet.  These length-value
@@ -773,8 +774,8 @@ The "port" SvcParamKey defines the TCP or UDP port
 that should be used to contact this alternative service.
 If this key is not present, clients SHALL use the origin server's port number.
 
-The value list of "port" SHALL contain a single value: a decimal number
-between 0 and 65535 in ASCII.  Any other value (e.g. an empty list)
+The presentation `value` of the SvcParamValue is a single decimal number
+between 0 and 65535 in ASCII.  Any other `value` (e.g. an empty value)
 is a syntax error.
 
 The wire format of the SvcParamValue
@@ -1025,7 +1026,7 @@ The SVCB "echconfig" parameter is defined for
 conveying the ECH configuration of an alternative service.
 In wire format, the value of the parameter is an ECHConfigs vector
 {{!ECH}}, including the redundant length prefix.  In presentation format,
-the value is encoded in {{!base64=RFC4648}}.
+the value is a single ECHConfigs encoded in {{!base64=RFC4648}}.
 
 When ECH is in use, the TLS ClientHello is divided into an unencrypted "outer"
 and an encrypted "inner" ClientHello.  The outer ClientHello is an implementation
@@ -1269,13 +1270,9 @@ and others for their feedback and suggestions on this draft.
 
 # Decoding text in zone files {#decoding}
 
-In order to represent lists of values in zone files, this specification uses
-an extended version of the `<character-string>` decoding algorithm described
-in Section 5.1 of {{RFC1035}}.  The algorithm used here is identical except
-for the use of "," as a delimiter.  When "," is not escaped by a preceding
-"\\", it separates values in the output, which is a list.  We refer to this
-modified procedure as "value-list decoding".
-
+DNS zone files are capable of representing arbitrary octet sequences in
+basic ASCII text, using various delimiters and encodings.  The algorithm
+for decoding these character-strings is defined in Section 5.1 of {{RFC1035}}.
 Here we summarize the allowed input to that algorithm, using ABNF:
 
     ; non-special is VCHAR minus DQUOTE, ";", "(", ")", and "\".
@@ -1289,13 +1286,37 @@ Here we summarize the allowed input to that algorithm, using ABNF:
     escaped     = "\" ( non-digit / dec-octet )
     contiguous  = 1*( non-special / escaped )
     quoted      = DQUOTE *( contiguous / ( ["\"] WSP ) ) DQUOTE
-    value-list  = contiguous / quoted
+    char-string = contiguous / quoted
 
-Decoding a `value-list` produces 0 or more `decoded-val`s as output.
+The decoding algorithm allows `char-string` to represent any `*OCTET`.
+In this document, this algorithm is referred to as "character-string decoding".
+The algorithm is the same as used by `<character-string>` in RFC 1035,
+although the output length in this document is not limited to 255 octets.
 
-    decoded-val = 1*OCTET
+## Decoding a value list {#value-list}
 
-For simplicity, empty values are not allowed.
+In order to represent lists of values in zone files, this specification uses
+an extended version of character-string decoding that adds the use of ","
+as a delimiter.  When "," is not escaped by a preceding "\\", it separates
+values in the output, which is a list of 1*OCTET.  (For simplicity, empty
+values are not allowed.)  We refer to this modified procedure as "value-list
+decoding".
+
+    value-list = char-string
+    list-value = 1*OCTET
+
+For example, consider this `char-string`:
+
+    "part1,part2\,part3"
+
+Character-string decoding would produce a single `*OCTET` output:
+
+    part1,part2,part3
+
+Value-list decoding would instead convert it to a list of two `list-value`s:
+
+    part1
+    part2,part3
 
 # Comparison with alternatives
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -270,51 +270,35 @@ In ServiceForm, the SvcFieldValue consists of zero or more elements separated
 by whitespace.  Each element represents a key=value pair.
 
 Keys are IANA-registered SvcParamKeys ({{svcparamregistry}})
-with both a case-insensitive string representation and
+with both a lower-case string representation and
 a numeric representation in the range 0-65535.
 Registered key names should only contain characters from the ranges
 "a"-"z", "0"-"9", and "-".  In ABNF {{!RFC5234}},
 
-    ALPHA-LC    = %x61-7A   ;  a-z
-    key         = 1*(ALPHA-LC / DIGIT / "-")
-    display-key = 1*(ALPHA / DIGIT / "-")
+    alpha-lc    = %x61-7A   ;  a-z
+    key         = 1*(alpha-lc / DIGIT / "-")
 
-Values are in a format specific to the SvcParamKey.
-Their definition should specify both their presentation format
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -736,20 +720,31 @@ support, and this informs the underlying transport protocol used (such
...
+by their decimal escape sequence ("\044") instead.
+Thus, a valid `pair` for the "alpn" key could look like
 
-    alpn-id = 1*255(OCTET)
+    alpn="example1,\"weird\\044 why?\",example2"
 
-The presentation value of "alpn" is a comma-separated list of one or
-more `alpn-id`s.  Any commas present in the protocol-id are escaped
-by a backslash:
+This would be decoded as three `alpn-id`s: "example1", "weird, why?", and
```
marka63, line 741: The double quotes are part of the second alpn-id. so this should read.

This would be decoded as three `alpn-id`s: 'example1', '"weird, why?"', and
'example2'.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -736,20 +720,31 @@ support, and this informs the underlying transport protocol used (such
...
+(`alpn-id`), which is a sequence of 1-255 octets.
+
+    alpn-id = 1*255OCTET
+
+The presentation `value` of "alpn" is a comma-separated list of one or
+more `alpn-id`s, each represented as a character-string:
+
+    alpn-value = char-string *("," char-string)
+
+To decode an `alpn-value`, the parser MUST split the `value` at each ","
```
marka63, line 732: Why doesn't this just say "unescaped comma"?  \044 really does not buy anything over "\," when parsing.   So is '\,' now also a separator or is it a syntax error?

draft-ietf-dnsop-svcb-https.md
bemasc: In this PR, this example is first parsed to a sequence of three `char-string`s: `example1`, `"weird\044 why?"`, and `example2`.  Then the character-string decoding algorithm is applied to each `char-string`, yielding `example1`, `weird, why?`, and `example2`.  This is because double-quotes are optional in `char-string`.

My goal here is to make the parsing algorithm easy to implement, so please suggest changes.  I proposed this formulation to minimize the amount of new parsing code required (reuse the same decoding algorithm), while making sure that all `alpn-id`s are representable, including those that contain commas and spaces.

draft-ietf-dnsop-svcb-https.md
bemasc: The proposed algorithm here is "character-string decode, split on commas, character-string decode again".  Therefore:
* `alpn=ex1,ex2\,ex3` is `ex1`, `ex2`, and  `ex3`, because the first-pass decoding unescapes the entire value
* `alpn=ex1,ex2\\,ex3` is presumably a syntax error, because `example2\` is not a valid character-string, because `\` is supposed to escape something but there is nothing to escape.
* `alpn=ex1,ex2\\\\,ex3` is `ex1`, `ex2\`, and `ex3`

marka63: If you do this then "foo\044bar" becomes "foo,bar"  which becomes "foo" and "bar".   Following your method you would have to have "foo\\\\044bar" which becomes "foo\044bar" which becomes "foo,bar".   And to get a alpn-id that is "foo\044bar" I would need to enter "foo\\\\\\\\044bar"  (4 backslashes) instead of "foo\\\\044bar" (2 backslashes).

All this for not extending the existing character string parser to know if comma is a special or not.  'if (c == ',' && !escaped) end-of-alpn-id;".

draft-ietf-dnsop-svcb-https.md
marka63: Detecting escaped commas is dead easy.  GO LOOK AT HOW BIND DOES IT.  All this just makes it harder for the person entering the data.

marka63: Also the point of using quoted strings is to turn off special processing to get the full token at places that would otherwise terminate the token.   Once you have the full token you don't need to drop them a second time.

marka63: And so you don't have to search around in the repository to find it. 

At this point the double quotes have been processed.  The token still contains all the escapes.

```
+static isc_result_t
+alpn_fromtxt(isc_textregion_t *source, isc_buffer_t *target) {
+       isc_textregion_t source0 = *source;
+       do {
+               RETERR(commatxt_fromtext(&source0, true, target));
+       } while (source0.length != 0);
+       return (ISC_R_SUCCESS);
+}
```

```
@@ -1390,7 +1432,8 @@ name_length(const dns_name_t *name) {
 }
 
 static isc_result_t
-txt_totext(isc_region_t *source, bool quote, isc_buffer_t *target) {
+commatxt_totext(isc_region_t *source, bool quote, bool comma,
+               isc_buffer_t *target) {
        unsigned int tl;
        unsigned int n;
        unsigned char *sp;
@@ -1435,10 +1478,12 @@ txt_totext(isc_region_t *source, bool quote, isc_buffer_t *target) {
                /*
                 * Escape double quote and backslash.  If we are not
                 * enclosing the string in double quotes also escape
-                * at sign and semicolon.
+                * at sign and semicolon unless comma is set.  If
+                * comma is set then escape commas.
                 */
-               if (*sp == 0x22 || *sp == 0x5c ||
-                   (!quote && (*sp == 0x40 || *sp == 0x3b))) {
+               if (*sp == 0x22 || *sp == 0x5c || (comma && *sp == 0x2c) ||
+                   (!comma && !quote && (*sp == 0x40 || *sp == 0x3b)))
+               {
                        if (tl < 2) {
                                return (ISC_R_NOSPACE);
                        }
@@ -1465,9 +1510,14 @@ txt_totext(isc_region_t *source, bool quote, isc_buffer_t *target) {
 }
 
 static isc_result_t
-txt_fromtext(isc_textregion_t *source, isc_buffer_t *target) {
+txt_totext(isc_region_t *source, bool quote, isc_buffer_t *target) {
+       return (commatxt_totext(source, quote, false, target));
+}
+
+static isc_result_t
+commatxt_fromtext(isc_textregion_t *source, bool comma, isc_buffer_t *target) {
        isc_region_t tregion;
-       bool escape;
+       bool escape, seen_comma = false;
        unsigned int n, nrem;
        char *s;
        unsigned char *t;
@@ -1496,6 +1546,10 @@ txt_fromtext(isc_textregion_t *source, isc_buffer_t *target) {
        }
        while (n-- != 0) {
                c = (*s++) & 0xff;
+               if (comma && !escape && c == ',') {
+                       seen_comma = true;
+                       break;
+               }
                if (escape && (d = decvalue((char)c)) != -1) {
                        c = d;
                        if (n == 0) {
@@ -1534,11 +1588,32 @@ txt_fromtext(isc_textregion_t *source, isc_buffer_t *target) {
        if (escape) {
                return (DNS_R_SYNTAX);
        }
+       if (comma) {
+               /*
+                * Disallow empty ALPN at start or in the middle.
+                */
+               if (s == source->base || (seen_comma && s == source->base + 1))
+               {
+                       return (DNS_R_SYNTAX);
+               }
+               isc_textregion_consume(source, s - source->base);
+               /*
+                * Disallow empty ALPN at end.
+                */
+               if (seen_comma && source->length == 0) {
+                       return (DNS_R_SYNTAX);
+               }
+       }
        *tregion.base = (unsigned char)(t - tregion.base - 1);
        isc_buffer_add(target, *tregion.base + 1);
        return (ISC_R_SUCCESS);
 }

 +static isc_result_t
+txt_fromtext(isc_textregion_t *source, isc_buffer_t *target) {
+       return (commatxt_fromtext(source, false, target));
+}
+
 static isc_result_t
 txt_fromwire(isc_buffer_t *source, isc_buffer_t *target) {
        unsigned int n;
```

bemasc: OK, I've updated the PR to reflect this approach.  Please review.

marka63: I would say that they are key value pairs.  If the definition of a value consists of a list (i.e. ipv4hints, ipv6hints, alpn) the list elements are separated by a unescaped ",".  To add a literal comma to an element in a list, the comma must be escaped.   At the moment you have everything defined to be a list including the generic form which clearly is not a list.

bemasc: OK, I've changed the text to reflect this idea.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -264,59 +264,39 @@ and SvcFieldValue is a set of key=value pairs present for the ServiceForm.
...
-Their definition should specify both their presentation format
-and wire encoding (e.g., domain names, binary data, or numeric values).
-The initial keys and formats are defined in {{keys}}.
+The definition of each key indicates that it is either single-valued or
+multi-valued.  To parse a single-valued key, the parser applies the
+character-string decoding algorithm ({{decoding}}), producing a `value`,
+and then performs key-specific processing to validate the input and produce
+the wire-format encoding.  To parse a multi-valued key, the parser applies
+the value-list decoding algorithm to the `char-string` ({{value-list}}),
+splitting on commas to produce a list of zero or more values.
```
marka63, line 289: s/splitting on commas/splitting on unescaped commas/

draft-ietf-dnsop-svcb-https.md
```diff
@@ -799,9 +774,9 @@ The "port" SvcParamKey defines the TCP or UDP port
 that should be used to contact this alternative service.
 If this key is not present, clients SHALL use the origin server's port number.
 
-The presentation format of the SvcParamValue is a numeric value
-between 0 and 65535 inclusive.  Any other values (e.g. the empty value)
-are syntax errors.
+The presentation `value` of the SvcParamValue is a single decimal number
+between 0 and 65535 in ASCII.  Any other `value` (e.g. an empty value)
+is a syntax error.
```
marka63, line 779: It would be useful to never have to apply escape processing to port so append.  "No escape processing should be applied to the port value."

draft-ietf-dnsop-svcb-https.md
```diff
@@ -843,6 +818,9 @@ addresses in response to the SvcDomainName query. Failure to use A and/or
 AAAA response addresses could negatively impact load balancing or other
 geo-aware features and thereby degrade client performance.
 
+Each decoded value in the value list SHALL be an IP address of the appropriate
+family in standard textual format {{!RFC5952}}.
```
marka63, line 822: Similarly disallow escape processing for ipv4hints and ipv6hints.

draft-ietf-dnsop-svcb-https.md

draft-ietf-dnsop-svcb-https.md

lcampbel:
I think these changes do make the text clearer. But there are two different ways to implement a parser for this. The suggested approach is essentially to have an alternate character-string parser that (1) splits on unescaped commas, and (2) returns an array of strings rather than a string. I think an equally appealing approach is to first use a character-string parser that always returns a string and ignores backslash escapes other than \". All other uses (such as TXT records) would unescape the returned string; the alpn parser would first split the returned string on unescaped commas, then unescape each component string. I believe these two approaches are functionally identical but I find the latter a bit cleaner.

bemasc:
@lcampbel Thanks for the review.  I think that would indeed be a fine way to implement, but describing that procedure in ABNF seems pretty complicated, especially due to the many ambiguities of RFC 1035.  (I think it would be great if someone would write a followup to 1035 that pins down the precise zone file format rules, but this draft is already further down that road than I would like.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1409,6 +1387,55 @@ and others for their feedback and suggestions on this draft.
...
+    quoted      = DQUOTE *( contiguous / ( ["\"] WSP ) ) DQUOTE
+    char-string = contiguous / quoted
+
+The decoding algorithm allows `char-string` to represent any `*OCTET`.
+In this document, this algorithm is referred to as "character-string decoding".
+The algorithm is the same as used by `<character-string>` in RFC 1035,
+although the output length in this document is not limited to 255 octets.
+
+## Decoding a value list {#value-list}
+
```
enygren, line 1415: It took me a few times reading this to really understand it.  Not sure ways to clarify?
Corner cases:
* What about \\44 ?  (encoded comma)
* DQUOTE
Perhaps have an example which shows how these interact?
For example, that:
"part1,part2","part3" 
still gets decoded as all three values?
And that "part1\\44part2,part3" gets decoded as two values?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -269,49 +269,30 @@ consisting of a SvcParamKey=SvcParamValue pair or a standalone SvcParamKey.
...
 
-    ALPHA-LC    = %x61-7A   ;  a-z
-    SvcParamKey = 1*(ALPHA-LC / DIGIT / "-")
-    display-key = 1*(ALPHA / DIGIT / "-")
-
-Values are in a format specific to the SvcParamKey.
-Their definition should specify both their presentation format
-and wire encoding (e.g., domain names, binary data, or numeric values).
-The initial keys and formats are defined in {{keys}}.
-When the value is omitted, or both the value and
```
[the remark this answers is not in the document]
bemasc: Done (with slight modifications)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1122,7 +1100,7 @@ The SVCB "echconfig" parameter is defined for
 conveying the ECH configuration of an alternative service.
 In wire format, the value of the parameter is an ECHConfigs vector
 {{!ECH}}, including the redundant length prefix.  In presentation format,
-the value is encoded in Base64 {{!BASE64=RFC4648}}.
+the value is a single ECHConfigs encoded in Base64 {{!base64=RFC4648}}.
```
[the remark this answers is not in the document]
bemasc: I added an explanation.  (I think the real reason is that we don't expect TLS stacks to gain support for producing DNS zone file escape codes.)  I also added the escape restriction.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, clarified relationship with \044 and quoting.

> "part1,part2","part3"

This is not a valid `char-string` so it's excluded.

I admit that this description isn't perfect, but there is no clear syntax definition for zone files so I don't think we should aim for extreme precision.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1409,6 +1387,55 @@ and others for their feedback and suggestions on this draft.
...
+values in the output, which is a list of 1*OCTET.  (For simplicity, empty
+values are not allowed.)  We refer to this modified procedure as "value-list
+decoding".
+
+    value-list = char-string
+    list-value = 1*OCTET
+
+For example, consider this `char-string`:
+
+    "part1,part2\,part3"
```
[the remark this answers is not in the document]
bemasc: I think this is a difference between github-markdown and the flavor we're using.  In our flavor, "\" doesn't need to be escaped inside a verbatim block (four space indent).

draft-ietf-dnsop-svcb-https.md
```diff
@@ -269,49 +269,30 @@ consisting of a SvcParamKey=SvcParamValue pair or a standalone SvcParamKey.
...
-
-When decoding values of unrecognized keys in the presentation format:
-
-* a character other than "\\" represents its ASCII value in wire format.
-* the character "\\" followed by three decimal digits, up to 255, represents
-  an octet in the wire format.
-* the character "\\" followed by any allowed character, except a decimal digit,
-  represents the subsequent character's ASCII value.
+Keys in this form are always treated as single-valued, and
+the decoded `value` SHALL be used as its wire format encoding.
```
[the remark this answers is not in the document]
bemasc: OK, I've tried to do this more consistently.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1409,6 +1398,59 @@ and others for their feedback and suggestions on this draft.
...
+as a delimiter after double-quote processing.  When "," is not escaped
+(by a preceding "\\" or as the escape sequence "\\044"), it separates
+values in the output, which is a list of 1*OCTET.  (For simplicity, empty
+values are not allowed.)  We refer to this modified procedure as "value-list
+decoding".
+
+    value-list = char-string
+    list-value = 1*OCTET
+
+For example, consider this `char-string`:
```
enygren, line 1438: Make plural:
```suggestion
For example, consider these `char-string` SvcParamValues:
```

PR #204, opened 2020-06-28
Remove use of "origin" for non-HTTP schemes

### LINKED ISSUE #183: Definition of "origin server"
> The "origin server" is the server that the client would reach when accessing the origin in the absence of the SVCB record or an HTTPS Alt-Svc.

HTTP has a definition of origin server that this is not compatible with.  In HTTP an "origin server" is a server that is authoritative.  As the goal of this document is to define alternative ways to reach a server that is authoritative, the definition this document uses is very confusing.

I don't have a solid alternative at this point, but I might suggest that the notion of "whatever you get when this record isn't used" isn't a very useful distinction.  As a hypothetical, consider a protocol that only uses SVCB derivatives for discovery, which would obviously have no such concept.  It might be better to dispense with the notion of "original server" entirely.

bemasc:
The use of "origin" for non-HTTP schemes, and the use of "origin server"
to mean only the normal server, were generating confusion.  This PR
aligns terminology with the URI specification and limits the use of
"origin" to HTTP.

Fixes #183

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -42,8 +42,8 @@ informative:
 This document specifies the "SVCB" and "HTTPS" DNS resource record (RR)
 types to facilitate the lookup of information needed to make connections
 to network services, such as for HTTPS URLs.  SVCB records
-allow a service to be provided from multiple network
-locations, each with associated parameters (such as transport protocol
+allow a service to be provided from multiple alternative endpoints, 
+each with associated parameters (such as transport protocol
 configuration and keys for encrypting the TLS ClientHello).  They also
 enable aliasing of apex domains, which is not possible with CNAME.
 The HTTPS RR is a variation of SVCB for HTTPS and HTTP origins.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -131,7 +131,7 @@ See {{?I-D.tapril-ns2}} as one example.
 The goal of the SVCB RR is to allow clients to resolve a single
 additional DNS RR in a way that:
 
-* Provides endpoints that are authoritative for the service,
+* Provides alternative endpoints that are authoritative for the service,
   along with parameters associated with each of these endpoints.
 * Does not assume that all alternative endpoints have the same parameters
   or capabilities, or are even
@@ -194,7 +194,7 @@ records in the Additional Section to responses for a SVCB query.
 When in the ServiceForm, the SvcFieldValue of the SVCB RR
 provides an extensible data model for describing alternative
 endpoints that are authoritative for the origin, including
-parameters associated with each of these endpoints.
+parameters associated with each of these alternative endpoints.
 
 For the HTTPS use-case, the HTTPS RR enables many of the benefits of {{?AltSvc=RFC7838}}
 without waiting for a full HTTP connection initiation (multiple roundtrips)
@@ -220,7 +220,7 @@ hostname as the `host`.
 * The "authority address" is the authority's hostname and a port number implied
   by the scheme or specified in the URI.
 * An "alternative endpoint" is a hostname, port number, and other associated
-  instructions to the client on how to reach the service.
+  instructions to the client on how to reach an instance of service.
 
 Additional DNS terminology intends to be consistent
 with {{?DNSTerm=RFC8499}}.
@@ -507,20 +507,19 @@ load-balancing.
 
 # Client behavior {#client-behavior}
 
-An SVCB-aware client resolves a service with hostname $HOST by attempting
-to determine
-the preferred SvcFieldValue and IP addresses for its service, using the
-following procedure:
+A SVCB-aware client resolves a service with hostname $HOST by attempting
+to determine endpoint (the preferred SvcFieldValue and IP addresses for its service),
+using the following procedure:
 
 1. Let $SVCB_QNAME be $HOST, plus appropriate prefixes for the scheme
-   (see {{svcb-names}}).
+   (see {{svcb-names}}).  Let $HOST_QNAME be $HOST.
 
-2. Issue parallel AAAA/A queries for $HOST and a SVCB query for $SVCB_QNAME.
+2. Issue parallel AAAA/A queries for $HOST_QNAME and a SVCB query for $SVCB_QNAME.
    The answers for these may or may not include CNAME pointers
    before reaching one or more of these records.
 
 3. If a SVCB record of AliasForm SvcRecordType is returned for $SVCB_QNAME,
-   clients MUST set $HOST and $SVCB_QNAME to SvcDomainName (without
+   clients MUST set $HOST_QNAME and $SVCB_QNAME to SvcDomainName (without
    additional prefixes) and loop back to step 2,
    subject to chain length limits and loop detection heuristics (see
    {{client-failures}}).
@@ -531,8 +530,8 @@ following procedure:
    SvcDomainName if they are not already available.  These are the
    preferred SvcFieldValue and IP addresses.  If the connection fails, the
    client MAY try to connect using values from a lower-priority record.
-   If none of the options succeed, the client SHOULD connect to the authority
-   address directly.
+   If none of the options succeed, the client SHOULD connect to $HOST_QNAME directly.
+   If that fails, the client SHOULD fall back to $HOST.
 
 5. If a SVCB record for $SVCB_QNAME does not exist, the received AAAA and/or A
    records are the preferred IP addresses and there is no SvcFieldValue.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -216,8 +216,10 @@ access a resource identified by a URI whose `authority` field contains a DNS
 hostname as the `host`.
 
 * The "service" is the information source identified by the `authority` and
-  `scheme` of the URI, capable of providing access to the resource.
-* The "authority address" is the authority's hostname and a port number implied
+  `scheme` of the URI, capable of providing access to the resource.  For HTTPS
+  URIs, the "service" corresponds to an HTTPS "origin" {{?RFC6454}}.
+* The "service name" is the `host` portion of the authority.
+* The "authority endpoint" is the authority's hostname and a port number implied
   by the scheme or specified in the URI.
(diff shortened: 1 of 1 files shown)


bemasc:
@martinthomson Does this address your concern related to "origin server"?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -211,20 +211,16 @@ keys. See {{echconfig}}.
...
-  origin over a specified protocol.
-
-For example within HTTPS, the origin consists of a scheme (typically
-"https"), a hostname, and a port (typically "443").
+* The "service" is the information source identified by the `authority` and
+  `scheme` of the URI, capable of providing access to the resource.
+* The "authority address" is the authority's hostname and a port number implied
+  by the scheme or specified in the URI.
+* An "alternative endpoint" is a hostname, port number, and other associated
+  instructions to the client on how to reach the service.
```
enygren, line 223:
```suggestion
* An "alternative endpoint" is a hostname, port number, and other associated
  instructions to the client on how to reach an instance of service.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -211,20 +211,16 @@ keys. See {{echconfig}}.
...
-  For services other than HTTPS, the exact definition will
-  need to be provided by the document mapping that service
-  onto the SVCB RR.
-* The "origin server" is the server that the client would reach when
-  accessing the origin in the absence of the SVCB record
-  or an HTTPS Alt-Svc.
-* An "alternative service" is a different server that can serve the
-  origin over a specified protocol.
-
-For example within HTTPS, the origin consists of a scheme (typically
```
enygren, line 226: Do we want to keep this illustrative example but reword it?  I think having an example here would help.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -211,20 +211,16 @@ keys. See {{echconfig}}.
 
 ## Terminology
```
enygren, line 213: Add "service name"?  (The hostname within the authority.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
 
 # Client behavior {#client-behavior}
 
-An SVCB-aware client resolves an origin HOST by attempting to determine
+An SVCB-aware client resolves a service with hostname $HOST by attempting
```
enygren, line 506:
```suggestion
A SVCB-aware client resolves a service with hostname $HOST by attempting
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
...
+to determine
 the preferred SvcFieldValue and IP addresses for its service, using the
 following procedure:
```
enygren, line 509:
```suggestion
to determine endpoint (the preferred SvcFieldValue and IP addresses for its service),
using the following procedure:
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
...
+4. If one or more SVCB records of ServiceForm SvcRecordType are returned
+   for $SVCB_QNAME, clients should select the highest-priority option with
    acceptable parameters, and resolve AAAA and/or A records for its
    SvcDomainName if they are not already available.  These are the
    preferred SvcFieldValue and IP addresses.  If the connection fails, the
    client MAY try to connect using values from a lower-priority record.
-   If none of the options succeed, the client SHOULD connect to the origin
-   server directly.
+   If none of the options succeed, the client SHOULD connect to the authority
+   address directly.
```
enygren, line 531: Should it connect to the authority address or $HOST_QNAME?  The latter implies that the AliasForm SVCB record should be ignored if ServiceForm SVCB records were present and didn't work.

enygren: (This is an interesting question which may want to be its own issue for discussion.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
...
 
-4. If a SVCB record for HOST does not exist, the received AAAA and/or A
+5. If a SVCB record for $SVCB_QNAME does not exist, the received AAAA and/or A
    records are the preferred IP addresses and there is no SvcFieldValue.
```
enygren, line 534:
```suggestion
5. If a SVCB record for $SVCB_QNAME does not exist, the received AAAA and/or A
   records from $HOST_QNAME are the preferred IP addresses 
   and there is no SvcFieldValue.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
...
 
-An SVCB-aware client resolves an origin HOST by attempting to determine
+An SVCB-aware client resolves a service with hostname $HOST by attempting
+to determine
 the preferred SvcFieldValue and IP addresses for its service, using the
 following procedure:
 
-1. Issue parallel AAAA/A and SVCB queries for the name HOST.
+1. Let $SVCB_QNAME be $HOST, plus appropriate prefixes for the scheme
+   (see {{svcb-names}}).
```
enygren, line 512:
```suggestion
1. Let $SVCB_QNAME be $HOST, plus appropriate prefixes for the scheme
   (see {{svcb-names}}).  Let $HOST_QNAME be $HOST.
```



Separate out $SVCB_HOST so we don't mutate $HOST in step 3?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
...
+
+2. Issue parallel AAAA/A queries for $HOST and a SVCB query for $SVCB_QNAME.
```
enygren, line 514:
```suggestion
2. Issue parallel AAAA/A queries for $HOST_QNAME and a SVCB query for $SVCB_QNAME.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -506,32 +503,37 @@ load-balancing.
...
    The answers for these may or may not include CNAME pointers
    before reaching one or more of these records.
 
-2. If a SVCB record of AliasForm SvcRecordType is returned for HOST,
-   clients MUST loop back to step 1 replacing HOST with SvcDomainName,
+3. If a SVCB record of AliasForm SvcRecordType is returned for $SVCB_QNAME,
+   clients MUST set $HOST and $SVCB_QNAME to SvcDomainName (without
```
enygren, line 519:
```suggestion
   clients MUST set $HOST_QNAME and $SVCB_QNAME to SvcDomainName (without
```

draft-ietf-dnsop-svcb-https.md
enygren, line 531:
```suggestion
   If none of the options succeed, the client SHOULD connect to $HOST_QNAME directly.
   If that fails, the client SHOULD fall back to $HOST.
```
(SHOULD or MAY for the final fall back to $HOST?)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -211,20 +211,16 @@ keys. See {{echconfig}}.
...
-  accessing the origin in the absence of the SVCB record
-  or an HTTPS Alt-Svc.
-* An "alternative service" is a different server that can serve the
-  origin over a specified protocol.
-
-For example within HTTPS, the origin consists of a scheme (typically
-"https"), a hostname, and a port (typically "443").
+* The "service" is the information source identified by the `authority` and
+  `scheme` of the URI, capable of providing access to the resource.
+* The "authority address" is the authority's hostname and a port number implied
```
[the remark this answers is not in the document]
enygren: Given how we use this, either "default endpoint" or "implicit endpoint" might be better.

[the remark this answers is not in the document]
bemasc: How about "authority endpoint"?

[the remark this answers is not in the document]
enygren: That's better.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -211,20 +211,16 @@ keys. See {{echconfig}}.
...
+hostname as the `host`.
 
-* An "origin" is an information source as in {{!RFC6454}}.
-  For services other than HTTPS, the exact definition will
-  need to be provided by the document mapping that service
-  onto the SVCB RR.
-* The "origin server" is the server that the client would reach when
-  accessing the origin in the absence of the SVCB record
-  or an HTTPS Alt-Svc.
-* An "alternative service" is a different server that can serve the
```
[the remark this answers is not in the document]
enygren: It sounds like "authority endpoint"?

enygren:
This is looking good to me, but I'll try and re-read it again later today.
I think it would be good for @MikeBishop to review this one as well.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -41,9 +41,9 @@ informative:
 
 This document specifies the "SVCB" and "HTTPS" DNS resource record (RR)
 types to facilitate the lookup of information needed to make connections
-for origin resources, such as for HTTPS URLs.  SVCB records
-allow an origin to be served from multiple network
-locations, each with associated parameters (such as transport protocol
+to network services, such as for HTTPS URLs.  SVCB records
```
MikeBishop, line 44:
```suggestion
to network services, such as for HTTPS origins.  SVCB records
```
Not text changed by this PR, but I'd probably talk about the HTTPS origin as one of the services being described, rather than the URL.  In fact, this is explicitly stated further down.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -131,9 +131,9 @@ See {{?I-D.tapril-ns2}} as one example.
 The goal of the SVCB RR is to allow clients to resolve a single
 additional DNS RR in a way that:
 
-* Provides service endpoints authoritative for the service,
+* Provides alternative endpoints that are authoritative for the service,
```
MikeBishop, line 134: While this doesn't need to be addressed before merging, I'm not certain that the concept of "authoritative" is the same in all possible protocols, particularly ones where the scope of a "service" might vary.  We might want to consider whether that's a term we want to keep as well.

PR #205, opened 2020-07-01
Define a "mandatory" SvcParamKey

### LINKED ISSUE #166: Consider a way to indicate some keys as "mandatory"
@enygren has proposed identifying a key range as "mandatory".  A client encountering an unrecognized "mandatory" key would discard the entire RR.

The current draft already relies on mandatory understanding for the "port" and "alpn" keys.  Without a mandatory range, keys of this kind could not be defined in a future draft.

PR #164 is an example of something that might make sense as a followup RFC, but currently can't be because of the lack of such a range.

Previously, this seemed somewhat artificial to me, but with the new partitioned key ranges it doesn't seem so strange.  Perhaps we could assign this meaning to the new 1-255 "Standards Action" range.

EDIT: Alternatively, this could be its own SvcParamKey, e.g. `mandatory=port,alpn`.
EDIT2: Using 1-255 seems like a very small range, but if we ever get close to running out, we can simply define a `mandatory` SvcParamKey in that range, and use it to extend the behavior indefinitely.  So if we choose a range, it can be small.

bemasc:
Fixes #166

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -520,7 +520,7 @@ following procedure:
    {{client-failures}}).
 
 3. If one or more SVCB records of ServiceForm SvcRecordType are returned
-   for HOST, clients should select the highest-priority option with
+   for HOST, clients should select the highest-priority compatible record with
    acceptable parameters, and resolve AAAA and/or A records for its
    SvcDomainName if they are not already available.  These are the
    preferred SvcFieldValue and IP addresses.  If the connection fails, the
@@ -867,7 +867,30 @@ compliant recursive resolvers.  When SvcDomainName is ".", server operators
 SHOULD NOT include these hints, because they are unlikely to convey any
 performance benefit.
 
+## "mandatory" {#mandatory}
 
+In a ServiceForm RR, a SvcParamKey is considered "essential" if the RR will not
+function correctly for clients that ignore this SvcParamKey.  Each SVCB
+protocol mapping SHOULD specify a default set of essential keys.  The
+SvcParamKey "mandatory" is used to indicate the set of essential keys for this
+RR (overriding the default set).
+
+A ServiceForm RR is considered "compatible" with a client if the client
+implements support for all its essential keys.
+
+In presentation format, "mandatory" contains a list of zero or more valid
+SvcParamKeys, either by their registered name or in the unknown-key format
+({{svcfieldvalue}}).  Keys MAY appear in any order, but MUST NOT appear more
+than once.  Any listed keys MUST also appear in the SvcFieldValue.
+For example, the following is a valid SvcFieldValue:
+
+    port=1234 key65333=ex1 key65444=ex2 mandatory=key65444,port
+
+In wire format, the essential keys are represented by their numeric values in
+network byte order, concatenated in ascending order.
+
+This SvcParamKey is always implicitly essential, and MUST NOT appear in its
+own value list.
 
 # Using SVCB with HTTPS and HTTP {#https}
 
@@ -886,6 +909,7 @@ semantics apply equally to HTTPS RRs unless specified otherwise.
 
 All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPS RRs.  The default set of ALPN IDs is the single value "http/1.1".
+By default, the essential keys are "port", "alpn", and "no-default-alpn".
 
 The presence of an HTTPS RR for an origin also indicates
 that all HTTP resources are available over HTTPS, as
@@ -1002,12 +1026,9 @@ the origin, not the SvcDomainName.
 
 ## HTTP Strict Transport Security {#hsts}
 
-By publishing an HTTPS RR, the server
-operator indicates that all useful HTTP resources on that origin are
-reachable over HTTPS, similar to HTTP Strict Transport Security
-{{!HSTS=RFC6797}}.  When an HTTPS RR is present for an origin,
-all "http" scheme requests for that origin SHOULD logically be redirected
-to "https".
+By publishing a usable HTTPS RR, the server operator indicates that all
+useful HTTP resources on that origin are reachable over HTTPS, similar to
+HTTP Strict Transport Security {{!HSTS=RFC6797}}.
 
 Prior to making an "http" scheme request, the client SHOULD perform a lookup
 to determine if any HTTPS RRs exist for that origin.  To do so,
@@ -1021,8 +1042,9 @@ the client SHOULD construct a corresponding "https" URL as follows:
 
 This construction is equivalent to Section 8.3 of {{HSTS}}, point 5.
 
-If an HTTPS RR query for this "https" URL returns any HTTPS RRs
-(AliasForm or ServiceForm), the client SHOULD act as if it has received an
+If an HTTPS RR query for this "https" URL returns any AliasForm HTTPS RRs,
+or any compatible ServiceForm HTTPS RRs (see {{mandatory}}), the client
+SHOULD act as if it has received an
 HTTP "307 Temporary Redirect" redirect to this "https" URL.
 Because HTTPS RRs are received over an often insecure channel (DNS),
 clients MUST NOT place any more trust in this signal than if they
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -867,7 +867,30 @@ compliant recursive resolvers.  When SvcDomainName is ".", server operators
 SHOULD NOT include these hints, because they are unlikely to convey any
 performance benefit.
 
+## "mandatory" {#mandatory}
 
+In a ServiceForm RR, a SvcParamKey is considered "essential" if the RR will not
+function correctly for clients that ignore this SvcParamKey.  Each SVCB
+protocol mapping SHOULD specify a default set of essential keys.  The
+SvcParamKey "mandatory" is used to indicate the set of essential keys for this
+RR (overriding the default set).
```
enygren, line 876: Overriding or adding to the default set?  The later seems more usable?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -867,7 +867,30 @@ compliant recursive resolvers.  When SvcDomainName is ".", server operators
 SHOULD NOT include these hints, because they are unlikely to convey any
 performance benefit.
 
+## "mandatory" {#mandatory}
 
+In a ServiceForm RR, a SvcParamKey is considered "essential" if the RR will not
+function correctly for clients that ignore this SvcParamKey.  Each SVCB
+protocol mapping SHOULD specify a default set of essential keys.  The
+SvcParamKey "mandatory" is used to indicate the set of essential keys for this
```
enygren, line 875: Is there a reason to switch between the words "essential" and "mandatory"?  Using both is somewhat distracting.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1021,8 +1042,9 @@ the client SHOULD construct a corresponding "https" URL as follows:
 
 This construction is equivalent to Section 8.3 of {{HSTS}}, point 5.
 
-If an HTTPS RR query for this "https" URL returns any HTTPS RRs
-(AliasForm or ServiceForm), the client SHOULD act as if it has received an
+If an HTTPS RR query for this "https" URL returns any AliasForm HTTPS RRs,
+or any compatible ServiceForm HTTPS RRs (see {{mandatory}}), the client
```
enygren, line 1046: Should explicitly state that if the set of compatible ServiceForm HTTPS RRs is empty (such as due to all requiring a mandatory key not supported) that clients MUST NOT apply the HSTS-like behavior?  (It's implicit here, but may be worth calling out outright.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1252,8 +1274,9 @@ be populated with the registrations below:
 | 2           | no-default-alpn | No support for default protocol | (This document) |
 | 3           | port            | Port for alternative service    | (This document) |
 | 4           | ipv4hint        | IPv4 address hints              | (This document) |
-| 5           | echconfig      | Encrypted ClientHello info      | (This document) |
+| 5           | echconfig       | Encrypted ClientHello info      | (This document) |
 | 6           | ipv6hint        | IPv6 address hints              | (This document) |
+| 7           | mandatory       | Essential keys in this RR       | (This document) |
```
enygren, line 1279: Did we want to give it ID 0 ?  (I liked the idea of having it first.)  I guess we can't make it 1 because we don't want to move alpn?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -867,7 +867,30 @@ compliant recursive resolvers.  When SvcDomainName is ".", server operators
...
 
+In a ServiceForm RR, a SvcParamKey is considered "essential" if the RR will not
+function correctly for clients that ignore this SvcParamKey.  Each SVCB
+protocol mapping SHOULD specify a default set of essential keys.  The
+SvcParamKey "mandatory" is used to indicate the set of essential keys for this
+RR (overriding the default set).
+
+A ServiceForm RR is considered "compatible" with a client if the client
+implements support for all its essential keys.
+
```
enygren, line 880: I'd add a note here that if no ServiceForm RRs in the RRset are compatible then the client must treat the RRset as empty.

draft-ietf-dnsop-svcb-https.md
tfpauly: I’d prefer to stick with “mandatory” and “default mandatory set”.  Agreed that the mix is a bit confusing.

draft-ietf-dnsop-svcb-https.md
tfpauly: I’d rather us not move alpn if that’s okay :)

Any reason not to give this 0? It is a meta-key, so making it 0 seems elegant.

bemasc: We'd previously discussed leaving 0 unassigned so that APIs could use it to mean something like "null".  If we don't think that's actually useful then I'm OK with making this 0.

draft-ietf-dnsop-svcb-https.md
bemasc: I chose this wording because "mandatory", in my view, is extremely confusing: "mandatory" sounds like domain owners are required to publish or use the "mandatory" keys, and clients are expected to enforce that they are present (neither of which is true).  These keys aren't "mandatory"; they're "mandatory-to-understand-if-present".  "Non-ignorable".

If you can think of a clearer way to express this I would appreciate a suggestion!

draft-ietf-dnsop-svcb-https.md
bemasc: Overriding seems more powerful, since it allows making default-essential keys optional.  Given that this feature should be very rare, I figured usability was not too important.

Then again, maybe clients have to have an understanding of all the default-essential keys for a scheme anyway, so I guess we can just make it additive.

bemasc: Changed: I made it additional instead.

draft-ietf-dnsop-svcb-https.md
bemasc: I changed this to "mandatory"/"automatically mandatory".

draft-ietf-dnsop-svcb-https.md
bemasc: Done

draft-ietf-dnsop-svcb-https.md
bemasc: I changed "mandatory" to ID 0 and made key65535 (a.k.a. -1) explicitly reserved as "invalid key".

draft-ietf-dnsop-svcb-https.md
bemasc: I think that is already mandated by each section on client behavior, but I added a non-normative note to this effect.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -867,7 +867,32 @@ compliant recursive resolvers.  When SvcDomainName is ".", server operators
...
+A ServiceForm RR is considered "compatible" with a client if the client
+implements support for all its mandatory keys.  If the SVCB RRSet contains
+no compatible RRs, the client will generally act as if the RRSet is empty.
+
+In presentation format, "mandatory" contains a list of one or more valid
+SvcParamKeys, either by their registered name or in the unknown-key format
+({{svcfieldvalue}}).  Keys MAY appear in any order, but MUST NOT appear more
+than once.  Any listed keys MUST also appear in the SvcFieldValue.
+For example, the following is a valid SvcFieldValue:
+
```
enygren, line 888: Are "automatically mandatory" ones allowed to appear?  Is this "MUST NOT appear more than once" needed as a contraint for Presentation Format?
(May "automatically mandatory" entries appear in the wire format list?)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -867,7 +867,32 @@ compliant recursive resolvers.  When SvcDomainName is ".", server operators
...
+    echconfig=... key65333=ex1 key65444=ex2 mandatory=key65444,echconfig
+
+In wire format, the keys are represented by their numeric values in
+network byte order, concatenated in ascending order.
+
+This SvcParamKey is always automatically mandatory, and MUST NOT appear in its
+own value list.
```
enygren, line 895: Perhaps reframe as:

"This SvcParamKey is always automatically mandatory.  Automatically mandatory entries MUST NOT appear in the value list."

(Unless we want to allow them to appear as doing so is idempotent?)

draft-ietf-dnsop-svcb-https.md
bemasc: > Are "automatically mandatory" ones allowed to appear?

I updated the text to make this a "SHOULD NOT".

> Is this "MUST NOT appear more than once" needed as a contraint for Presentation Format?

No, but there's no use in having duplicates, and it's easy for the zone file parser to detect.

draft-ietf-dnsop-svcb-https.md
bemasc: I don't want to make this "MUST NOT", because zone file parsers can't detect it (they are not scheme-aware), and I would like to avoid introducing syntax errors that are caught by clients but not by the zone file parser.

PR #207, opened 2020-07-07
Recommend against DNAME-after-SVCB

### LINKED ISSUE #206: The target of a HTTPS/SVCB record SHOULD NOT be below a DNAME
While DNAMEs if present should be followed, the target of a HTTPS/SVCB records should not be below a DNAME.  All a DNAME does is introduce additional processing and records in the responses for no benefit.

bemasc:
Fixes #206


bemasc:
@marka63, does this work?

marka63:
yes, that should be fine

Issue #184, opened 2020-06-23
64k priorities!


martinthomson:
I realize that this might be going against an established trend, but the notion that 65535 different values for priority might be useful is a little fanciful.  HTTP are proposing just 8, DSCP had 64, and neither has any real expectation that there will be any value in having all of those options.  Why not 255?

bemasc:
Mainly for consistency with existing DNS types.  MX, SRV, NAPTR, and URI all have a similar "Priority" or "Preference" field, and they are all 16-bit unsigned.

I don't have a use case for more than 255 values, but I wouldn't want to break the pattern without a positive signal from the DNS experts.  I'm happy to solicit input from DNSOP if you think it's worth changing.

tfpauly:
I'd argue for leaving this as-is. Is it more priorities than we need? Yes, but it also fits the existing DNS RR type implementations better.

moonshiner:
It's true that Priority/Preference fields are 16-bit unsigned, but based on 20yr old decisions. 
I don't think I have seen anyone using a value > 255, and this is from years of looking at DNS zone data. 
If it is left as is, I'd mention it's to align with those DNS RR Types.

andrewtj:
It's odd to want to retain this for consistency with other types and also have no qualms about changing the meaning of the name ".".

enygren:
I propose leaving this as 64k for consistency and closing.

moonshiner:
I've made a note to mention this when we do a WGLC to see what the WG says

enygren:
Given the code point assignment for the RR types and this being a wire-format change, we should either change this ASAP or leave it as-is.

tfpauly:
Indeed. We can't really change this if we have the wire format stable. If we do anything, let's have editorial text explaining why it's a 16-bit field.

PR #208, opened 2020-07-09
Note on why priority is 16 bits

enygren:
Resolves issue #184


enygren:
(Note this uses an alternate CamelCase name as proposed elsewhere.)

PR #212, opened 2020-07-10
Alias to "." means repudiation

### LINKED ISSUE #203: Period is owner name for additional section processing is different to every other type.
For every other type the next name to be looked up is embedded in the rdata itself.  Additional "." often means "no such service".   For BIND adding "." is owner name meant touching 102 files as every other RR types method for processing additional data needed to changed as well as every place where additional data is added to the response.

I'm wondering if making "." special like this is worth it? 

I'm also wondering if we should be giving up the ability to signal "no such service" as is done in SRV, MX is worth it?

bemasc:
This change also adjusts the advice about the use of ".".

Fixes #203

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -490,7 +490,7 @@ same SvcFieldPriority value, clients SHOULD apply a random shuffle within a
 priority level to the records before using them, to ensure uniform
 load-balancing.
 
-## Special handling of "." {#dot}
+## Special handling of "." {#dot} in SvcDomainName
 
 If SvcDomainName has the value "." (represented in the wire format as a
 zero-length label), special rules apply.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -478,23 +478,6 @@ Unless specified otherwise by the
 protocol mapping, clients MUST ignore SvcFieldValue parameters that they do
 not recognize.
 
-### Special handling of "." for SvcDomainName in ServiceForm {#svcdomainnamedot}
-
-For ServiceForm SVCB RRs, if SvcDomainName has the value "." (represented in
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -715,10 +726,10 @@ the client MAY prefer those records, regardless of their priorities.
...
 
 To avoid a delay for clients using a nonconforming recursive resolver,
-domain owners SHOULD use a single SVCB record whose SvcDomainName is
-"." if possible.  This will ensure that the required
-address records are already present in the client's DNS cache as part of the
-responses to the address queries that were issued in parallel.
+domain owners SHOULD minimize the use of AliasForm records, and choose
+SvcDomainName to be a domain for which the client will have already issued
+address queries (see {{client-behavior}}).  The simplest such configuration
+would be a single SVCB record whose SvcDomainName is the origin hostname.
```
enygren, line 732:
```suggestion
domain owners SHOULD minimize the use of AliasForm records, and choose
SvcDomainName to be a domain for which the client will have already issued
address queries (see {{client-behavior}}), such as by using "." as the SvcDomainName for ServiceForm records.  The simplest SVCB configuration
would be a single ServiceForm SVCB record whose SvcDomainName is the origin hostname.  (Note that using the origin hostname as the SvcDomainName can be problematic when using {{!ECH}} 
({{svcparamkeys-echconfig}}) and switching between different
endpoints as clients may get IP addresses for one endpoint
and {{svcparamkeys-echconfig}} for a different endpoint.)
```

I'm not sure if the note should be moved elsewhere, but it's an important warning for explaining why using the simplest form may not be preferable.

bemasc: Rather than describe a simple configuration and then advise against it, I changed this section to have an explicit example that we would actually recommend, using a layer of indirection as intended.

PR #215, opened 2020-07-11
Add reference to HSTS on first use

### LINKED ISSUE #181: EOFU: HSTS
HSTS is mentioned in Section 1.1, but it is not expanded or a citation provided.

bemasc:
Fixes #181

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -155,8 +155,8 @@ Additional goals specific to HTTPS RRs and the HTTPS use-case include:
   implementing support for SRV records, as well as due to a limitation
   that a DNS name can not have both CNAME and NS RRs
   (as is the case for zone apex names)
-* Provide an HSTS-like indication signaling
-  for the duration of the DNS RR TTL that the HTTPS scheme should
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -154,8 +154,8 @@ Additional goals specific to HTTPS RRs and the HTTPS use-case include:
 * Support non-default TCP and UDP ports
 * Enable SRV-like benefits (e.g. apex delegation, as mentioned above) for HTTP(S),
   where SRV {{?SRV=RFC2782}} has not been widely adopted
-* Provide an HSTS-like indication signaling
-  for the duration of the DNS RR TTL that the HTTPS scheme should
+* Provide an HSTS-like indication {{!HSTS=RFC6797}} signaling
+  that the HTTPS scheme should
```
enygren:
```suggestion
* Provide a HSTS-like (but DNS TTL limited) indication {{!HSTS=RFC6797}} signaling
  that the HTTPS scheme should
```
Another wording might be to call it "temporary"?  Or is the real difference that HSTS lifetimes may be much longer than typical DNS TTLs?

PR #216, opened 2020-07-11
Simplify description of SRV/HTTP status

### LINKED ISSUE #180: SRV language is unnecesarily perjorative
> Address a set of long-standing issues due to HTTP(S) clients not implementing support for SRV records, as well as due to a limitation that a DNS name can not have both CNAME and NS RRs (as is the case for zone apex names)

This implies fault in clients for a refusal to implement, as opposed to SRV not providing sufficient advantages for HTTP(S) clients.

Rather than get into the why, it might be best to say which of the features of SRV are provided and why.  

I might also separate out the point about NS and CNAME not playing well together.

bemasc:
This also removes duplicate statements about CNAME at apex.

Fixes #180

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -151,10 +151,8 @@ Additional goals specific to HTTPS RRs and the HTTPS use-case include:
   alternative service endpoints
 * Obtain the {{!ECH}} keys associated with an alternative service endpoint
 * Support non-default TCP and UDP ports
-* Address a set of long-standing issues due to HTTP(S) clients not
-  implementing support for SRV records, as well as due to a limitation
-  that a DNS name can not have both CNAME and NS RRs
-  (as is the case for zone apex names)
+* Enable SRV-like benefits for HTTP(S), where SRV {{?SRV=RFC2782}} has not
+  been widely adopted
 * Provide an HSTS-like indication signaling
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -151,10 +151,8 @@ Additional goals specific to HTTPS RRs and the HTTPS use-case include:
   alternative service endpoints
 * Obtain the {{!ECH}} keys associated with an alternative service endpoint
 * Support non-default TCP and UDP ports
-* Address a set of long-standing issues due to HTTP(S) clients not
-  implementing support for SRV records, as well as due to a limitation
-  that a DNS name can not have both CNAME and NS RRs
-  (as is the case for zone apex names)
+* Enable SRV-like benefits for HTTP(S), where SRV {{?SRV=RFC2782}} has not
+  been widely adopted
```
enygren, line 155: I actually think this context (can't have a CNAME and NS RR at the zone apex) is helpful as this is one of the driving factors.  SRV is just one of the mechanisms that could do this.   Perhaps still include "..., such as allowing service delegation for a zone apex name where CNAMEs aren't allowed, ..." ?

bemasc: OK.  Rather than duplicate the statement about CNAME-at-apex, I added a pointer back to the CNAME-at-apex bullet point.

PR #220, opened 2020-07-13
Use non-normative language in intro

### LINKED ISSUE #182: Non-normative normative language
Seciton 1.2 is explicitly non-normative (whatever that means), but it uses some language that looks very much like normative requirements:

> Clients must either use responses included in the additional section returned by the recursive resolver or perform necessary SVCB, A, and AAAA record resolutions.  DNS authoritative servers may attach in-bailiwick SVCB, A, AAAA, and CNAME records in the Additional Section to responses for a SVCB query.

The use of lowercase here means non-normative, but the choice of words is unfortunate.  Why not use different words:

> Clients either use responses included in the additional section returned by the recursive resolver or perform necessary SVCB, A, and AAAA record resolutions.  DNS authoritative servers can attach in-bailiwick SVCB, A, AAAA, and CNAME records in the Additional Section to responses for a SVCB query.

enygren:
Fixes #182

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -183,10 +183,10 @@ The SVCB RR has two mandatory fields and one optional.  The fields are:
 
 Cooperating DNS recursive resolvers will perform subsequent record
 resolution (for SVCB, A, and AAAA records) and return them in the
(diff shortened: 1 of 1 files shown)


bemasc:
:+1:

PR #224, opened 2020-07-17
Recursives, handling loops, and clarifying AliasMode targets

### LINKED ISSUE #221: Relax "MUST NOT" on  target equalling owner name
From @marka63:

Section 2.4.1.  AliasMode paragraph 2.  Why have "MUST NOT" here?                         

   The primary purpose of AliasMode is to allow aliasing at the zone
   apex, where CNAME is not allowed.  In AliasMode, TargetName MUST be
   the name of a domain that has SVCB, AAAA, or A records.  It MUST NOT
   be equal to the owner name, as this would cause a loop.

Nameservers and clients need to detect loops regardless of the number of
records that form the loop.  All this does is force vendors to write additional
code to detect this one sub-case of a loop.  These are also the only RR types
that need to have their owner names available to parse them.

enygren:
Fixes #221 

Change the "TargetName of AliasMode SHOULD NOT equal owner name" (vs MUST NOT).

Clarify that recursives MUST have reasonable loop prevention.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -401,7 +401,7 @@ resolvers SHOULD pick one at random.
 The primary purpose of AliasMode is to allow aliasing at the zone
 apex, where CNAME is not allowed.  In AliasMode, the TargetName will
 be the name of a domain that resolves to SVCB
-(or other SVCB compatible record such as HTTPS),
+(or other SVCB-compatible record such as HTTPS),
 AAAA, and/or A records.  The TargetName SHOULD NOT be equal
 to the owner name, as this would result in a loop.
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -400,7 +400,8 @@ resolvers SHOULD pick one at random.
 
 The primary purpose of AliasMode is to allow aliasing at the zone
 apex, where CNAME is not allowed.  In AliasMode, the TargetName will
-be the name of a domain that resolves to SVCB (or HTTPS),
+be the name of a domain that resolves to SVCB
+(or other SVCB compatible record such as HTTPS),
 AAAA, and/or A records.  The TargetName SHOULD NOT be equal
 to the owner name, as this would result in a loop.
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -401,7 +401,7 @@ resolvers SHOULD pick one at random.
 The primary purpose of AliasMode is to allow aliasing
 at the zone apex, where CNAME is not allowed.
 In AliasMode, TargetName MUST be the name of a domain that has SVCB, AAAA,
-or A records.  It MUST NOT be equal to the owner name, as this would cause a
+or A records.  It SHOULD NOT be equal to the owner name, as this would cause a
 loop.
 
 For example, the operator of foo://example.com:8080 could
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -642,6 +642,10 @@ resolution procedure, as if processing a query for that RRSet.
 This includes following any aliases that the resolver would ordinarily
 follow (e.g. CNAME, DNAME {{!DNAME=RFC6672}}).
 
+Recursive resolvers MUST implement reasonable loop prevention,
```
bemasc, line 645: The AliasForm section already says "To avoid unbounded alias chains, clients and recursive resolvers MUST impose a limit on the total number of SVCB aliases they will follow".  I don't understand whether this requirement is equivalent or not.

bemasc:
Ping.  I'm happy to incorporate this if we can avoid duplicate normative language.

draft-ietf-dnsop-svcb-https.md
bemasc: To clarify, I think we need to answer whether "a limit on the total number of SVCB aliases" counts as "reasonable loop prevention".  If it does, then this line is redundant.  If it doesn't, then the intent here is to require a specific loop-detection behavior (before hitting the alias limit), which I think would need to be stated more explicitly.

enygren:
Based on discussion: 
* Only normative text in one place
* Make this not specific to loops

draft-ietf-dnsop-svcb-https.md
ericorth: Chrome intends an alias limit small enough that extra loop detection beyond that would just be silly.  Please don't make me implement silly loop detection.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -391,18 +391,18 @@ any ServiceMode records in the set.
...
 
-The primary purpose of AliasMode is to allow aliasing
-at the zone apex, where CNAME is not allowed.
-In AliasMode, TargetName MUST be the name of a domain that has SVCB, AAAA,
-or A records.  It MUST NOT be equal to the owner name, as this would cause a
-loop.
+The primary purpose of AliasMode is to allow aliasing at the zone
+apex, where CNAME is not allowed.  In AliasMode, the TargetName will
+be the name of a domain that resolves to SVCB (or HTTPS),
+AAAA, or A records.  The TargetName SHOULD NOT be equal
```
enygren, line 404:
```suggestion
AAAA, and/or A records.  The TargetName SHOULD NOT be equal
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -391,18 +391,18 @@ any ServiceMode records in the set.
...
 resolvers SHOULD pick one at random.
 
-The primary purpose of AliasMode is to allow aliasing
-at the zone apex, where CNAME is not allowed.
-In AliasMode, TargetName MUST be the name of a domain that has SVCB, AAAA,
-or A records.  It MUST NOT be equal to the owner name, as this would cause a
-loop.
+The primary purpose of AliasMode is to allow aliasing at the zone
+apex, where CNAME is not allowed.  In AliasMode, the TargetName will
+be the name of a domain that resolves to SVCB (or HTTPS),
```
enygren, line 403:
```suggestion
be the name of a domain that resolves to SVCB
(or other SVCB compatible record such as HTTPS),
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -391,18 +391,19 @@ any ServiceMode records in the set.
...
 
-The primary purpose of AliasMode is to allow aliasing
-at the zone apex, where CNAME is not allowed.
-In AliasMode, TargetName MUST be the name of a domain that has SVCB, AAAA,
-or A records.  It MUST NOT be equal to the owner name, as this would cause a
-loop.
+The primary purpose of AliasMode is to allow aliasing at the zone
+apex, where CNAME is not allowed.  In AliasMode, the TargetName will
+be the name of a domain that resolves to SVCB
+(or other SVCB compatible record such as HTTPS),
```
enygren, line 404:
```suggestion
(or other SVCB-compatible record such as HTTPS),
```

bemasc (approved):

PR #230, opened 2020-07-28
Proposal: Skip the default port prefix

bemasc:
This changes the definition of "Port Prefix Naming" to only include the
_port prefix if a non-default port is specified.  None of the examples
exercise this case, and the HTTPS RR overrides it, so this only matters
for future protocol mappings that choose to follow this advice.


tfpauly:
Looks good to me.

tfpauly (approved):

vcunat:
I see no issue with it, and reducing the depth is good for performance in case QNAME minimization is used.

enygren (approved):
(Unless we want to give an example or be explicit about when this can be omitted?)

PR #231, opened 2020-07-28
Explain "SVCB" at the top

### LINKED ISSUE #229: Add the expanded form of "SVCB" when it's first used
Typically, when introducing an acronym, it's useful to include its expanded form with it. That doesn't seem to happen in this document for "SVCB," nor does SVCB appear to be a common DNS term like RR. In fact, SVCB doesn't appear to be an acronym at all, but rather a shortened version of "Service binding." Is this true? If so, indicating that would be helpful.

bemasc:
Fixes #229

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -72,7 +72,7 @@ available there.  The authors (gratefully) accept pull requests.
 
 # Introduction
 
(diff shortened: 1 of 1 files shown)


bemasc:
@cbartle891 Is this better?

cbartle891:
Yes, thank you!

enygren (approved):

PR #238, opened 2020-08-11
Adjust additional section processing

bemasc:
This change makes the SHOULD-level requirement less specific, and adds a
very specific optional procedure that would optimize performance.


bemasc:
@wtoorop @marka63 This is my attempt to incorporate recent feedback regarding in-bailiwick CNAME chains.  Please review.

bemasc:
It seems like there was insufficient interest in tightening additional section processing on the mailing list, so I will close this PR.  This behavior is "opt-in", so it can always be defined in a later RFC if the working group decides it is desirable.

PR #241, opened 2020-09-06
Tjw editorial edits

moonshiner:
One minor editorial edit, and one spelling nit.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -244,7 +244,7 @@ The SVCB DNS resource record (RR) type (RR type 64)
 is used to locate alternative endpoints for a service.
 
 The algorithm for resolving SVCB records and associated
-address records is specified in {{client-behavior}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -244,7 +244,7 @@ The SVCB DNS resource record (RR) type (RR type 64)
 is used to locate alternative endpoints for a service.
 
 The algorithm for resolving SVCB records and associated
-address records is specified in {{client-behavior}}.
+address records are specified in {{client-behavior}}.
```
bemasc, line 247: Nope, singular :).  "The algorithm ... is specified".  We can rephrase if that sounds weird.

bemasc (changes requested):

draft-ietf-dnsop-svcb-https.md
moonshiner: Arrfg - you are right.  I did miss this

PR #243, opened 2020-09-08
Add an example of a full multi-CDN configuration

bemasc:


bemasc:
We've decided to describe a different architecture here.  @enygren will write up a proposal.

PR #259, opened 2020-10-01
Remove redundant requirement on Alias target RRs

### LINKED ISSUE #251: TargetName must match RRTYPE
§ 2.4.1 says:

```
The primary purpose of AliasMode is to allow aliasing at the zone apex, where
CNAME is not allowed. In AliasMode, TargetName MUST be the name of
a domain that has SVCB, AAAA, or A records. It MUST NOT be equal to
the owner name, as this would cause a loop.
```

From my reading, this should say that TargetName MUST be the name of a domain that has AAAA or A records, or records of the same RRTYPE as that of the source record. The intro says that

```
All behaviors described as applying to the SVCB RR also apply to the
HTTPS RR unless explicitly stated otherwise.
```

but that does not (and should not) explicitly encourage readers to alpha-substitute HTTPS for SVCB in every normative statement.

bemasc:
This requirement is obvious anyway, is hard to state precisely, and has
generated confusion from multiple readers.

Fixes #251

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -383,9 +383,8 @@ resolvers SHOULD pick one at random.
 
 The primary purpose of AliasMode is to allow aliasing
 at the zone apex, where CNAME is not allowed.
-In AliasMode, TargetName MUST be the name of a domain that has SVCB, AAAA,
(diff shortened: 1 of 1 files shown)


bemasc:
@enygren will take a look at improving the phrasing here.

PR #262, opened 2020-10-02
Proposal: Simplify SvcParamKey registration policy

bemasc:
This change removes the First Come First Served registration policy
and the Expert Review policy, and combines them into a single
Specification Required range, with guidelines for what constitutes a
minimal specification.  The goal of this change is to avoid zone file
incompatibilities that could arise from the registration of keys whose
presentation format is poorly documented.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1453,9 +1453,7 @@ Apart from the initial contents, the SvcParamKey name MUST NOT start with "key".
 Entries in this registry are subject to a Specification Required registration
 policy ({{!RFC8126}}, Section 4.6).  The provided specification MUST specify
 the SvcParamValue's presentation format and MAY detail its intended meaning
-and use.  Additionally, the registry SHALL allow registrants to have a Number
-temporarily marked as "Reserved" (without an associated Name)
-on a First Come First Served basis.
+and use.
 
 This arrangement supports the development of new parameters while ensuring that
 zone files can be made interoperable.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -793,7 +793,7 @@ For compatibility with clients that require default transports,
 zone operators SHOULD ensure that at least one RR in each RRSet supports the
 default transports.
 
-## "port"
+## "port" {#svcparamkeys-port}
 
 The "port" SvcParamKey defines the TCP or UDP port
 that should be used to reach this alternative endpoint.
@@ -1446,14 +1446,16 @@ A registration MUST include the following fields:
 * Number: wire format numeric identifier (range 0-65535)
 * Name: unique presentation name
 * Meaning: a short description
-* Reference: pointer to specification text
+* Format Reference: pointer to specification text
 
 Apart from the initial contents, the SvcParamKey name MUST NOT start with "key".
 
-Entries in this registry are subject to a Specification Required registration
-policy ({{!RFC8126}}, Section 4.6).  The provided specification MUST specify
-the SvcParamValue's presentation format and MAY detail its intended meaning
-and use.
+Entries in this registry are subject to a First Come First Served registration
+policy ({{!RFC8126}}, Section 4.6).  The Format Reference MUST specify
+how to convert the SvcParamValue's presentation format to wire format and MAY
+detail its intended meaning and use.  An entry MAY specify a Format Reference of
+the form "Same as (other key Name)" if it uses the same presentation and wire
+formats as an existing key.
 
 This arrangement supports the development of new parameters while ensuring that
 zone files can be made interoperable.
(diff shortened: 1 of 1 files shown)


bemasc:
@tfpauly FYI, this is relevant to a discussion we had.  Assuming this policy is considered permissible by IANA, I think it should provide a better balance of flexibility and interoperability.

bemasc:
TODO: replace "empty string" with "invalid".

bemasc:
@moonshiner (as chair) Do you think these new proposed IANA instructions will work?  For example, is the "temporary" instruction something that IANA can process?

tfpauly:
@bemasc can you provide more explanation for the problem for keys that don't have well-defined presentation formats?

I guess my concern is that we're creating a larger barrier towards experimentation, while not really solving the issue, since we can't expect all implementations to update presentations formats in line immediately when new specifications come up.

Shouldn't there be some default handling for presentation formats for unknown kinds?

tfpauly:
I think it'd be more acceptable to move the expert-review section to spec required if you want to ensure that there's a definition of format, or even just have instructions to the expert review to include that.

Alternatively, you can just have a row in the IANA registration for the presentation format. Wouldn't that be the same effect?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1443,25 +1443,22 @@ ACTION: create and include a reference to this registry.
...
-| 256-32767   | Expert Review           |
-| 32768-65280 | First Come First Served |
-| 65280-65534 | Private Use             |
-| 65535       | Standards Action        |
+Entries in this registry are subject to a Specification Required registration
+policy ({{!RFC8126}}, Section 4.6).  The provided specification MUST specify
+the SvcParamValue's presentation format and MAY detail its intended meaning
+and use.  Additionally, the registry SHALL allow registrants to have a Number
+temporarily marked as "Reserved" (without an associated Name)
+on a First Come First Served basis.
```
tfpauly, line 1458: This temporary first-come-first-served seems awkward. I think requiring the IANA registration to define the presentation format is much simpler and consistent with IANA policy.

tfpauly (changes requested):

bemasc:
> @bemasc can you provide more explanation for the problem for keys that don't have well-defined presentation formats?

Suppose we're transferring a zone file from auth server implementation A to implementation B, and it contains a SVCB RR with a particular key.

* If A doesn't know the key, it has to be represented in the unknown-key format.  That's fine.
* If A knows the key and B doesn't, the zone file can't be transferred (parse error).  That's unfortunate but unambiguous, and easily solved: B implements support for the key, or the owner rewrites the zone file to use the unknown-key format.
  * But if the key isn't publicly documented, how can B implement it?
* If A and B both think they know the key, but disagree on the exact presentation format, then arbitrary failures and silent corruption are possible.
  *  But if the key isn't publicly documented, how can we resolve the discrepancy?

Having named-but-undocumented SVCB keys floating around seems like a bad outcome for authoritative DNS interoperability, and therefore also for portability of domains between hosting services.

> Alternatively, you can just have a row in the IANA registration for the presentation format.

I think that's precisely what "Specification Required" means.  The conversion between wire format and presentation format won't literally fit inside an IANA registry entry, so we use a pointer to it.

Note that this is not "RFC required".  Any document that describes the format qualifies as a "Specification", so long as IANA believes that it is not going to disappear.  For example, individual I-Ds are considered sufficient.

draft-ietf-dnsop-svcb-https.md
bemasc: Does that mean you're OK with having to specify the presentation format before IANA allocates you a number?  Remember that FCFS doesn't mean you get to choose your number (except at IANA's discretion).

tfpauly: Yeah, I think it's quite reasonable to have to give information like that on the allocation.

tfpauly:
> The conversion between wire format and presentation format won't literally fit inside an IANA registry entry, so we use a pointer to it.

Is there any limit to what goes in the IANA registry? I've seen some pretty verbose registry tables.

draft-ietf-dnsop-svcb-https.md
MikeBishop: The apparent goal of this text is to be able to claim a code-point for experimental purposes before the presentation format is fixed.  RFC 7120 requires any early allocations to be from within the IETF, and for the implementation to be fully stable.  FCFS-to-reserve is an end-run around those requirements.  (And something that IANA specifically says [does not exist](https://www.iana.org/help/protocol-registration):  "There is an early allocation process available, but there is no reservation system.")  Is the intended audience here non-IETF and/or unstable usage?  If not, let's just stick with early allocation.

Note that there is still a Private Use range, which would work as a place to implement until the specification is stable enough for EA.

tfpauly: I don't think the FCFS range should be meant for code points that are intended to be later reserved and standardized. Rather, I'd expect them to be used for non-private use for specs that aren't fully standardized, but have value for deploying and support. Stable specs and RFCs would use the more official space once they have final formats, etc.

An example is how Oblivious DoH is using a value in the FCFS range with a specific definition that is used for interoperating implementations on the public Internet; it's well-defined, but it may not be the same as some later RFC that could eventually take precedence.

tfpauly:
I still think that we can specify what we need in the first-come first-served bucket. From RFC 8126:

```
   For the First Come First Served policy, assignments are made to
   anyone on a first come, first served basis.  There is no substantive
   review of the request, other than to ensure that it is well-formed
   and doesn't duplicate an existing assignment.  However, requests must
   include a minimal amount of clerical information, such as a point of
   contact (including an email address, and sometimes a postal address)
   and a brief description of how the value will be used.  Additional
   information specific to the type of value requested may also need to
   be provided, as defined by the namespace.
```

Let's just define that in order to be well-formed in the namespace, you need the zone file format.

bemasc:
I contacted an IANA representative and got some advice on how this works,as well as [this example](https://www.iana.org/assignments/websocket/websocket.xhtml#subprotocol-name) of a relatively recent registry that seems to be working well.  On that basis, I've changed this to FCFS and attempted to clarify the Reference requirement (now "Format Reference").

bemasc:
@tfpauly Please review.

tfpauly:
@bemasc looks good to me, thanks!

tfpauly (approved):

enygren (approved):

PR #263, opened 2020-10-02
Adjust fallback behavior

### LINKED ISSUE #256: Make a clean break with the past
To avoid the problem SSL/TLS had for years with ambiguity re: the CN in the presence of the SAN extension, you should clearly state somewhere that when SVCB is available for an authority endpoint, the legacy defaults (hostname on port 80/443) must never be used as a fallback or as part of a pool of options for the client. That, or explain clearly the justification for the current design choice.

bemasc:
Editorial changes only, to clarify the fallback sequence.

Fixes #256

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -503,15 +503,19 @@ procedure:
    subject to chain length limits and loop detection heuristics (see
    {{client-failures}}).
 
-4. If one or more ServiceMode records are returned
-   for $SVCB_QNAME, clients SHOULD select the highest-priority compatible record with
-   acceptable parameters.  This TargetName and SvcParams represent the
-   preferred endpoint.  If connection to this endpoint fails, the
-   client MAY try to connect using values from a lower-priority record.
-   If all attempts fail, clients SHOULD go to step 5 (except as noted in
-   {{ech-client-behavior}}.
+4. If one or more "compatible" ({{mandatory}}) ServiceMode records are returned
+   for $SVCB_QNAME, clients SHOULD select the highest-priority compatible record.
+   This record's TargetName and SvcParams represent the preferred endpoint.  If
+   connection to this endpoint fails, the client SHOULD try to connect using
+   values from the next-highest-priority compatible record, etc. If all attempts
+   fail, clients SHOULD go to step 5 (except as noted in {{ech-client-behavior}}).
 
-5. There are no usable ServiceMode records.  Clients SHALL connect to the
+5. At this point there are no usable ServiceMode records, because
+   - there were no SVCB records found for $SVCB_QNAME, OR
+   - all records found were incompatible with this client, OR
+   - all connection attempts using ServiceMode records failed.
+
+   Accordingly, clients SHALL connect to the
    endpoint consisting of $ADDR_QNAME, the authority endpoint's
    port number, and no SvcParams.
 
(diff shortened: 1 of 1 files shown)


bemasc:
@squarooticus FYI, this is the change based on your suggestion in #256.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -508,13 +508,16 @@ procedure:
    acceptable parameters.  This TargetName and SvcParams represent the
    preferred endpoint.  If connection to this endpoint fails, the
    client MAY try to connect using values from a lower-priority record.
-   If all attempts fail, clients SHOULD go to step 5.
+   If all attempts fail, clients SHOULD go to step 5 (except as noted in
+   {{ech-client-behavior}}.
 
-5. Try to use the endpoint consisting of $ADDR_QNAME, the authority endpoint's
+5. There are no usable ServiceMode records.  Clients SHALL connect to the
```
enygren, line 514:
```suggestion
5. At this point there are no usable ServiceMode records, either because $SVCB_QNAME has no SVCB records or because no SVCB records have acceptable parameters. In this case, clients SHALL connect to the
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -508,13 +508,16 @@ procedure:
    acceptable parameters.  This TargetName and SvcParams represent the
    preferred endpoint.  If connection to this endpoint fails, the
    client MAY try to connect using values from a lower-priority record.
-   If all attempts fail, clients SHOULD go to step 5.
+   If all attempts fail, clients SHOULD go to step 5 (except as noted in
```
enygren, line 511:
```suggestion
   If there are no usable ServiceMode records, clients MUST go to step 5.  If all attempts fail, clients SHOULD go to step 5 (except as noted in
```

enygren (changes requested):
On re-reading a few times, I think it would help to be crystal clear that if Step 5 is reached due to $SVCB_QNAME not having a SVCB record then the A/AAAA ($ADDR_QNAME) must be used instead.  In the section where we talk about DNSSEC it may make sense to say that authoritative answers (NODATA/NOERROR and NXDOMAIN) for SVCB records are acceptable and should result in $ADDR_QNAME fallback.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I incorporated a version of this change.

draft-ietf-dnsop-svcb-https.md
bemasc: "Usable" is ambiguous here.  I expanded this section and moved to our notion of "compatible" records (which may not have existed yet when we last updated this text).

draft-ietf-dnsop-svcb-https.md
```diff
@@ -895,7 +902,8 @@ mandatory", i.e. mandatory if they are present in an RR.  The SvcParamKey
 any automatically mandatory keys that are present.
 
 A ServiceMode RR is considered "compatible" with a client if the client
-implements support for all its mandatory keys.  If the SVCB RRSet contains
+recognizes all the mandatory keys, and their values indicate that successful
```
enygren, line 905: Should we rename this subsection from "mandatory" to "ServiceMode RR compatibility and mandatory keys" ?  That might be more descriptive?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -542,7 +549,7 @@ selectively dropping SVCB queries or responses, based on their size or
 other observable patterns.
 
 Similarly, if the client enforces DNSSEC validation on A/AAAA responses,
-it SHOULD NOT fall back to steps 5 or 6 if the SVCB
+it SHOULD NOT fall back to non-SVCB connection if a SVCB
```
enygren, line 552: Should we clear on what it should do?  (eg, fail the connection?  or use the failed-to-validate SVCB response?   I assume the former, but a little ambiguous here.)

enygren (changes requested):
Looking good other than the minor remaining questions.

draft-ietf-dnsop-svcb-https.md
bemasc: Done

draft-ietf-dnsop-svcb-https.md
bemasc: OK, done

PR #270, opened 2020-10-06
Move multi-value references into the keys

### LINKED ISSUE #247: Clarify presentation format for unrecognized keys
§ 2.1 says:

```
Unrecognized keys are represented in presentation format as
"keyNNNNN" where NNNNN is the numeric value of the key type without
leading zeros. SvcParams in this form are always treated as single-
valued, and the decoded "value" SHALL be used as its wire format
encoding.
```

Does this mean an unvalued key (empty value) would be presented as `keyNNNNN=`?

bemasc:
Fixes #247

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -266,21 +266,17 @@ In ABNF {{!RFC5234}},
     SvcParamValue = char-string
     value         = *OCTET
 
-The definition of each SvcParamKey indicates that its SvcParamValue is empty,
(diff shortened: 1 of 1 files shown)


enygren (changes requested):
We should define "value list" somewhere before its first use. (currently on line 749)  Otherwise it just gets used there (and subsequently) with no context.  Is "value list" a first class thing?

bemasc:
OK, I added a definition to try to make "value-list" a first-class term.

enygren (approved):

PR #281, opened 2020-12-14
Add a mapping summary table for the HTTPS mapping

bemasc:


draft-ietf-dnsop-svcb-https.md
```diff
@@ -1589,6 +1589,25 @@ two `list-value`s:
...
     part2,part3
 
+# HTTP Mapping Summary
+
+This table serves as a non-normative summary of the HTTP mapping for SVCB.
+Future protocol mappings may provide a similar summary table.
+
+|                                  |                                          |
+|----------------------------------+------------------------------------------|
+| **Mapped scheme**                | "https"                                  |
```
enygren, line 1599: How should we treat http (and ws/wss?) in this section?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1589,6 +1589,25 @@ two `list-value`s:
...
+| **RR type**                      | HTTPS                                    |
+| **Name prefixing**               | None for port 443, otherwise Port-Prefix |
```
enygren, line 1601: Include a reference for "Port-Prefix"?  (and/or give an example?)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1589,6 +1589,25 @@ two `list-value`s:
...
 
+# HTTP Mapping Summary
+
+This table serves as a non-normative summary of the HTTP mapping for SVCB.
+Future protocol mappings may provide a similar summary table.
+
+|                                  |                                          |
+|----------------------------------+------------------------------------------|
+| **Mapped scheme**                | "https"                                  |
+| **RR type**                      | HTTPS                                    |
```
enygren, line 1600: Include the RR code point in parens?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1589,6 +1589,25 @@ two `list-value`s:
...
+| **Name prefixing**               | None for port 443, otherwise Port-Prefix |
+| **Automatically Mandatory Keys** | `port`, `alpn`, `no-default-alpn`        |
+| **SvcParam defaults**            | `alpn`: \["http/1.1"\]                   |
+| **Special behaviors**            | HTTP to HTTPS upgrade                    |
```
enygren, line 1604: Include a section reference?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1589,6 +1589,25 @@ two `list-value`s:
...
+|----------------------------------+------------------------------------------|
+
+This table does not indicate any SvcParamKeys that servers are required to
+publish, or that clients are required to implement, because there are none
+in this mapping.
+
```
enygren, line 1610: * What happens if a client specifies no-default-alpn but doesn't include an alpn key?
* Aren't clients required to implement all of the automatically-mandatory keys?  Or just that if a server sends them but a client doesn't implement then the record is pruned?  Does this mean alpn is effectively required for clients to implement?

enygren (changes requested):

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I added a row for them.

draft-ietf-dnsop-svcb-https.md
bemasc: I changed this to an example/template.

draft-ietf-dnsop-svcb-https.md

draft-ietf-dnsop-svcb-https.md
bemasc: I added a reference to the overall section in the intro sentence.  I don't think we should try to squeeze references into this (non-normative) summary table.

draft-ietf-dnsop-svcb-https.md
ericorth: > > What happens if a client specifies no-default-alpn but doesn't include an alpn key?
> 
> Assuming s/client/origin, that record is null: its set of advertised ALPNs is empty, so no client will ever attempt to use it.
> 

For consistency with all the other rules for calling non-sensible stuff invalid, should this be made into a more formal "don't do that"? And should clients consider it malformed (at least those clients that can handle "alpn" if you remove it from the mandatory keys)?

Maybe: 'If "no-default-alpn" is indictated, "alpn" MUST be included.  If a clients that recognizes "alpn" SvcParamKeys receives an RR with "no-default-alpn" but no "alpn", the RR must be considered malformed.'

enygren: This state does violate the "alpn list is one or more entries" scenario, but it would be good to call it out explicitly somewhere, but likely elsewhere from this table.  (If no-default-alpn is specified without an alpn, the RR should be discarded seems like a reasonable behavior.)

enygren: Opened issue #283

enygren: And split out PR #284

draft-ietf-dnsop-svcb-https.md
```diff
@@ -936,7 +936,7 @@ the Internet ("IN") Class {{!RFC1035}}.
 
 All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPS RRs.  The default set of ALPN IDs is the single value "http/1.1".
-The "automatically mandatory" keys ({{mandatory}}) are "port", "alpn",
+The "automatically mandatory" keys ({{mandatory}}) are "port"
 and "no-default-alpn".
```
enygren, line 940: Do we want a clarifying note as to what it means for alpn to not be automatically mandatory given that no-default-alpn requires alpn to be present?

bemasc: I don't have any clarification to add on that point.  Suggestions welcome.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -936,7 +936,7 @@ the Internet ("IN") Class {{!RFC1035}}.
 
 All the SvcParamKeys defined in {{keys}} are permitted for use in
 HTTPS RRs.  The default set of ALPN IDs is the single value "http/1.1".
```
enygren, line 941:
```suggestion
HTTPS RRs.  The default set of ALPN IDs is the single value "http/1.1".
Clients not implementing the "alpn" key SHOULD behave as if the default
ALPN ID is present and also discard records containing no-default-alpn.
```

draft-ietf-dnsop-svcb-https.md
enygren: Added a suggestion above.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I added a variation on that sentence.

MikeBishop (approved):

bemasc:
@brian-peter-dickson Does this match your expectations for a summary table?

brian-peter-dickson:
Yes, that is great, thank you.

Issue #268, opened 2020-10-03
Please PLEASE simplify alpn and list representations and encoding/decoding


brian-peter-dickson:
The stuff in Appendix A.1 is both moot and confusing.

If I understand correctly (after having to read and cross-reference many parts of this document, the ALPN RFC(s), and the IANA table for ALPN), the following is the case:
NO ALPN IDs HAVE EMBEDDED COMMAS.
Thus, there is NO need for quoted commas in any ALPN.
And, no other values (single or multi) have (or allow) embedded commas.

Can we please dispense with all of the fancy things showing how to quote commas and how to interpret quoted commas?
If necessary, place a caveat into the IANA table that ALPN values MUST NOT contain commas in their ID.

This reduces ALPN to wire format of one or more <length>string (where length is a single octet), and the presentation format is simply a comma-separated list of strings, possibly enclosed in quotes, with no other quoting necessary.

The SvcParameters object itself is, in presentation format, a list separated with whitespace, so there is no ambiguity between the list within an ALPN, and the list of SvcParameters. 

Please?

It would be nice to have a much simpler document, and much easier task for implementing things.

bemasc:
You're correct: there are no defined ALPN IDs with embedded commas, and adopting that as an assumption would simplify the parsing considerably.  However, ALPN is defined as an "8-bit clean" field that can hold anything, so we would presumably have to "Update" RFC 7301 to introduce such a restriction.  I'm not convinced that we will find consensus in the TLS WG for that change.

I'm not really happy about where this has ended up, but I don't have an alternative that seems more likely to get consensus.

MikeBishop:
This is similar in spirit to RFC 7838; while all ALPN tokens currently defined can be expressed as `tchar` for use in HTTP headers, that RFC still has to define how to percent-encode bytes which are not `tchar`s because they remain legal bytes for ALPN.  While modifying the ALPN registry would simplify that, I doubt that's something that should be driven through dnsop.

If you bring a draft to the TLS WG adding such restrictions to ALPN and it is adopted and progressed, we will happily update this draft to rely on that simplification, though.

enygren:
Agreed with @MikeBishop and @bemasc .  I've never been a fan of this, but we're stuck with it unless we want to restrict to only support a subset of ALPNs.

bemasc:
[...]

PR #282, opened 2020-12-14
Disentangle comma-separated lists from char-string

bemasc:
This change allows unconditional char-string parsing to occur before
inspecting the SvcParamKey in order to determine whether it is
single-valued or multi-valued.  I've now encountered two implementations
of zone file syntax that have this structure, which makes the current
text unimplementable.

Addresses #268

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -266,7 +266,7 @@ In ABNF {{!RFC5234}},
     SvcParamValue = char-string
     value         = *OCTET
 
-Unless otherwise specified, the SvcParamValue is parsed using the
+The SvcParamValue is parsed using the
 character-string decoding algorithm ({{decoding}}), producing a `value`.
 The `value` is then validated and converted into wire-format in a manner
 specific to each key.
@@ -279,9 +279,9 @@ value of the key type without leading zeros.
 A SvcParam in this form SHALL be parsed as specified above, and
 the decoded `value` SHALL be used as its wire format encoding.
 
-For some SvcParamKeys, the SvcParamValue corresponds to a list or set of
-items.  Presentation formats for such keys SHOULD make use of the decoding
-algorithm in {{value-list}}, producing a "value-list".
+For some SvcParamKeys, the `value` corresponds to a list or set of
+items.  Presentation formats for such keys SHOULD use a comma-separated list
+({{value-list}}).
 
 SvcParams in presentation format MAY appear in any order, but keys MUST NOT be
 repeated.
@@ -736,10 +736,8 @@ ALPNs are identified by their registered "Identification Sequence"
 
     alpn-id = 1*255OCTET
 
-"alpn" is a multi-valued SvcParamKey.  To construct its value-list, apply the
-value-list decoding algorithm ({{value-list}}) to the SvcParamValue.
-Each decoded value in the "alpn" value-list
-SHALL be an `alpn-id`.  The value-list MUST NOT be empty.
+The presentation `value` SHALL be a comma-separated list ({{value-list}})
+of one or more `alpn-id`s.
 
 The wire format value for "alpn" consists of at least one
 `alpn-id` prefixed by its length as a single octet, and these length-value
@@ -851,9 +849,8 @@ addresses in response to the TargetName query. Failure to use A and/or
 AAAA response addresses could negatively impact load balancing or other
 geo-aware features and thereby degrade client performance.
 
-To construct the value-list, apply the value-list decoding algorithm
-({{value-list}}) to the SvcParamValue.
-Each decoded value in the value-list SHALL be an IP address of the appropriate
+The presentation `value` SHALL be a comma-separated list ({{value-list}})
+of one or more IP addresses of the appropriate
 family in standard textual format {{!RFC5952}}.  To enable simpler parsing,
 this SvcParamValue MUST NOT contain escape sequences.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -736,10 +736,8 @@ ALPNs are identified by their registered "Identification Sequence"
 
     alpn-id = 1*255OCTET
 
-"alpn" is a multi-valued SvcParamKey.  To construct its value-list, apply the
-value-list decoding algorithm ({{value-list}}) to the SvcParamValue.
-Each decoded value in the "alpn" value-list
-SHALL be an `alpn-id`.  The value-list MUST NOT be empty.
```
enygren, line 742: Do we still need the "value-list MUST NOT be empty" constraint?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1560,34 +1557,34 @@ In this document, this algorithm is referred to as "character-string decoding".
...
-In order to represent lists of values in zone files, this specification uses
-an extended version of character-string decoding that adds the use of ","
-as a delimiter after double-quote processing.  When "," is not escaped
-(by a preceding "\\" or as the escape sequence "\\044"), it separates
-values in the output, which is a list of 1*OCTET.  (For simplicity, empty
-values are not allowed.)  We refer to this modified procedure as "value-list
-decoding".
+In order to represent lists of items in zone files, this specification uses
+comma-separated lists.  When "," is not escaped (by a preceding "\\"), it
+separates items in the list.  (For simplicity, empty items are not allowed.)
```
enygren, line 1564: Should we specify how empty items are handled?  (ie, clients MUST discard records where the value-list is empty or where a 'value' is empty?)

enygren: (Although this may want to go elsewhere since it impacts wire format handling)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1560,34 +1557,34 @@ In this document, this algorithm is referred to as "character-string decoding".
...
 
-    value-list = char-string
-    list-value = 1*OCTET
+    item            = 1*OCTET
+    ; item-allowed is OCTET minus "," and "\".
```
enygren, line 1567: What about double-quote?  (""")

enygren (changes requested):

draft-ietf-dnsop-svcb-https.md
bemasc: It's still there: "one or more".

draft-ietf-dnsop-svcb-https.md
bemasc: Nothing here impacts clients; this is only relevant during zone-file parsing.  I think it's clear from the text here that `alpn=,,` is a syntax error that halts zone file parsing.

draft-ietf-dnsop-svcb-https.md
bemasc: Double-quote is not special-cased here.  It is treated just like any other character, e.g, `alpn="\"foo\",bar"` -> `"foo",bar` -> (`"foo"`, `bar`).

draft-ietf-dnsop-svcb-https.md
moonshiner: Not "halting" parsing, but just throwing out this record as invalid.  
(or the behavior to halt or not to halt is up to the implementation)

draft-ietf-dnsop-svcb-https.md
enygren: I should have re-read with more context.  Perhaps being clear below would help?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1560,34 +1557,34 @@ In this document, this algorithm is referred to as "character-string decoding".
...
+    item-allowed    = %x00-2B / %x2D-5B / %x5D-FF
+    escaped-item    = 1*(item-allowed / "\," / "\\")
+    comma-separated = [escaped-item *("," escaped-item)]
 
 For example, consider these `char-string` SvcParamValues:
```
enygren, line 1572:
```suggestion
Decoding of value-lists happens after character-string decoding.  
For example, consider these `char-string` SvcParamValues:
```

enygren (approved):

marka63:
> On 6 Jan 2021, at 05:06, Tim Wicinski <notifications@github.com> wrote:
> 
> 
> @moonshiner commented on this pull request.
> 
> In draft-ietf-dnsop-svcb-https.md:
> 
> >  
> -In order to represent lists of values in zone files, this specification uses
> -an extended version of character-string decoding that adds the use of ","
> -as a delimiter after double-quote processing.  When "," is not escaped
> -(by a preceding "\\" or as the escape sequence "\\044"), it separates
> -values in the output, which is a list of 1*OCTET.  (For simplicity, empty
> -values are not allowed.)  We refer to this modified procedure as "value-list
> -decoding".
> +In order to represent lists of items in zone files, this specification uses
> +comma-separated lists.  When "," is not escaped (by a preceding "\\"), it
> +separates items in the list.  (For simplicity, empty items are not allowed.)
> 
> Not "halting" parsing, but just throwing out this record as invalid.
> (or the behavior to halt or not to halt is up to the implementation)

If the DNS server loads the zone then it is broken.  STD13 REQUIRES that servers
not load a zone if there is any error detected when reading the master file.
You can continue processing and report multiple errors but the zone is not to
be loaded.

> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

-- 
Mark Andrews, ISC
1 Seymour St., Dundas Valley, NSW 2117, Australia
PHONE: +61 2 9871 4742              INTERNET: marka@isc.org

moonshiner:
Thanks Mark. I fail to remember that more often than not.

PR #284, opened 2021-01-05
When no-default-alpn is present, alpn must also be present

### LINKED ISSUE #283: Clarify behavior when no-default-alpn is specified but without alpn specified
See discussion:  https://github.com/MikeBishop/dns-alt-svc/pull/281/files#r551698492

enygren:
Fixes #283

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -750,7 +750,8 @@ pairs are concatenated to form the SvcParamValue.  These pairs MUST exactly
 fill the SvcParamValue; otherwise, the SvcParamValue is malformed.
 
 For "no-default-alpn", the presentation and wire format values MUST be
-empty.
+empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified
+(else the record should be discarded as invalid).
 
 Each scheme that uses this SvcParamKey defines a
 "default set" of supported ALPNs, which SHOULD NOT
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -750,8 +750,7 @@ pairs are concatenated to form the SvcParamValue.  These pairs MUST exactly
 fill the SvcParamValue; otherwise, the SvcParamValue is malformed.
 
 For "no-default-alpn", the presentation and wire format values MUST be
-empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified
-(else the record should be discarded as invalid).
+empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified.
 
 Each scheme that uses this SvcParamKey defines a
 "default set" of supported ALPNs, which SHOULD NOT
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -750,7 +750,9 @@ pairs are concatenated to form the SvcParamValue.  These pairs MUST exactly
 fill the SvcParamValue; otherwise, the SvcParamValue is malformed.
 
 For "no-default-alpn", the presentation and wire format values MUST be
-empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified.
+empty.  When "no-default-alpn" is specified in an RR, 
+"alpn" MUST also be specified in-order for the RR 
+to be "self-consistent" ({{service-mode}}).
 
 Each scheme that uses this SvcParamKey defines a
 "default set" of supported ALPNs, which SHOULD NOT


draft-ietf-dnsop-svcb-https.md
```diff
@@ -750,7 +750,8 @@ pairs are concatenated to form the SvcParamValue.  These pairs MUST exactly
 fill the SvcParamValue; otherwise, the SvcParamValue is malformed.
 
 For "no-default-alpn", the presentation and wire format values MUST be
-empty.
+empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified
+(else the record should be discarded as invalid).
```
bemasc, line 754: I would prefer to distinguish here between what we want the zone file parser to do vs. what we want the client to do.

I'm fine with telling the zone file parser that this is invalid, but "discarded" sounds silent.  "flagged as invalid" might be clearer.

For the client, I'd prefer to avoid imposing new requirements or repeating normative language.  If it's necessary, we could say something like "Compliant clients will ignore ...".

enygren, line 754:
```suggestion
(else compliant clients should ignore the record as invalid).
```

enygren: Better?

ericorth: There's already the rule that clients must reject the entire rrset if any RR is malformed. As written in this change, I think it is ambiguous whether or not the missing "alpn" counts as a malformed key, so I wouldn't be sure whether to ignore the RR (just for not having a supported alpn value) or the RRSet (as malformed).

Hence, I think we need the client expectation clarified here and avoid saying something like "invalid" unless we mean the malformed case. My preference would be to treat is as malformed to stop bad configurations, but I won't fight for that as long as it's unambiguous.

bemasc, line 754: After thinking about this some more, I've changed my mind.  We can just say "this is not allowed" and leave the details to the implementors.
```suggestion
empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified.
```

bemasc: "malformed" has a precise definition in Section 2.2.  This would not qualify as "malformed".

enygren: Making this more complicated is having no-default-alpn be mandatory but not alpn.  (See discussion in #281.)  This should cover and we should be crystal clear on what happens if no-default-alpn and alpn are present on some records but some other records have no no-default-alpn, and the client doesn't implement alpn support.  The net result should probably be that clients discard the no-default-alpn records but then use the other records.

ericorth: The potentially relevant part of that definition:
"the SvcParamValue for an SvcParamKey does not have the expected format"

Does an expected key/value not being present and declared "invalid" count as an unexpected format? Probably not, but I would still argue it's ambiguous enough to require clarification.

(And even if it's decided this doesn't meet the current definition, I think it would be better to modify the definition to cover this case, but that's getting into the area I said I wouldn't fight for.)

bemasc: @ericorth I think there's a logical distinction between "syntactic validity" (malformed/unparseable) and "semantic validity" (inter-key logical agreement).  If a record is malformed, something Really Bad happened (memory corruption?).  If there's a configuration error, that's different.  I don't have strong feelings about what clients ought to do in each case, but I can certainly imagine that we might recommend or allow different behaviors.

BTW, this isn't the only case of inter-key logical agreement.  We already have the case of "mandatory" indicating a key that is not actually present.

ericorth: Agreed that they're reasonably distinct cases and it would be very reasonable to treat the cases differently.

But I think my specific points still stand:
- My soft preference would be to add semantic validity (at least this case and maybe also the "mandatory" case) to the cases where the RRSet is rejected.  If somebody creates a stupidly bad config, it should be obvious so they know to fix it.  But it wouldn't be the end of the world either if we just make it discarded-RR or just let the stupid results of the stupid config happen.
- Logically distinct or not, there's still room for confusion.  Maybe we just need a sentence in 2.2 clarifying that it only covers syntactic validity of the individual values themselves?

ericorth: And also, absent a clearly specified rule for how the client should behave given a violation of a semantic rule like this one, we should avoid the word "invalid".  As a client implementor, I want to know the proper way to enforce validity.  If the answer is that it's not enforced in any way and just continue to handle it logically (eg by treating a record as one without a matching ALPN and thus discarded RR), then that is not "invalid".

bemasc:
@enygren WDYT about harmonizing this language with #285?

draft-ietf-dnsop-svcb-https.md
enygren: With #285 this does cover things.  Although does that mean that "alpn" should be default mandatory but with a clarification that this doesn't mean clients need to understand all possible values?  As if "no-default-alpn" is mandatory and requires "alpn" for consistency then it would seem like "alpn" needs to be mandatory transitively?

bemasc: No, I don't think this means "alpn" needs to be "automatically mandatory".  Fundamentally, "automatically mandatory" mean "subtractive", i.e. "this key can subtract functionality from the endpoint".  "alpn" is purely additive, so it is always safe to ignore.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -750,7 +750,7 @@ pairs are concatenated to form the SvcParamValue.  These pairs MUST exactly
 fill the SvcParamValue; otherwise, the SvcParamValue is malformed.
 
 For "no-default-alpn", the presentation and wire format values MUST be
-empty.
+empty.  When "no-default-alpn" is specified, "alpn" MUST also be specified.
```
enygren, line 753:
```suggestion
empty.  When "no-default-alpn" is specified in an RR, 
"alpn" MUST also be specified in-order for the RR 
to be "self-consistent" ({{service-mode}}).
```

enygren: (to align with #285)

bemasc: :+1:

bemasc (approved):

draft-ietf-dnsop-svcb-https.md
ericorth: Nit: If I'm remembering the language correctly, being self-consistent is only a SHOULD, so any "in order for the RR to be 'self-consistent'" rules should probably also be a SHOULD, rather than a MUST.  You can create records that violate it, but if you do, you have to expect clients will ignore the RR or RRSet.

enygren: When not "self-consistent", dropping the RR is a MUST for clients, rejecting the entire RRset is a MAY.

bemasc: I read this as "The record is not self-consistent if it has no-default-alpn but not alpn.".  I think it's fine.  It is perhaps slightly unusual use of normative language ... we could make it a lower-case "must" :).

ericorth: But zone implementors enforcing that records are self-consistent was only a SHOULD, right? Making this a MUST would add an additional strong requirement specific to this rule.

ericorth: To further clarify my point, ""alpn" MUST also be specified in-order for the RR to be "self-consistent" is a rule for the zone implementors to follow.  They're the ones able to make that happen or not.  That rule is then inconsistent with "Zone-file implementations SHOULD enforce self-consistency."

I think either ""alpn" SHOULD also be specified in-order for the RR to be "self-consistent" or ""alpn" must also be specified in-order for the RR to be "self-consistent" would be sufficient to break the inconsistency, but I think the former is slightly clearer.

bemasc: Let's argue in #290.

PR #285, opened 2021-01-11
Clarify self-consistency rules

bemasc:
Requested in https://github.com/MikeBishop/dns-alt-svc/pull/284/files#r552202056

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -459,6 +459,12 @@ Unless specified otherwise by the
 protocol mapping, clients MUST ignore any SvcParam that they do
 not recognize.
 
+Some SvcParams impose requirements on other SvcParams in the RR.  A
+ServiceMode RR is called "self-consistent" if its SvcParams all comply with
+each others' requirements.  Zone-file implementations SHOULD enforce
+self-consistency.  Clients MUST reject any RR that they identify is not
+self-consistent, and MAY reject the entire RRSet.
+
 ## Special handling of "." in TargetName {#dot}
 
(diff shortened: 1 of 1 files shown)


bemasc:
@ericorth Does this address your concern?

ericorth:
> @ericorth Does this address your concern?

Yes.  It does so nicely.  Thanks.

Further suggestion: We should call out "would make the record not self-consistent" or something like that for the alpn/no-default-alpn case we were calling out as "invalid" before (and any other cases we know of).  Good to keep the language in the draft consistent and precise ("malformed" vs "invalid" vs "not self-consistent").

bemasc:
> Further suggestion: We should call out "would make the record not self-consistent" or something like that for the alpn/no-default-alpn case we were calling out as "invalid" before (and any other cases we know of). Good to keep the language in the draft consistent and precise ("malformed" vs "invalid" vs "not self-consistent").

Agreed; I was just trying to avoid a collision with #284.

enygren:
One thing that does get messy here is how to handle new types of keys.  A zone file parser clearly can't handle keyNNNN form SvcParams, but for clients it seems like there could be cases where a new non-mandatory SvcParm shows up that is declared in its spec to conflicts with some other SvcParam.  There is no way for clients which are aware of these to reject the RR.  We also want to avoid cases where clients confused by some new SvcParams throw out the entire RRset and cause compatibility issues.  I'm not sure what we might want to add to help here?

bemasc:
> There is no way for clients which are aware of these to reject the RR.

Do you mean "not aware"?

> I'm not sure what we might want to add to help here?

I adjusted the text.  Is it clearer now?

enygren (approved):

Issue #287, opened 2021-01-14
Can you use SVCB without A/AAAA?


martinthomson:
It does not look like the draft considers the possibility that a protocol might depend on SVCB exclusively.

I think that it should be possible to deploy a protocol that only uses SVCB and not A/AAAA to find services.

The changes are fairly simple, you go from this text in Section 3:

>   2.  In parallel, issue AAAA/A queries for $ADDR_QNAME and a SVCB query for $SVCB_QNAME.  The answers for these may or may not include CNAME pointers before reaching one or more of these records.

To something like:

>   2.  If the service allows the use of address records, issue AAAA/A queries for $ADDR_QNAME and a SVCB query for $SVCB_QNAME in parallel.  Otherwise, issue a SVCB query for $SVCB_QNAME.  The answers for these may or may not include CNAME pointers before reaching one or more of these records.

... and

> Accordingly, clients SHALL connect to the endpoint consisting of $ADDR_QNAME, the authority endpoint's port number, and no SvcParams.

... becomes:

> Accordingly, if the service allows the use of address records, clients connect to the endpoint consisting of $ADDR_QNAME, the authority endpoint's port number, and no SvcParams.

Edit to note: The reason you might want to do this is that you have a service name (like "example.com"), but you don't want the A/AAAA records you use for another protocol (like HTTPS) being used for your fancy new protocol.

bemasc:
I've proposed a paragraph on this in #288; please review.

I don't think the change proposed here is quite right.  Even if some new protocol relies on SVCB, I think it still makes sense for clients to issue address queries in parallel to the most likely TargetName, for latency hiding.  Issuing those queries doesn't mean they'll be used, even if they resolve, because the TargetName has precedence (if the prediction was wrong).

Ultimately, almost everything about SVCB can be modified by the protocol mapping specification for any protocol, so we don't need to enumerate every plausible behavior here, but I do think it's useful to mention the "SVCB-reliant" case.

enygren:
To clarify, is the scenario that:

1) the A/AAAA on the same name as the record would never get used (unless the SvcDomainName was "." or otherwise pointed to the record)
or
2) The scenario that this is a protocol for which no address records are involved at all (eg, saying "use non-IP ansible protocol via a subspace network for a service located by some identifier specified as a SvcParam", or "connect via a non-IP unix-domain-socket via some namespace identified in a SvcParam")

sftcd:
[...]

PR #288, opened 2021-01-14
Discuss hypothetical SVCB-reliant protocols

bemasc:
Addresses #287

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -535,6 +535,10 @@ such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
 Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
+Although SVCB can only provide alternative endpoints for any pre-existing
+protocol, new protocols could be defined that require a ServiceMode record.
+Such "SVCB-reliant" protocols SHALL terminate upon reaching step 5.
+
 ## Handling resolution failures {#client-failures}
 
 If a SVCB query results in a SERVFAIL error, transport error, or timeout,
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -536,7 +536,7 @@ Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
 Although SVCB can only provide alternative endpoints for any pre-existing
-protocol, new protocols could be defined that require a ServiceMode record.
+protocol, new protocols could be defined that require a ServiceMode record with no fallback to AAAA/A address records.
 Such "SVCB-reliant" protocols SHALL terminate upon reaching step 5.
 
 ## Handling resolution failures {#client-failures}


draft-ietf-dnsop-svcb-https.md
```diff
@@ -535,6 +535,10 @@ such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
 Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
+Although SVCB can only provide alternative endpoints for any pre-existing
+protocol, new protocols could be defined that require a ServiceMode record.
```
enygren, line 539:
```suggestion
protocol, new protocols could be defined that require a ServiceMode record with no fallback to AAAA/A address records.
```

enygren:
Waiting on clarification of https://github.com/MikeBishop/dns-alt-svc/issues/287#issuecomment-760346439  if (1) is what is meant, this seems fine.  If (2) is meant, more is needed.

martinthomson (changes requested):
This is not responsive to the changes.

To Erik's question, this is because I want to be able to deploy a protocol that uses SVCB for discovery on a name that has A/AAAA.  I don't want the servers that the A/AAAA records point at getting requests for that protocol.  (Honestly, I don't know which of the options Erik presents that is, so I'm restating it instead.)

bemasc:
@martinthomson I don't understand.  This change* has precisely that effect.  Connections will only ever go to the ServiceMode TargetName.  As stated in the new text, if no ServiceMode records are present, the connection attempt terminates (i.e. fails) rather than fall back to bare-AAAA connection.

*I don't think this is actually a "change"; it's more of a hint to future standards authors.

martinthomson:
Given how definitively the process in Section 3 is about making A/AAAA queries, I don't agree that this is sufficient.

bemasc:
Yes, I am not proposing to remove those queries.  Those queries are a useful latency optimization even when the connection is required to use ServiceMode records, so long as the TargetName is reasonably likely to coincide with the origin name.  Removing step 5 means that those queries will never be used for a "legacy" connection; they will only be used if they happen to coincide with (and pre-empt) a followup query for a ServiceMode TargetName.

I recognize that you are considering a particular deployment in which the TargetName does not coincide with the origin name, in which case those queries will be wasted, but will otherwise have no effect.

martinthomson:
I'm contemplating a model where those queries could produce answers that should not be used.  That's different from waste.  That's a problem.  Maybe I just don't understand why making them conditional on them being useful is a problem.

bemasc:
> I'm contemplating a model where those queries could produce answers that should not be used.

Yes.  If they should not be used, they will not be used.

> Maybe I just don't understand why making them conditional on them being useful is a problem.

I'm focusing on a model where the client doesn't know whether they are useful until it learns the server configuration.  Suppose the client is connecting to `foo://origin.example`, a future SVCB-reliant URI scheme.  Under your proposed behavior, the client would issue only a SVCB query for `_foo.origin.example.`.  Suppose they receive a response of the form 

```dns-zone
_foo.origin.example. IN SVCB 900 1 origin.example. alpn=bar`
```

If the recursive resolver is not performing Additional Section processing, the client now must issue followup A/AAAA queries for `origin.example.`, and wait for a response before proceeding.  If they had preissued those queries, they would not have to wait.

Preissuing the A/AAAA queries is inexpensive, so if there's a reasonable likelihood of them being useful, I think they're worth sending.  That likelihood decreases if `foo://` server deployments don't typically set TargetName to the origin ... but `foo://` deployments don't exist yet.  Encouraging such a convention enables this latency-hiding to work (at least sometimes), improving average performance.

We could add text describing _both_ a future-protocol that is likely to use origin-located address records, and a future-protocol that is not.  I think this kind of speculation, enumerating behaviors of multiple nonexistent protocols, is more likely to obscure than elucidate.

When it comes to future protocols, this whole section is just a hint to our future selves.  If you think this PR doesn't provide a suitable hint, perhaps you'd prefer the solution in #289.

enygren:
I like the approach in #289 as it defers this to the future without trying to pre-suppose what future protocols will want.  Since this is client behavior (not resolver or authority behavior) there is no reason hypothetical futures need to be defined here.

enygren:
Any objections to abandoning this in-favor of #289?  (It is unclear what value we get from adding this in-addition to #289)

bemasc:
> Any objections to abandoning this in-favor of #289? (It is unclear what value we get from adding this in-addition to #289)

That's fine with me, but I'd like to get some input from @martinthomson before proceeding with either change.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -535,6 +535,10 @@ such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
 Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
+Although SVCB can only provide alternative endpoints for any pre-existing
+protocol, new protocols could be defined that require a ServiceMode record with no fallback to AAAA/A address records.
+Such "SVCB-reliant" protocols SHALL terminate upon reaching step 5.
```
enygren, line 540:
```suggestion
Such "SVCB-reliant" protocols SHALL:
* ... start off step 1 with ADDR_QNAME being undefined
* ... not issue A/AAAA queries when ADDR_QNAME is undefined
* ... terminate upon reaching step 5
```

enygren (commented):
Would this change or one similar address @martinthomson 's concern?

draft-ietf-dnsop-svcb-https.md
martinthomson: Yes, I think that would do.  Though, as I noted in mail, it would be better if the current optimization were framed as such and the conditions under which the optimization apply more clearly articulated.

bemasc: I've drawn up #291 as a _third_ alternative solution here, which implements that reframing.  I'm not sure it's more comprehensible than this text, but it is probably more precise.  Please review.

bemasc:
Superseded by #291

PR #290, opened 2021-01-15
Use lower-case MUST for self-consistency rule

bemasc:
This is probably more correct use of normative language.


ericorth:
LGTM.  Makes it clear enough that the statement is just "you need to do this or it won't be self-consistent".  Whereas "you MUST do this..." made it sound like a stand-alone rule which was not correct.

enygren (approved):

PR #291, opened 2021-02-01
Restructure client requirements

bemasc:
This change significantly refactors the client requirements, separating
recommended performance optimizations from minimal compliance
requirements.

Addresses #287

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -518,7 +518,7 @@ comply with this specification or have any awareness of SVCB.
 
 Once SVCB resolution is completed, the client proceeds with connection
 establishment.  Clients SHOULD try higher-priority alternatives first, with
-fallback to lower-priority alternatives.  Clients MUST issue AAAA and/or A
+fallback to lower-priority alternatives.  Clients issue AAAA and/or A
 queries for the selected TargetName, and MAY choose between them using an
 approach such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
 
@@ -702,7 +702,7 @@ though the other record in the RRSet has higher priority.
 
 If none of the SVCB records are consistent
 with any active or in-progress connection,
-clients must proceed with connection establishment as described in
+clients proceed with connection establishment as described in
 {{client-behavior}}.
 
 ## Generating and using incomplete responses {#incomplete-response}
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -495,59 +495,53 @@ is the effective TargetName:
 
 # Client behavior {#client-behavior}
 
-A SVCB-aware client selects an endpoint for a service using the following
-procedure:
+"SVCB resolution" is the process of enumerating the priority-ordered endpoints
+for a service, as performed by the client.  It MUST be implemented as follows:
 
-1. Let $ADDR_QNAME be the service name.  Let $SVCB_QNAME be the service name
-   plus appropriate prefixes for the scheme (see {{svcb-names}}).
+1. Let $QNAME be the service name plus appropriate prefixes for the
+   scheme (see {{svcb-names}}).
 
-2. In parallel, issue AAAA/A queries for $ADDR_QNAME
-   and a SVCB query for $SVCB_QNAME.
-   The answers for these may or may not include CNAME pointers
-   before reaching one or more of these records.
+2. Issue a SVCB query for $QNAME.
 
-3. If an AliasMode SVCB record is returned for $SVCB_QNAME,
-   clients MUST set $ADDR_QNAME and $SVCB_QNAME to its TargetName (without
+3. If an AliasMode SVCB record is returned, set $QNAME to its TargetName (without
    additional prefixes) and loop back to step 2,
    subject to chain length limits and loop detection heuristics (see
    {{client-failures}}).
 
-4. If one or more "compatible" ({{mandatory}}) ServiceMode records are returned
-   for $SVCB_QNAME, clients SHOULD select the highest-priority compatible record.
-   This record's TargetName and SvcParams represent the preferred endpoint.  If
-   connection to this endpoint fails, the client SHOULD try to connect using
-   values from the next-highest-priority compatible record, etc. If all attempts
-   fail, clients SHOULD go to step 5 (except as noted in {{ech-client-behavior}}).
+4. If one or more "compatible" ({{mandatory}}) ServiceMode records are returned,
+   these represent the alternative endpoints.
 
-5. At this point there are no usable ServiceMode records, because
-   - there were no SVCB records found for $SVCB_QNAME, OR
-   - all records found were incompatible with this client, OR
-   - all connection attempts using ServiceMode records failed.
-
-   Accordingly, clients SHALL connect to the
-   endpoint consisting of $ADDR_QNAME, the authority endpoint's
-   port number, and no SvcParams.
-
-If all of the above connection attempts fail, clients MAY connect to the
-authority endpoint (except as noted in {{client-failures}} and
-{{ech-client-behavior}}).
+5. Otherwise, SVCB resolution has failed.
 
 This procedure does not rely on any recursive or authoritative DNS server to
 comply with this specification or have any awareness of SVCB.
 
-When selecting between AAAA and A records to use, clients may use an approach
-such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
+Once SVCB resolution is completed, the client proceeds with connection
+establishment.  Clients SHOULD try higher-priority alternatives first, with
+fallback to lower-priority alternatives.  Clients MUST issue AAAA and/or A
+queries for the selected TargetName, and MAY choose between them using an
+approach such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
+
+A client is called "SVCB-reliant" if it connects only to the indicated
+ServiceMode endpoints, and "SVCB-optional" otherwise.  Clients for pre-existing
+protocols SHALL implement SVCB-optional behavior (except as noted in
+{{client-failures}} and {{ech-client-behavior}}).
+
+SVCB-optional clients SHOULD issue in parallel any other DNS queries that might
+be needed for connection establishment.  SVCB-optional clients SHALL append an
+alternative endpoint consisting of the final value of $QNAME, the authority
+endpoint's port number, and no SvcParams, and MAY append the authority endpoint.
 
 Some important optimizations are discussed in {{optimizations}}
 to avoid additional latency in comparison to ordinary AAAA/A lookups.
 
 ## Handling resolution failures {#client-failures}
 
-If a SVCB query results in a SERVFAIL error, transport error, or timeout,
+If SVCB resolution fails due to a SERVFAIL error, transport error, or timeout,
 and DNS exchanges between the client and the recursive resolver are
 cryptographically protected (e.g. using TLS {{!DoT=RFC7858}} or HTTPS
-{{!DoH=RFC8484}}), the client SHOULD NOT fall back to $ADDR_QNAME (step 5
-above) or the authority endpoint.  Otherwise, an active attacker
+{{!DoH=RFC8484}}), a SVCB-optional client SHOULD terminate the connection like
+a SVCB-reliant client would.  Otherwise, an active attacker
 could mount a downgrade attack by denying the user access to the SvcParams.
 
 A SERVFAIL error can occur if the domain is DNSSEC-signed, the recursive
@@ -670,6 +664,15 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+To improve performance when using a non-conforming recursive resolver,
+clients SHOULD issue speculative A and AAAA queries based on a predicted value
+of TargetName.  In the absence of better information, clients SHOULD predict
+that TargetName will be the service name for the initial query and "." for
+subsequent queries (see {zone-performance}).
+
+After a ServiceMode RRSet is received, clients MAY try more than one option
+in parallel, and MAY prefetch A and AAAA records for multiple TargetNames.
+
 ## Optimistic pre-connection and connection reuse
 
 If an address response arrives before the corresponding SVCB response, the
@@ -681,7 +684,7 @@ implementing this optimization SHOULD wait for 50 milliseconds before
 starting optimistic pre-connection, as per the guidance in
 {{HappyEyeballsV2}}.
 
-An SVCB record is consistent with a connection
+A SVCB record is consistent with a connection
 if the client would attempt an equivalent connection when making use of
 that record. If a SVCB record is consistent with an active or in-progress
 connection C, the client MAY prefer that record and use C as its connection.
@@ -699,7 +702,8 @@ though the other record in the RRSet has higher priority.
 
 If none of the SVCB records are consistent
 with any active or in-progress connection,
-clients must proceed as described in Step 3 of the procedure in {{client-behavior}}.
+clients must proceed with connection establishment as described in
+{{client-behavior}}.
 
 ## Generating and using incomplete responses {#incomplete-response}
 
@@ -1048,8 +1052,9 @@ if a usable Alt-Svc value is available in the local cache.
 If Alt-Svc connection fails, these clients SHOULD fall back to the HTTPS RR
 client connection procedure ({{client-behavior}}).
 
-For clients that implement support for ECH, the interaction between
-HTTPS RRs and Alt-Svc is described in {{ech-client-behavior}}.
+Clients that implement support for ECH MUST perform the HTTPS RR query first,
+and MUST only make use of Alt-Svc when operating in SVCB-optional mode (see
+{{ech-client-behavior}}).
 
 This specification does not alter the DNS queries performed when connecting
 to an Alt-Svc hostname (typically A and/or AAAA only).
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,53 @@ is the effective TargetName:
...
+5. Otherwise, SVCB resolution has failed.
 
 This procedure does not rely on any recursive or authoritative DNS server to
 comply with this specification or have any awareness of SVCB.
 
-When selecting between AAAA and A records to use, clients may use an approach
-such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
+Once SVCB resolution is completed, the client proceeds with connection
+establishment.  Clients SHOULD try higher-priority alternatives first, with
+fallback to lower-priority alternatives.  Clients MUST issue AAAA and/or A
```
martinthomson, line 521:
```suggestion
fallback to lower-priority alternatives.  Clients issue AAAA and/or A
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,53 @@ is the effective TargetName:
...
+queries for the selected TargetName, and MAY choose between them using an
+approach such as Happy Eyeballs {{!HappyEyeballsV2=RFC8305}}.
+
+A client is called "SVCB-reliant" if it connects only to the indicated
+ServiceMode endpoints, and "SVCB-optional" otherwise.  Clients for pre-existing
+protocols SHALL implement SVCB-optional behavior (except as noted in
+{{client-failures}} and {{ech-client-behavior}}).
```
martinthomson, line 528: This definition of SVCB-reliant doesn't match with my understanding.  I think that what you want to do is note that there are two reasons to make an A/AAAA query in parallel.  The primary one being that the client might be willing to fall back to A/AAAA queries if there are no SVCB records, but you might also want to allow for the possibility that a TargetName form SVCB record references the same name and so there might be an advantage in making the A/AAAA queries in parallel to allow for that outcome.

(Unfortunately, if you *need* SVCB, you need the TargetName form, and you don't get to predict when that will appear, so you basically have to make the parallel query always once you decide that it is worth the parallel.)

You mention the latter below, which is good, but I think that it applies here.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,53 @@ is the effective TargetName:
...
 ## Handling resolution failures {#client-failures}
 
-If a SVCB query results in a SERVFAIL error, transport error, or timeout,
+If SVCB resolution fails due to a SERVFAIL error, transport error, or timeout,
 and DNS exchanges between the client and the recursive resolver are
 cryptographically protected (e.g. using TLS {{!DoT=RFC7858}} or HTTPS
-{{!DoH=RFC8484}}), the client SHOULD NOT fall back to $ADDR_QNAME (step 5
-above) or the authority endpoint.  Otherwise, an active attacker
+{{!DoH=RFC8484}}), a SVCB-optional client SHOULD terminate the connection like
+a SVCB-reliant client would.  Otherwise, an active attacker
```
martinthomson, line 544: terminate the connection or terminate the connection attempt?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,53 @@ is the effective TargetName:
...
+
+A client is called "SVCB-reliant" if it connects only to the indicated
+ServiceMode endpoints, and "SVCB-optional" otherwise.  Clients for pre-existing
+protocols SHALL implement SVCB-optional behavior (except as noted in
+{{client-failures}} and {{ech-client-behavior}}).
+
+SVCB-optional clients SHOULD issue in parallel any other DNS queries that might
+be needed for connection establishment.  SVCB-optional clients SHALL append an
+alternative endpoint consisting of the final value of $QNAME, the authority
+endpoint's port number, and no SvcParams, and MAY append the authority endpoint.
```
martinthomson, line 533: I don't understand this.  Are you trying to back-fill a default value for SVCB in the case you don't need one?  Why not just say "if the endpoint does not require SVCB, then it can use A/AAAA records when SVCB is not available"?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -670,6 +664,15 @@ in cache before performing any followup queries.
 With these optimizations in place, and conforming DNS servers,
 using SVCB does not add network latency to connection setup.
 
+To improve performance when using a non-conforming recursive resolver,
+clients SHOULD issue speculative A and AAAA queries based on a predicted value
+of TargetName.  In the absence of better information, clients SHOULD predict
+that TargetName will be the service name for the initial query and "." for
+subsequent queries (see {zone-performance}).
```
martinthomson, line 671: This isn't something you recommend.  You just need to say that the reason you issue A/AAAA in parallel is that you are predicting that TargetName is present and that TargetName is ".", which is what {{zone-performance}} (two curlies!) recommends for those that provision records.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -699,7 +702,8 @@ though the other record in the RRSet has higher priority.
 
 If none of the SVCB records are consistent
 with any active or in-progress connection,
-clients must proceed as described in Step 3 of the procedure in {{client-behavior}}.
+clients must proceed with connection establishment as described in
```
martinthomson, line 705:
```suggestion
clients proceed with connection establishment as described in
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1048,8 +1052,9 @@ if a usable Alt-Svc value is available in the local cache.
...
 client connection procedure ({{client-behavior}}).
 
-For clients that implement support for ECH, the interaction between
-HTTPS RRs and Alt-Svc is described in {{ech-client-behavior}}.
+Clients that implement support for ECH MUST perform the HTTPS RR query first,
+and MUST only make use of Alt-Svc when operating in SVCB-optional mode (see
+{{ech-client-behavior}}).
 
 This specification does not alter the DNS queries performed when connecting
 to an Alt-Svc hostname (typically A and/or AAAA only).
```
martinthomson, line 1060: Citation?

martinthomson (commented):
This is a lot more changes than the other options, but the general shape is better.  I think that you have added a bunch of things that improve the spec in other ways here, so it is definitely more disruptive, but worthwhile.

draft-ietf-dnsop-svcb-https.md
bemasc: I'm not sure I understand your concern here, but I hope these changes have clarified it somewhat.

draft-ietf-dnsop-svcb-https.md
bemasc: There are actually _two_ "default" fallbacks listed here, in a specific order and with different normativity.  The main motivation is to support using AliasMode to point to a hosting provider that doesn't use SVCB.

I've adjusted this text to improve clarity (and avoid assuming that non-SVCB connection means A/AAAA, since it could also be SRV or something else).

draft-ietf-dnsop-svcb-https.md
bemasc: Connection attempt.  Reworded.

draft-ietf-dnsop-svcb-https.md
bemasc: {{zone-performance}} actually didn't recommend that.  I've moved this recommendation from here to there.

draft-ietf-dnsop-svcb-https.md
bemasc: Clarified

draft-ietf-dnsop-svcb-https.md
martinthomson: I think that the key difference between reliant and optional is whether the information in the ServiceMode is critical to operation of the client.

Related question: does an SVCB-optional client switch to SVCB-reliant if it encounters an AliasMode record?  Or, can it follow to the target name and just consume A/AAAA (or whatever) instead, without ever having seen a ServiceMode record?

draft-ietf-dnsop-svcb-https.md
martinthomson: As the text still says that you issue A/AAAA when you hit a ServiceMode record, I'm now more confused by this.

I'm still totally short on clarity here.  What is being appended to?  And I still don't know what "the authority endpoint" is or what it is appended to.  Are you saying that if you exhaust SVCB lookups and you are don't need the ServiceForm information, you could just pretend to have received a SVCB ServiceForm record that consists of $QNAME, the port, and no SvcParams?  Why say that at all?  What advantage does this have over letting the client fall back to non-SVCB resolution?

draft-ietf-dnsop-svcb-https.md
bemasc: > I think that the key difference between reliant and optional is whether the information in the ServiceMode is critical to operation of the client.

Agreed.  How does that differ from the definition given here?

> Related question: does an SVCB-optional client switch to SVCB-reliant if it encounters an AliasMode record?

No.

> Or, can it follow to the target name and just consume A/AAAA (or whatever) instead, without ever having seen a ServiceMode record?

Yes.  That's important, to support use of AliasMode pointing to a hosting provider that doesn't have a SVCB record.

Additionally, as currently specified, SVCB-optional clients are generally encouraged (and in some cases required) to fall back to non-SVCB connection if all the SVCB alternative endpoints are tried and they all fail. (See #256 for some debate about that.)

draft-ietf-dnsop-svcb-https.md
martinthomson: I was noting that this might need citation, though the clarification is fine.

draft-ietf-dnsop-svcb-https.md
martinthomson: >  How does that differ from the definition given here?

The definition here concentrates on what the client does (connect to the ServiceForm endpoint), rather than the reasons for the same.

Does the draft explain the AliasForm-CNAME isomorphism?  There's a lot of text, but I can't see anything there.  Seems important.

draft-ietf-dnsop-svcb-https.md
bemasc: > As the text still says that you issue A/AAAA when you hit a ServiceMode record, I'm now more confused by this.

Connection using a ServiceMode record always requires an address lookup for TargetName.  Connection _not_ using SVCB doesn't strictly require an address lookup ... who knows what the non-SVCB connection procedure is for an arbitrary protocol?

> What is being appended to?

The priority-ordered list of alternative endpoints.

> And I still don't know what "the authority endpoint" is or what it is appended to.

That's defined in [the definitions section](https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-02#section-1.4).

> Are you saying that if you exhaust SVCB lookups and you are don't need the ServiceForm information, you could just pretend to have received a SVCB ServiceForm record that consists of $QNAME, the port, and no SvcParams?

Yes.

> Why say that at all?

Mostly to handle the case where there is an AliasMode record but no ServiceMode record.  Adding this endpoint unconditionally for fallback is not necessary but seems like it could improve reliability (one more endpoint to try) without expanding the DNS response.

draft-ietf-dnsop-svcb-https.md
bemasc: > The definition here concentrates on what the client does (connect to the ServiceForm endpoint), rather than the reasons for the same.

OK, I've reversed the polarity of the definitions.  Does that help?

> Does the draft explain the AliasForm-CNAME isomorphism?

It's not really an isomorphism.  [The AliasMode section](https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-02#section-2.4.2) discusses the relationship in some detail.  I'm happy to improve the description but I think it's out of scope for this PR.

ericorth:
All this seems reasonable to me.  From the perspective of an HTTPS client, this is a little harder to follow than the previous text (as I would expect since the previous text was targeted really to just telling HTTPS clients what to do), but it's still reasonable enough and hopefully resolves concerns regarding clients other than HTTPS clients.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,54 @@ is the effective TargetName:
...
+A client is called "SVCB-optional" if it can connect without the use of
+ServiceMode records, and "SVCB-reliant" otherwise.  Clients for pre-existing
+protocols SHALL implement SVCB-optional behavior (except as noted in
+{{client-failures}} and {{ech-client-behavior}}).
+
+SVCB-optional clients SHOULD issue in parallel any other DNS queries that might
+be needed for connection establishment.  SVCB-optional clients SHALL append an
+alternative endpoint consisting of the final value of $QNAME, the authority
+endpoint's port number, and no SvcParams, to be attempted before falling back
+to non-SVCB connection modes.
```
martinthomson, line 534:
```suggestion
endpoint's port number, and no SvcParams, to the list of alternative endpoints, which is 
attempted before falling back to non-SVCB connection modes.
```

Does this need to be explicitly ServiceMode?

martinthomson (approved):

moonshiner:
impartial third party:  I like these phrasing adjustments.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,55 @@ is the effective TargetName:
...
+for a service, as performed by the client.  It MUST be implemented as follows:
 
-1. Let $ADDR_QNAME be the service name.  Let $SVCB_QNAME be the service name
-   plus appropriate prefixes for the scheme (see {{svcb-names}}).
+1. Let $QNAME be the service name plus appropriate prefixes for the
+   scheme (see {{svcb-names}}).
 
-2. In parallel, issue AAAA/A queries for $ADDR_QNAME
-   and a SVCB query for $SVCB_QNAME.
-   The answers for these may or may not include CNAME pointers
```
enygren, line 506: Do we want to keep a reference to potential of CNAMEs in-place, or this is adequately covered elsewhere and in examples?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,55 @@ is the effective TargetName:
...
+A client is called "SVCB-optional" if it can connect without the use of
+ServiceMode records, and "SVCB-reliant" otherwise.  Clients for pre-existing
+protocols SHALL implement SVCB-optional behavior (except as noted in
+{{client-failures}} and {{ech-client-behavior}}).
+
+SVCB-optional clients SHOULD issue in parallel any other DNS queries that might
+be needed for connection establishment.  SVCB-optional clients SHALL append an
+alternative endpoint consisting of the final value of $QNAME, the authority
+endpoint's port number, and no SvcParams, to the list of alternative endpoints, which is 
+attempted before falling back to non-SVCB connection modes.
```
enygren, line 535: The case somewhat ambiguous here is the scenario where an AliasMode record points to a name with only A/AAAA records.  It may help to call this out explicitly.  Which may be clarifying that each time $QNAME changes, SVCB-optional clients SHOULD issue AAAA/A queries.  We may also want some text such as:  "Note that SVCB-optional clients should be prepared to handle the case where the TargetName of an AliasMode SVCB RR has only AAAA/A records and no ServiceMode record."

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1133,24 +1137,18 @@ identifies the desired service.
...
-1. Perform connection establishment using HTTPS RRs as described in
-   {{client-behavior}}.  After step 4, if there were compatible HTTPS RRs,
-   they all had an "echconfig" key, and attempts to connect to them all failed,
-   terminate connection establishment.
-2. If the client implements Alt-Svc, try to connect using any entries from
-   the Alt-Svc cache.
-3. Continue connection establishment as in {{client-behavior}} if necessary.
+1. Perform SVCB resolution as described in {{client-behavior}}.
+2. If SVCB resolution succeeded, and all alternative endpoints have an "echconfig"
+   key, use SVCB-reliant connection establishment.
```
enygren, line 1148: Does this mean that ECH-capable SVCB-optional clients can't use Alt-Svc until we define an Alt-Svc-bis that addresses relevant issues?

enygren (commented):
This does seem to be cleaner and more readable and the previous text, so does seem to be a general improvement.

draft-ietf-dnsop-svcb-https.md
enygren: Whereas with a precise read of the text does cover the "the case where there is an AliasMode record but no ServiceMode record" it is not obvious to a reader as shown by both Martin and I not quite getting this until reading a few times.  At a minimum we should have a parenthetical which concretely calls out this case.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I've added a clarification for that case.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, added to Step 3.

draft-ietf-dnsop-svcb-https.md
bemasc: No.  ECH-capable SVCB-optional clients can use Alt-Svc unless all returned ServiceForm records include `echconfig`.

enygren: Is there a reason we're missing what had been the previous steps 2+3?  (ie, to handle the case where the conditional in step 2 doesn't hold?)

bemasc: It's effectively moved to line 1054 (in the Alt-Svc section, instead of talking about Alt-Svc in the ECH section).

enygren: In-general this numbered list is still confusing since it doesn't have an "else" clause to 2.  Perhaps switching it to be a bulletted list or just some flat text would be better?

enygren: Perhaps:

Accordingly, ECH-capable SVCB-optional clients MUST use SVCB-reliant connection establishment if SVCB resolution succeeded (following {{client-behavior}}), and if all alternative endpoints have an "echconfig" key.

bemasc: OK, done.

enygren (approved):
Looks good.  Do we want to define HTTPS as an explicit SVCB-optional protocol?  (Perhaps a separate issue so we can ship this?)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -495,59 +495,58 @@ is the effective TargetName:
 
 # Client behavior {#client-behavior}
 
-A SVCB-aware client selects an endpoint for a service using the following
-procedure:
+"SVCB resolution" is the process of enumerating the priority-ordered endpoints
+for a service, as performed by the client.  It MUST be implemented as follows:
```
martinthomson, line 499:
```suggestion
for a service, as performed by the client.  SVCB resolution is implemented as follows:
```

I think that the MUST here is too strong.  The W3C and WHATWG, who specialize in writing algorithmic specifications, only ever say that the outcome should be equivalent to (or indistinguishable from) the algorithm as described.

martinthomson (approved):

bemasc:
> Looks good. Do we want to define HTTPS as an explicit SVCB-optional protocol?

I added a tiny clarification: https://github.com/MikeBishop/dns-alt-svc/pull/291/commits/a382a5783be10fd0507c5618adbff79a7a962871.  If you'd like to expand that, let's do it in a separate PR.

PR #292, opened 2021-02-17
Add changelog for -03

bemasc:


enygren (approved):
Looks fine unless we want to address #293 before publishing.  (We can also do a -04 with that.)

bemasc:
OK.  Let's push out -03 and follow up with any additional clarifications as needed.

PR #294, opened 2021-02-18
Proposal: Recommend not exceeding chain length 8

bemasc:
This change instructs zone operators not to create excessively long
alias chains.  This change does not affect the normative requirements
for resolvers, which are only required to follow at least one AliasMode
alias.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1196,6 +1196,8 @@ foo://foo.example.com:8080, this might look like:
 
 Domain owners SHOULD avoid using a TargetName that is below a DNAME, as
 this is likely unnecessary and makes responses slower and larger.
+Also, zone structures that require following more than 8 aliases
+(counting both AliasMode and CNAME records) are NOT RECOMMENDED.
 
 ## Examples
 


tfpauly (approved):
8 is a sensible limit. If anything, it could be lower.

ericorth:
Since other parties are only required to follow at least one, I would make the zone operator recommendation similarly short to maximize compatibility.  I don't think that would be an overburdensome recommendation since it's only a SHOULD, not a MUST.

enygren (approved):

enygren:
8 seems reasonable as proposed

bemasc:
@ericorth, Are you saying you would prefer a lower number here?  Note that, as currently phrased, this is the combined total number of CNAME and AliasMode records in the chain, and this section already says "domain owners SHOULD minimize the use of AliasMode records".

I'd like to avoid setting too tight a bound here, given previous discussions in DNSOP where participants emphasized that resolvers must be free to choose their own limits.  While this is not quite the same context, it seems like smaller numbers may make it harder to get consensus.

ericorth:
I have a weak preference for a lower number but won't fight for it if you think it would cause consensus-building issues.

How about a compromise that might be easier for consensus.  Leave the formal "SHOULD" recommendation at 8, but include a sentence after it pointing out that shorter chain links, ideally at most 1, are best for compatibility and performance.  (Assuming there's not already language to about that effect that I'm forgetting about.)

enygren:
If this was just AliasMode chains something shorter would be fine.  With this also including CNAMEs, I don't think we can go below 8.

ericorth:
Fair enough.  Do you think it would be more reasonable to, after the overall chain recommendation of 8, to more explicitly recommend a short number of SVCB-alias links in that chain? Or would that just be too redundant with the "should avoid alias" part already in there?

bemasc:
> Fair enough. Do you think it would be more reasonable to, after the overall chain recommendation of 8, to more explicitly recommend a short number of SVCB-alias links in that chain? Or would that just be too redundant with the "should avoid alias" part already in there?

That seems redundant to me.  It's also a compatibility issue (not performance), so this doesn't seem like exactly the right place for it.  I feel like we already have relevant normative text about the number of AliasModes twice (once in the AliasMode section and again in this section) so a third version is not needed.

ericorth:
Okay.  Sounds like I'm just being forgetful on how much relevant recommendation is already in the draft, and that we're already in a good place.  So recommended full chain length maximum of 8 SGTM.

enygren (approved):

Issue #298, opened 2021-02-25
"echconfig" is an "ECHConfigs" struct in ECH


sayrer:
> The SvcParamKey to enable Encrypted ClientHello (ECH) is "echconfig".

The ECH draft says this is a list of "ECHConfig" structs: "ECHConfigs".

> The client-facing server advertises a sequence of ECH configurations to clients, serialized as follows.
> ECHConfig ECHConfigs<1..2^16-1>;

This pluralization mismatch tripped me up testing ECH draft-09.

davidben:
Oh oops, I guess this slipped through while this was going through some churn on the ECH side. I'd support renaming the SVCB name to match.

bemasc:
This was deliberate.  From the DNS perspective, this key contains a unitary base64 blob.  Also, even the list-valued keys here have singular names (alpn=..., mandatory=..., ipv4hint=...).

sayrer:
The draft does clarify this distinction in Section 9.

Maybe section 6.3 could say
> The SvcParamKey to enable Encrypted ClientHello (ECH) is "echconfig". Its value, an ECHConfigs vector [ECH], is defined in Section 9.  It is applicable to most TLS-based protocols.

A little bit of redundancy, but the naming is confusing, imo.

davidben:
The singular "echonfig" name seems likely to confuse things. Plain "ech" seems reasonable. (Perhaps we should also re-re-re-reconsider the names in ECH. "ECHConfigs" may have been too clever... ECHConfigList?)

sayrer:
Now would be the right time to do either of those changes. It looks like ECH will need to be revised for the newest HPKE draft, and the wire format of the DNS data will change at that time, too. See https://tools.ietf.org/rfcdiff?url1=https://tools.ietf.org/id/draft-ietf-tls-esni-09.txt&url2=https://tlswg.github.io/draft-ietf-tls-esni/draft-ietf-tls-esni.txt

davidben:
Yes, there are a lot of revisions we know we need for ECH. :-P

moonshiner:
from a DNS implementer/operator point of view, I would like to +1 this small bit of redundancy.

bemasc:
The original context for singular/plural is in #127.

I would prefer to change the name in draft-ietf-tls-esni.  In my view, having structs there named for both the plural and the singular is the main source of confusion.  However, changing the name here is certainly still possible, if increasingly annoying.

[...]

PR #299, opened 2021-02-26
s/echconfig/ech/

bemasc:
See #298 and https://github.com/tlswg/draft-ietf-tls-esni/pull/391

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -188,7 +188,7 @@ intended destination to all entities along the network path.
 ## Parameter for Encrypted ClientHello
 
 This document also defines a parameter for Encrypted ClientHello {{!ECH}}
-keys. See {{echconfig}}.
+keys. See {{ech-param}}.
 
 ## Terminology
 
@@ -356,7 +356,7 @@ to indicate that "foo://api.example.com:8443" is aliased to "svc4.example.net".
 The owner of example.net, in turn, could publish this record:
 
     svc4.example.net.  7200  IN SVCB 3 svc4.example.net. (
-        alpn="bar" port="8004" echconfig="..." )
+        alpn="bar" port="8004" ech="..." )
 
 to indicate that these services are served on port number 8004,
 which supports the protocol "bar" and its associated transport in
@@ -487,7 +487,7 @@ is the effective TargetName:
 
     example.com.      7200  IN HTTPS 0 svc.example.net.
     svc.example.net.  7200  IN CNAME svc2.example.net.
-    svc2.example.net. 7200  IN HTTPS 1 . port=8002 echconfig="..."
+    svc2.example.net. 7200  IN HTTPS 1 . port=8002 ech="..."
     svc2.example.net. 300   IN A     192.0.2.2
     svc2.example.net. 300   IN AAAA  2001:db8::2
 
@@ -587,7 +587,7 @@ Providing the proxy with the final TargetName has several benefits:
 * It allows the client to use the SvcParams, if present, which is
   only usable with a specific TargetName.  The SvcParams may
   include information that enhances performance (e.g. alpn) and privacy
-  (e.g. echconfig).
+  (e.g. ech).
 
 * It allows the service to delegate the apex domain.
 
@@ -682,7 +682,7 @@ If an address response arrives before the corresponding SVCB response, the
 client MAY initiate a connection as if the SVCB query returned NODATA, but
 MUST NOT transmit any information that could be altered by the SVCB response
 until it arrives.  For example, a TLS ClientHello can be altered by the
-"echconfig" value of a SVCB response ({{svcparamkeys-echconfig}}).  Clients
+"ech" value of a SVCB response ({{svcparamkeys-ech}}).  Clients
 implementing this optimization SHOULD wait for 50 milliseconds before
 starting optimistic pre-connection, as per the guidance in
 {{HappyEyeballsV2}}.
@@ -695,12 +695,12 @@ For example, suppose the client receives this SVCB RRSet for a protocol
 that uses TLS over TCP:
 
     _1234._bar.example.com. 300 IN SVCB 1 svc1.example.net. (
-        echconfig="111..." ipv6hint=2001:db8::1 port=1234 )
+        ech="111..." ipv6hint=2001:db8::1 port=1234 )
                                    SVCB 2 svc2.example.net. (
-        echconfig="222..." ipv6hint=2001:db8::2 port=1234 )
+        ech="222..." ipv6hint=2001:db8::2 port=1234 )
 
 If the client has an in-progress TCP connection to `[2001:db8::2]:1234`,
-it MAY proceed with TLS on that connection using `echconfig="222..."`, even
+it MAY proceed with TLS on that connection using `ech="222..."`, even
 though the other record in the RRSet has higher priority.
 
 If none of the SVCB records are consistent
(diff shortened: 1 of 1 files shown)


enygren (approved):

marka63:
This should not be happening.  You are abusing the code point allocation.  The record format is supposed to be FIXED when the code point is allocated.  I complained before the code point was allocated that it was not the right time in the development process to perform the allocation.  When the code point is allocated the DNS record should be stable. That is both WIRE and PRESENTATION formats.  Remember master file format is a INTERCHANGE specification.

The DNS has a reserved range for private types which are fine for experimentation.

Mark

> On 27 Feb 2021, at 07:13, Benjamin M. Schwartz <notifications@github.com> wrote:
> 
> See #298 and tlswg/draft-ietf-tls-esni#391
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/MikeBishop/dns-alt-svc/pull/299
> 
> Commit Summary
> 
> 	• s/echconfig/ech/
> File Changes
> 
> 	• M draft-ietf-dnsop-svcb-https.md (60)
> Patch Links:
> 
> 	• https://github.com/MikeBishop/dns-alt-svc/pull/299.patch
> 	• https://github.com/MikeBishop/dns-alt-svc/pull/299.diff
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

-- 
Mark Andrews, ISC
1 Seymour St., Dundas Valley, NSW 2117, Australia
PHONE: +61 2 9871 4742              INTERNET: marka@isc.org

enygren:
Does this argue that we should change the codepoint and abandon echconfig/5 as part of the switch to the "ech" key name?  (In some ways this is happening due to the dependent ECH draft making changes.)  That then makes this a key addition rather than a change.

ericorth:
+1.  If there's any concern whatsoever about compatibility problems from renaming "echconfig", let's add a new service parameter codepoint for the renamed param.  In a quick review of the current draft, I don't see any reason that would break stability of SVCB/HTTPS themselves.  I could see an argument if "echconfig" were in the automatically-mandatory list, but it's not.

bemasc:
As a practical matter, I don't think we need to go through a deprecation procedure.  The "echconfig" contents are defined by a draft (draft-ietf-tls-esni) that is not finished, and indeed has multiple breaking changes in progress.  I don't think anyone who is using "echconfig" in a zone today expects it to be interoperable.  The only deployment I'm aware of exists specifically for interoperability testing of draft versions of ECH, and that is the only appropriate use at this stage.

As for the Early Allocation rules, note that "echconfig" is not part of the SVCB draft standard presentation format.  Rather, it is a proposed entry in an IANA registry that doesn't exist yet.

While I think it's annoying to have to burn a codepoint in order to change the name, I'm happy to do so if necessary.  If that's the consensus, I think we might as well wait until the ECH draft is final before making any change, to avoid having to do this again.

enygren:
How do we get to consensus here?  Options seem to be:

1. Stick with echconfig and current codepoint.  (Not preferable.)
2. Switch to ech and use new codepoint.   (Burns a codepoint, not strictly necessary.)
3. Switch to ech and keep existing codepoint.   (Changes the presentation to wire format mapping of an existing codepoint.)

I could go with either 2/3.  @marka63 seems to strongly prefer 2.  @bemasc seems to strongly prefer 3.

bemasc:
I am fine with any of these three things.  From most preferred to least: 3, 1, 2.  My main point is that, if we go with 2, we should delay the switch as long as possible, to wait until after any possible compatibility-breaking changes in ECH.  (Whereas if we go with 3, we should do it as soon as possible to minimize disruption.)

enygren:
What about going with 2, but leaving instructions for that switch to happen once TLS ECH passes its last-call? 
Can we do something like updating the text to use "ech" but leave the codepoint listed as "to-be-assigned"
and then assign it shortly before publication?

bemasc:
@enygren That sounds complicated.  I don't know how to explain it to the RFC Editor, and I don't know what advice I would give to an implementer.  It seems a lot easier to just change nothing (which might be less confusing after s/ECHConfigs/ECHConfigList/).

davidben:
Changing nothing will remain confusing because ECH has a structure called ECHConfig, and that is *not* the structure that goes in SVCB. The intent hasn't been for SVCB to contain an ECHConfig since draft-ietf-tls-esni-07, the very first draft that used the ECH name. The "echconfig" name was a mistake from the beginning.

Given SVCB and ECH are getting published at different times, should the code point definition just be moved out of SVCB and into ECH? Though this is somewhat moot because, on the compatibility side, the ECHConfigList structure is already internally versioned.

bwelling:
An alternative solution could be to change the parameter name to `ech`, and support `echconfig` as an alias.  I think that would satisfy at least some backwards compatibility issues.

bemasc:
Most of the SVCB parameters are singular-name, multiple-value, so `echconfig=<ECHConfigList>` seems to fit the pattern.  However, I'm certainly happy to change it if it still seems confusing.

Maintaining an alias is possible, but I would rank it lower than the other three options.  It makes "echconfig" different from all other parameters, it seems to create a permanent burden to deal with a temporary problem, and I'm not sure it actually satisfies the desire for zone-file compatibility with existing draft implementations.  (Of course, burning the codepoint is even worse for zone file compatibility.)

marka63:
You have already introduced incompatible changes with alpn comma escape processing.  All these incompatible changes are, in part, why BIND hasn't shipped a version which supports these record types despite the requests for them from our customers.

Whether the type codes and mnemonics are already burnt or not I don't know.  Every vendor that has already shipped parsers could treat the changes as incompatible bug fixes and one could get away with it.

sayrer:
> How do we get to consensus here? Options seem to be:
> 
> 1. Stick with echconfig and current codepoint.  (Not preferable.)
> 2. Switch to ech and use new codepoint.   (Burns a codepoint, not strictly necessary.)
> 3. Switch to ech and keep existing codepoint.   (Changes the presentation to wire format mapping of an existing codepoint.)
> 
> I could go with either 2/3. @marka63 seems to strongly prefer 2. @bemasc seems to strongly prefer 3.

3 is good, 2 is fine but seems a bit officious, and 1 seems kind of bad but livable.

So, they're all ok.

chris-wood:
Like @enygren, I'm fine with 3 or 2, but lean towards 3 given the state of this record and its users.

MikeBishop:
I prefer 3.  There's a certain appeal to just letting ECH register the codepoint in their doc, but that seems orthogonal to this question.

davidben:
Yeah, moving the registration mostly relates if we're doing (2) + delay. If we do the delay thing, we may as well move the codepoint IMO.

I also prefer (3), in which case I agree it's orthogonal and doesn't matter much.

bemasc:
For comparison, I've created a draft meta-PR that allocates a new codepoint for "ech": #316.  Hopefully this helps people weigh the options here in terms of complexity, compatibility, and potential for confusion.

tfpauly:
I prefer 3 as well, FWIW.

moonshiner:
I'm with option 3 as well.

bemasc:
OK, I'm going to merge this as-is and publish a new version for further discussion.

marka63:
Hello
I have a business partnership proposal for you. Contact me on my
personal email for more details please :
***@***.***


Regards
Saif

marka63:
Hello
I have a business partnership proposal for you. Contact me on my
personal email for more details please :
***@***.***


Regards
Saif

PR #301, opened 2021-03-02
Remove all shortnames for RFC references

bemasc:
I surveyed the ten most recent RFCs, and many of them use shortname
references, but not for RFCs.


bemasc:
It looks like RFC 8470 is an example that _does_ use shortnames for RFCs, so there's not a clear rule against this.  I think we can leave this as-is until closer to publication.

PR #307, opened 2021-03-19
Add some test-vectors

pieterlexis:
As discussed on the mailinglist, this is a PR with test vectors. I've tried to explain each set of bytes in the wire format. Let me know if there's anything that needs changing.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1810,17 +1810,19 @@ Here, two IPv6 hints are quoted in the presentation format.
     \x20\x01\x0d\xb8\x00\x00\x00\x00
          \x00\x00\x00\x00\x00\x53\x00\x01              # second address
 
-In the next vector, the SvcParamValues are not sorted in presentation format,
-but are sorted in the wire-format.
+In the next vector, neither the SvcParamValues nor the mandatory keys are 
+sorted in presentation format, but are correctly sorted in the wire-format.
 
-    16 foo.example.org alpn=h2,h3-19 mandatory=alpn ipv4hint=192.0.2.1
+    16 foo.example.org (alpn=h2,h3-19 mandatory=ipv4hint,alpn
+                        ipv4hint=192.0.2.1)
 
-    \# 46 (
+    \# 48 (
     00 10                                              ; priority
     03 66 6f 6f 07 65 78 61 6d 70 6c 65 03 6f 72 67 00 ; target
     00 00                                              ; key 0
-    00 02                                              ; param length 2
+    00 04                                              ; param length 4
     00 01                                              ; value: key 1
+    00 04                                              ; value: key 4
     00 01                                              ; key 1
     00 09                                              ; param length 9
     02                                                 ; alpn length 2
@@ -1846,6 +1848,32 @@ but are sorted in the wire-format.
     \x00\x04                                           # key 4
     \x00\x04                                           # param length 4
     \xc0\x00\x02\x01                                   # param value
+
+This last vector has an alpn value with an escaped comma and an escaped
+backslash in two presentation formats.
+
+    16 foo.example.org alpn="f\\\\oo\\,bar,h2"
+    16 foo.example.org alpn=f\\\092oo\092,bar,h2
+
+    \# 35 (
+    00 10                                              ; priority
+    03 66 6f 6f 07 65 78 61 6d 70 6c 65 03 6f 72 67 00 ; target
+    00 01                                              ; key 1
+    00 0c                                              ; param length 12
+    08                                                 ; alpn length 8
+    66 5c 6f 6f 2c 62 61 72                            ; alpn value
+    02                                                 ; alpn length 2
+    68 32                                              ; alpn value
+    )
+
+    \x00\x10                                           # priority
+    \x03foo\x07example\x03org\x00                      # target
+    \x00\x01                                           # key 1
+    \x00\x0c                                           # param length 12
+    \x08                                               # alpn length 8
+    f\oo,bar                                           # alpn value
+    \x02                                               # alpn length 2
+    h2                                                 # alpn value
 
 # Change history
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1702,6 +1702,151 @@ the authoritative.  Thus, splitting the RR type would double, or in
 some cases triple, the load on clients and servers, and would not
 reduce implementation complexity.
 
+# Test vectors
+These test vectors only contain the RDATA portion of SVCB/HTTPS records in
+presentation format, generic format ({{!RFC5397}}) and wire format. The wire
+format uses hexadecimal (\xNN) for each non-ascii byte. As the wireformat is
+long, it is broken into several lines.
+
+## AliasForm
+
+    0 foo.example.com
+
+    \# 19 (
+    00 00                                              ; priority
+    03 66 6f 6f 07 65 78 61 6d 70 6c 65 03 63 6f 6d 00 ; target
+    )
+
+    \x00\x00                                           # priority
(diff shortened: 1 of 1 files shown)


bemasc (approved):
Looks good to me.  I wonder if it would be clearer to use a little less whitespace in the RFC 3597 examples, to highlight the distinction between uint16, uint8, and bytestring sequences, but that's up to you.

Waiting for another coauthor to confirm before merging.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,151 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    \x00\x03                                           # key 3
+    \x00\x02                                           # length: 2 bytes
+    \x00\x35                                           # value
+
+This example has a key that is not registered, its value is unquoted.
+
+    1 foo.example.com key667=hello
+
+    \# 28 (
+    00 10                                              ; priority
```
bwelling, line 1759: This should be 00 01, not 00 10.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,151 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    \x00\x05                                           # length 5
+    hello                                              # value
+
+This example has a key that is not registered, its value is quoted and contains
+a decimal-escaped character.
+
+    1 foo.example.com key667="hello\210qoo"
+
+    \# 32 (
+    00 10                                              ; priority
```
bwelling, line 1778: This should be 00 01, not 00 10.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,151 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    \x02\x9b                                           # key 667
+    \x00\x09                                           # length 9
+    hello\xd2qoo                                       # value
+
+Here, two IPv6 hints are quoted in the presentation format.
+
+    1 foo.example.com ipv6hint="2001:db8::1,2001:db8::53:1"
+
+    \# 57 (
+    00 10                                              ; priority
```
bwelling, line 1796: This should be 00 01, not 00 10.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,151 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+         \x00\x00\x00\x00\x00\x00\x00\x01              # first address
+    \x20\x01\x0d\xb8\x00\x00\x00\x00
+         \x00\x00\x00\x00\x00\x53\x00\x01              # second address
+
+In the next vector, the SvcParamValues are not sorted in presentation format,
+but are sorted in the wire-format.
+
+    16 foo.example.org alpn=h2,h3-19 mandatory=alpn ipv4hint=192.0.2.1
+
+    \# 48 (
```
bwelling, line 1818: This should be \\# 46, not \\# 48

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,151 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    \x03foo\x07example\x03com\x00                      # target
+    \x02\x9b                                           # key 667
+    \x00\x09                                           # length 9
+    hello\xd2qoo                                       # value
+
+Here, two IPv6 hints are quoted in the presentation format.
+
+    1 foo.example.com ipv6hint="2001:db8::1,2001:db8::53:1"
+
+    \# 57 (
```
bwelling, line 1795: This should be \\# 55, not \\# 57

bwelling (commented):
I just tried testing the RFC 3597 for of these test vectors, and found a number of issues.

marka63:
Needed an escaped comma alpn example.

pieterlexis:
> Needed an escaped comma alpn example.

I'll add that as well

@bwelling thanks! I had to hand-craft these values :). I'll fix them

marka63:
add out-of-order mandatory example

pieterlexis:
> add out-of-order mandatory example

I modified an existing vector to showcase this.

bemasc:
@pieterlexis Thanks.  I guess we might as well add an ALPN with an embedded backslash to that list...

pieterlexis:
> @pieterlexis Thanks. I guess we might as well add an ALPN with an embedded backslash to that list...

Done, please check, this was artisinally handcrafted :) (and found an issue in the PowerDNS text writer)

bwelling:
If there are examples of escaping in ALPN parameters, it might also be good to have examples of escaping in the unregistered key (keyNN) format.

pieterlexis:
> If there are examples of escaping in ALPN parameters, it might also be good to have examples of escaping in the unregistered key (keyNN) format.

The section on mandatory says "To enable simpler parsing, this SvcParamValue MUST NOT contain escape sequences."

bwelling:
> > If there are examples of escaping in ALPN parameters, it might also be good to have examples of escaping in the unregistered key (keyNN) format.
> 
> The section on mandatory says "To enable simpler parsing, this SvcParamValue MUST NOT contain escape sequences."

The text for all of the defined formats other than ALPN says that, but the text for the unregistered key format (which is what I was suggesting an example for) does not.  It specifically says that the value is a char-string, whose definition (as far as I can tell) includes escaping.

bemasc:
@bwelling correct.  For example, you can spell `port=123` as `key3=\000\123`.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,178 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    )
+
+    \x00\x01                                           # priority
+    \x03foo\x07example\x03com\x00                      # target
+    \x00\x06                                           # key 6
+    \x00\x20                                           # length 32
+    \x20\x01\x0d\xb8\x00\x00\x00\x00
+         \x00\x00\x00\x00\x00\x00\x00\x01              # first address
+    \x20\x01\x0d\xb8\x00\x00\x00\x00
+         \x00\x00\x00\x00\x00\x53\x00\x01              # second address
```
wtoorop, line 1811:
```suggestion
    1 foo.example.com ipv6hint="2001:db8::1,2001:db8:ffff:ffff:ffff:ffff:203.0.113.255"

    \# 55 (
    00 01                                              ; priority
    03 66 6f 6f 07 65 78 61 6d 70 6c 65 03 63 6f 6d 00 ; target
    00 06                                              ; key 6
    00 20                                              ; length 32
    20 01 0d b8 00 00 00 00 00 00 00 00 00 00 00 01    ; first address
    20 01 0d b8 ff ff ff ff ff ff ff ff cb 00 71 ff    ; second address
    )

    \x00\x01                                           # priority
    \x03foo\x07example\x03com\x00                      # target
    \x00\x06                                           # key 6
    \x00\x20                                           # length 32
    \x20\x01\x0d\xb8\x00\x00\x00\x00
         \x00\x00\x00\x00\x00\x00\x00\x01              # first address
    \x20\x01\x0d\xb8\xff\xff\xff\xff
         \xff\xff\xff\xff\xcb\x00\x71\xff              # second address
```

A suggestion for IPv4 mapped IPv6 that also crosses the 39 characters boundry of IPv6 addresses represented as string that would be the max string size if IPv4 mapped IPv6 was not anticipated.

bemasc: Works for me (as long as that line isn't too wide for TXT format).

wtoorop: Ouch...  that is unfortunately not the case as TXT format has only a width of 69 characters. The example above has 83.

wtoorop:
FYI, we're planning to create a pull request including the commits in this one, with some failure cases added

bemasc:
@pieterlexis Should we close this in favor of #312?

PR #308, opened 2021-03-24
Specify interaction with ECS

bemasc:
This change explains how to make SVCB play nicely with ECS, for both
recursive and authoritative servers.  The proposed behavior is
conservative, staying fully compliant with RFC 7871 (ECS) while allowing
improved caching performance when it is safe to do so.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -597,11 +597,11 @@ Providing the proxy with the final TargetName has several benefits:
 
 # DNS Server Behavior {#server-behavior}
 
-## Authoritative servers
+## Authoritative servers {#authoritative-behavior}
 
 When replying to a SVCB query, authoritative DNS servers SHOULD return
-A, AAAA, and SVCB records in the Additional Section for any
-in-bailiwick TargetNames.  If the zone is signed, the server SHOULD also
+A, AAAA, and SVCB records in the Additional Section for any TargetNames
+that are in the zone.  If the zone is signed, the server SHOULD also
 include positive or negative DNSSEC responses for these records in the Additional
 section.
 
@@ -659,6 +659,31 @@ each RRSet in the Additional section with the same DNSSEC-related records
 that they would send when providing that RRSet as an Answer (e.g. RRSIG, NSEC,
 NSEC3).
 
+## EDNS Client Subnet (ECS)
+
+The EDNS Client Subnet extension (ECS, {{!RFC7871}}) allows recursive
+resolvers to request IP addresses that are suitable for a particular client
+IP range.  IP addresses appear in two places in Authoritative SVCB responses:
+
+* ipv*hint SvcParams in a ServiceMode response.  Clients only use these hints
+when the recursive resolver does not provide A/AAAA records for TargetName in
+the Additional section, as recommended in {{recursive-behavior}}.
+* A/AAAA records returned by the Authoritative server in the Additional
+section of a SVCB response when TargetName is in-zone, as recommended in
+{{authoritative-behavior}}.
+
+A recursive resolver that uses ECS for A/AAAA queries SHOULD send the
+same ECS prefix for SVCB queries unless it knows that any ipv*hint SvcParams
+will not be used, due to the resolver's implementation of Additional section
+processing.  To disable ECS for SVCB, such a resolver SHOULD include an ECS
+extension with a SOURCE PREFIX-LENGTH of zero, to indicate that ECS is
+supported but disabled.
+
+According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
+Additional section\] MUST NOT be tied to a network".  Accordingly,
+when responding to a SVCB query with the ECS extension, Authoritative
+servers MAY omit these records to ensure that prefix-specific A/AAAA
+records are returned by a subsequent or parallel query.
 
 # Performance optimizations {#optimizations}
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -683,7 +683,7 @@ same ECS extension that it would use for an A/AAAA query.
 
 According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
-Resolvers SHOULD assume that any A/AAAA records in the Additional section
+resolvers SHOULD assume that any A/AAAA records in the Additional section
 are network-invariant.  When responding to a SVCB query with an ECS
 extension (even one whose SOURCE PREFIX-LENGTH is zero), an Authoritative
 server SHOULD NOT add any A/AAAA records for TargetName to the response if
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -661,6 +661,14 @@ each RRSet in the Additional section with the same DNSSEC-related records
 that they would send when providing that RRSet as an Answer (e.g. RRSIG, NSEC,
 NSEC3).
 
+According to Section 5.4.1 of {{!RFC2181}}, "Unauthenticated RRs received
+and cached from ... the additional data section ... should not be cached in
+such a way that they would ever be returned as answers to a received query.
+They may be returned as additional information where appropriate.".
+Recursive resolvers therefore MAY cache records from the Additional section
+for use in populating Additional section responses, and MAY cache them
+for general use if they are authenticated by DNSSEC.
+
 ## EDNS Client Subnet (ECS) {#ecs}
 
 The EDNS Client Subnet option (ECS, {{!RFC7871}}) allows recursive
@@ -673,12 +681,11 @@ According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
 resolvers SHOULD treat any records in the Additional section as having
 SOURCE PREFIX-LENGTH zero and SCOPE PREFIX-LENGTH as specified
-in the ECS option, and MAY cache them on this basis.  Authoritative
-servers MUST omit such records if they are not suitable
-for use by any stub resolvers that set SOURCE PREFIX-LENGTH to zero.
-This will cause the resolver to perform a followup query that can receive
-properly tailored ECS.  (This is similar to the usage of CNAME with ECS
-discussed in {{!RFC7871}} Section 7.2.1.)
+in the ECS option.  Authoritative servers MUST omit such records if they are
+not suitable for use by any stub resolvers that set SOURCE PREFIX-LENGTH to
+zero.  This will cause the resolver to perform a followup query that can
+receive properly tailored ECS.  (This is similar to the usage of CNAME with
+ECS discussed in {{!RFC7871}} Section 7.2.1.)
 
 Authoritative servers that omit Additional records can avoid the added
 latency of a followup query by following the advice in {{zone-performance}}.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -672,18 +672,22 @@ the Additional section, as recommended in {{recursive-behavior}}.
 section of a SVCB response when TargetName is in-zone, as recommended in
 {{authoritative-behavior}}.
 
-A recursive resolver that uses ECS for A/AAAA queries SHOULD send the
-same ECS prefix for SVCB queries unless it knows that any ipv*hint SvcParams
-will not be used, due to the resolver's implementation of Additional section
-processing.  To disable ECS for SVCB, such a resolver SHOULD include an ECS
-extension with a SOURCE PREFIX-LENGTH of zero, to indicate that ECS is
-supported but disabled.
+SVCB records SHOULD NOT otherwise vary based on ECS.
+
+A recursive resolver that always returns A/AAAA records for TargetName in the
+Additional section, as recommended in {{recursive-behavior}}, SHOULD include
+an ECS extension with a SOURCE PREFIX-LENGTH of zero in SVCB queries, to
+indicate that ECS is supported but is disabled for this query.  If the
+resolver might not return these Additional records, it SHOULD send the
+same ECS extension that it would use for an A/AAAA query.
 
 According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
-when responding to a SVCB query with the ECS extension, Authoritative
-servers MAY omit these records to ensure that prefix-specific A/AAAA
-records are returned by a subsequent or parallel query.
+Resolvers SHOULD assume that any A/AAAA records in the Additional section
+are network-invariant.  When responding to a SVCB query with an ECS
+extension (even one whose SOURCE PREFIX-LENGTH is zero), an Authoritative
+server SHOULD NOT add any A/AAAA records for TargetName to the response if
+TargetName has any A/AAAA records that are tied to a network.  
 
 # Performance optimizations {#optimizations}
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -661,7 +661,7 @@ NSEC3).
 
 ## EDNS Client Subnet (ECS)
 
-The EDNS Client Subnet extension (ECS, {{!RFC7871}}) allows recursive
+The EDNS Client Subnet option (ECS, {{!RFC7871}}) allows recursive
 resolvers to request IP addresses that are suitable for a particular client
 IP range.  IP addresses appear in two places in Authoritative SVCB responses:
 
@@ -676,16 +676,16 @@ SVCB records SHOULD NOT otherwise vary based on ECS.
 
 A recursive resolver that always returns A/AAAA records for TargetName in the
 Additional section, as recommended in {{recursive-behavior}}, SHOULD include
-an ECS extension with a SOURCE PREFIX-LENGTH of zero in SVCB queries, to
+an ECS option with a SOURCE PREFIX-LENGTH of zero in SVCB queries, to
 indicate that ECS is supported but is disabled for this query.  If the
 resolver might not return these Additional records, it SHOULD send the
-same ECS extension that it would use for an A/AAAA query.
+same ECS option that it would use for an A/AAAA query.
 
 According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
 resolvers SHOULD assume that any A/AAAA records in the Additional section
 are network-invariant.  When responding to a SVCB query with an ECS
-extension (even one whose SOURCE PREFIX-LENGTH is zero), an Authoritative
+option (even one whose SOURCE PREFIX-LENGTH is zero), an Authoritative
 server SHOULD NOT add any A/AAAA records for TargetName to the response if
 TargetName has any A/AAAA records that are tied to a network.  
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -672,8 +672,10 @@ SHOULD include the same ECS option in SVCB queries as in A/AAAA queries.
 According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
 resolvers SHOULD treat any records in the Additional section as having
-SCOPE PREFIX-LENGTH zero, and MAY cache them on this basis.  Authoritative
-servers MUST omit such records if they are not suitable for global use.
+SOURCE PREFIX-LENGTH zero and SCOPE PREFIX-LENGTH as specified
+in the ECS option, and MAY cache them on this basis.  Authoritative
+servers MUST omit such records if they are not suitable
+for use by any stub resolvers that set SOURCE PREFIX-LENGTH to zero.
 This will cause the resolver to perform a followup query that can receive
 properly tailored ECS.  (This is similar to the usage of CNAME with ECS
 discussed in {{!RFC7871}} Section 7.2.1.)
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -605,6 +605,8 @@ that are in the zone.  If the zone is signed, the server SHOULD also
 include positive or negative DNSSEC responses for these records in the Additional
 section.
 
+See {{ecs}} for exception.
+
 ## Recursive resolvers {#recursive-behavior}
 
 Recursive resolvers that are aware of SVCB SHOULD help the client to
@@ -659,7 +661,7 @@ each RRSet in the Additional section with the same DNSSEC-related records
 that they would send when providing that RRSet as an Answer (e.g. RRSIG, NSEC,
 NSEC3).
 
-## EDNS Client Subnet (ECS)
+## EDNS Client Subnet (ECS) {#ecs}
 
 The EDNS Client Subnet option (ECS, {{!RFC7871}}) allows recursive
 resolvers to request IP addresses that are suitable for a particular client
@@ -685,9 +687,10 @@ According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
 resolvers SHOULD assume that any A/AAAA records in the Additional section
 are network-invariant.  When responding to a SVCB query with an ECS
-option (even one whose SOURCE PREFIX-LENGTH is zero), an Authoritative
-server SHOULD NOT add any A/AAAA records for TargetName to the response if
-TargetName has any A/AAAA records that are tied to a network.  
+option (even one whose SOURCE PREFIX-LENGTH is zero), Authoritative servers
+must omit such records if they wish to serve subnet-specific addresses.
+Such Authoritative servers can avoid the added latency of a followup query
+by setting TargetName to the value recommended in {{zone-performance}}.
 
 # Performance optimizations {#optimizations}
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -671,12 +671,15 @@ SHOULD include the same ECS option in SVCB queries as in A/AAAA queries.
 
 According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
 Additional section\] MUST NOT be tied to a network".  Accordingly,
-resolvers SHOULD assume that any A/AAAA records in the Additional section
-are network-invariant.  When responding to a SVCB query with an ECS
-option (even one whose SOURCE PREFIX-LENGTH is zero), Authoritative servers
-must omit such records if they wish to serve subnet-specific addresses.
-Such Authoritative servers can avoid the added latency of a followup query
-by setting TargetName to the value recommended in {{zone-performance}}.
+resolvers SHOULD treat any records in the Additional section as having
+SCOPE PREFIX-LENGTH zero, and MAY cache them on this basis.  Authoritative
+servers MUST omit such records if they are not suitable for global use.
+This will cause the resolver to perform a followup query that can receive
+properly tailored ECS.  (This is similar to the usage of CNAME with ECS
+discussed in {{!RFC7871}} Section 7.2.1.)
+
+Authoritative servers that omit Additional records can avoid the added
+latency of a followup query by following the advice in {{zone-performance}}.
 
 # Performance optimizations {#optimizations}
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -605,7 +605,7 @@ that are in the zone.  If the zone is signed, the server SHOULD also
 include positive or negative DNSSEC responses for these records in the Additional
 section.
 
-See {{ecs}} for exception.
+See {{ecs}} for exceptions.
 
 ## Recursive resolvers {#recursive-behavior}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +659,31 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+* ipv*hint SvcParams in a ServiceMode response.  Clients only use these hints
+when the recursive resolver does not provide A/AAAA records for TargetName in
+the Additional section, as recommended in {{recursive-behavior}}.
+* A/AAAA records returned by the Authoritative server in the Additional
+section of a SVCB response when TargetName is in-zone, as recommended in
+{{authoritative-behavior}}.
+
+A recursive resolver that uses ECS for A/AAAA queries SHOULD send the
+same ECS prefix for SVCB queries unless it knows that any ipv*hint SvcParams
+will not be used, due to the resolver's implementation of Additional section
```
enygren, line 677: I'm not sure I follow the "unless it knows they will not be used".  How does the resolver know that ipv*hint might not be used?

bemasc: Clients only use ipv*hint if the response doesn't include A+AAAA records for TargetName in the Additional section.  If the recursive resolver knows that it is going to include those additional address records in the response, then it knows the hints will never be used.

enygren: Where this gets confusing is with the subsequent guidance that records in the Additional section should not be included if ECS is being used.   Isn't the guidance effectively then "Send ECS so SVCB records can be usable with ipv*hint.  Plan to not send Additional records." ?

mmuhaime: I think the proposal is saying: 
when a recursive resolver does not include additional records for the client, it knows hints may be used by the client, so it needs to ensure that the ipv4hint answer is answered properly honoring ECS info. So the recursive needs to send ECS to the authorities. Whether the auths include additionals or not in this case (that is the the subject of "According to Section 7.3.1 of {{!RFC7871}}, " paragraph) does not matter because the recursive is not including the additionals in its response to the client anyway. 

When recursive resolver does include additional records for the client, ipv[46]hint should not be used, so the ipv[46]hint from auths don't need to vary by ECS, hence the /0 suggestion to avoid extra per-subnet caching. Additionally, in this case, the recommendation in the draft is for auths to not return additionals if they think the additionals should also vary by ECS. The recursive, not finding additionals, will subsequently ask A/AAAA queries with proper ECS and will include ECS based answers in in its additional section that it sends to clients. I think the following is implicit: if auths do respond with additionals then recrusives will use those non-ECS based additionals in its response to the client. 

This seems ok but I will note that Section 7.3.1 and 7.4 of RFC7871 does not say auths should not return additionals when receiving an ECS request, or that recursives should not be using the returned additionals from auths, it just prohibits associating the additionals with ECS data. So the recommendation about auths not responding with additionals seems like a new requirement for SVCB handling, it does not follow directly from RFC7871. Was that desired? I am worried that many auths will return non-ECS additionals when sent /0 and it will take time for people to realize and fix this for SVCB.

jdreed: I agree with Mashooq, I don't like the idea of the recursive potentially changing its ECS behavior for SVCB.  That's going to be a mess for debugging, and will potentially trigger some authoritative behavior that tries to infer whether the resolver supports ECS or not.  Sending a source prefix of 0 means "I understand ECS, but I'm not going to use it", and is used as a signal for a desire for privacy on the part of the end user.   I don't like the idea of overloading that for a SVCB-specific meaning.  

RFC7871 is pretty clear about the circumstances under which a resolver that supports ECS would not send it:  a) when it's asking for a TLD/PSL; b) when it has learned that a given auth doesn't support ECS; c) when it is using an explicit ACL and the auth isn't on that list.  (b) is already pretty ambiguous and highly subjective, I'm not excited about adding another exception to that list.

bemasc: @mmuhaime's interpretation matches my intent here.

I don't think the "/0" behavior is too tricky for Authoritative servers: "see ECS -> no Additional" seems simpler than "see ECS with nonzero prefix -> no Additional".  I also don't think that this will take time to sort out.  Neither of these codepaths exists today, so if we clearly specify which behavior nameservers should implement, I think we can get it.  If we don't specify, we might get a mix of both.

@jdreed Having different behavior for SVCB is entirely in line with RFC 7871, which says

>   Implementations MAY also allow additional configuring of this based
>   on other criteria, such as zone or query type.

Sending "/0" means "I support ECS but I have disabled it for this query".  That's exactly how this PR uses it.

Here are some alternative designs:
* ECS recursives send the same ECS for SVCB as they do for A/AAAA.  This significantly reduces cache efficiency when IP hints are included and the recursive resolver knows they will not be used ("full service resolver").
  - Also bind A/AAAA in the Additional section to the ECS scope of the SVCB response.  This reduces latency (no followup query to get ECS A/AAAA) but violates (updates?) the letter of RFC 7871's prohibition against ECS in the Additional section.  (I don't think it violates the spirit: that requirement is for records that will be used by the recursive resolver itself, which these are not.)
     * And also let full service resolvers treat all SVCB records as cacheable for "/0" (i.e. ECS applies _only_ to the Additional section).  This maximizes performance but substantially deviates from RFC 7871.
* Let full service resolvers disable ECS for SVCB, but don't require nameservers to distinguish "/0" from "no ECS".  This has good caching efficiency but adds latency: it requires the resolver to ignore all Additional A/AAAA records if ECS could be used, even when ECS is not actually in use.

jdreed: OK, fair, I think I misread this initially and thought it was proposing one case under which no ECS option would be sent (vs sending a /0), and that was the part I was most worried about.

ericorth: Possible improvement to help with any confusion over "unless it knows they will not be used, due to resolver's implementation of Additional section processing": maybe convert to something along the lines of "unless it implements Additional section processing, and therefore knows they will be unnecessary to a client that properly prefers A/AAAA results over ipv*hint SvcParams".  As written, the statement is kinda backwards in that it explains the justification ("knows they will not be used") before the clear, concise condition ("implements Additional section processing").

mmuhaime: The original technical proposal seems good to me. I do like @ericorth 's wording better because I think it will make it easier to understand. 
We could also use something like: " If a recursive resolver knows that the client may use the ipv[46]hints, it should use the same ECS prefix for SVCB query that it uses for A/AAAA queries when questioning Authoritative servers.  
 If a recursive resolver knows that the client will not use ipv[46]hints, then it MAY include an ECS with a SOURCE PREFIX-LENGTH of zero to indicate that ECS is supported but disabled. One scenario where the recursive knows ipv[46]hints won't be used is when it includes resolved addresses for the SVCB target names in the additionals section in its response to clients."

bemasc: OK, I've reworked the text for clarity based on the feedback in this thread.

MikeBishop (approved):

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +659,35 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+IP range.  IP addresses appear in two places in Authoritative SVCB responses:
+
+* ipv*hint SvcParams in a ServiceMode response.  Clients only use these hints
+when the recursive resolver does not provide A/AAAA records for TargetName in
+the Additional section, as recommended in {{recursive-behavior}}.
+* A/AAAA records returned by the Authoritative server in the Additional
+section of a SVCB response when TargetName is in-zone, as recommended in
+{{authoritative-behavior}}.
+
+SVCB records SHOULD NOT otherwise vary based on ECS.
```
enygren, line 675: Would it be acceptable to encode geo information into a TargetName based on ECS?  Or to not include QUIC in ASNs where QUIC is known to be broken?  What is the rationale for this SHOULD NOT?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +659,35 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+
+A recursive resolver that always returns A/AAAA records for TargetName in the
+Additional section, as recommended in {{recursive-behavior}}, SHOULD include
+an ECS extension with a SOURCE PREFIX-LENGTH of zero in SVCB queries, to
+indicate that ECS is supported but is disabled for this query.  If the
+resolver might not return these Additional records, it SHOULD send the
+same ECS extension that it would use for an A/AAAA query.
+
```
enygren, line 683: Do we also want a bullet/entry suggesting that recursive resolvers implementing ECS for client performance MAY/SHOULD disable sending Additional A/AAAA records to clients in-favor of being able to use ECS to authorities?  The DNS performance may be better but the overall experience/performance could be much worse.

draft-ietf-dnsop-svcb-https.md
bemasc: The behavior described here means that many (most?) ECS-capable resolvers will send "/0" in SVCB queries, so any attempt to make use of ECS for those features will fail.

Also, the majority of resolvers simply don't do ECS at all, so ECS generally is not usable for anything other than performance optimizations.

draft-ietf-dnsop-svcb-https.md
bemasc: I don't think that makes sense.  If the recursive resolver disables ECS for SVCB, clients will still be using ECS-derived A and AAAA records, so connection performance should not be impaired.

puneetsood (commented):
LGTM with slight wording changes.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +659,35 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+{{authoritative-behavior}}.
+
+SVCB records SHOULD NOT otherwise vary based on ECS.
+
+A recursive resolver that always returns A/AAAA records for TargetName in the
+Additional section, as recommended in {{recursive-behavior}}, SHOULD include
+an ECS extension with a SOURCE PREFIX-LENGTH of zero in SVCB queries, to
+indicate that ECS is supported but is disabled for this query.  If the
+resolver might not return these Additional records, it SHOULD send the
+same ECS extension that it would use for an A/AAAA query.
```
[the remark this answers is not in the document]

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +659,35 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+* ipv*hint SvcParams in a ServiceMode response.  Clients only use these hints
+when the recursive resolver does not provide A/AAAA records for TargetName in
+the Additional section, as recommended in {{recursive-behavior}}.
+* A/AAAA records returned by the Authoritative server in the Additional
+section of a SVCB response when TargetName is in-zone, as recommended in
+{{authoritative-behavior}}.
+
+SVCB records SHOULD NOT otherwise vary based on ECS.
+
+A recursive resolver that always returns A/AAAA records for TargetName in the
```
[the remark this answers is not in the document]
bemasc: OK, changed to "If the resolver plans to resolve A/AAAA records for TargetName and return them in the Additional section, ...".

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +659,35 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+Additional section, as recommended in {{recursive-behavior}}, SHOULD include
+an ECS extension with a SOURCE PREFIX-LENGTH of zero in SVCB queries, to
+indicate that ECS is supported but is disabled for this query.  If the
+resolver might not return these Additional records, it SHOULD send the
```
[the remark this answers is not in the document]
bemasc: OK, changed to "If the resolver does not plan to perform this resolution...".

draft-ietf-dnsop-svcb-https.md
rs: What is the reasoning here? The HTTPSSVC record has great potentials for DNS steering with the use of the priority field to favor closer location while providing some higher latency fallback locations.

Defining per qtype behavior for ECS will also add complexity and is most likely unexpected by most auth DNS operators.

bemasc: DNS steering with priority is allowed within this rule.  It would look like
```dns
; static
www.example.com HTTPS IN 600 1 .
                             2 fallback.example.com.
; ECS-controlled
www.example.com AAAA ...
fallback.example.com AAAA ...
```

From the auth side, the per-QTYPE behavior here is only for authoritative servers that already have per-QTYPE behavior for SVCB (appending the Additional A/AAA records).  Arguably, this is not actually per-QTYPE behavior: the same logic applies to MX and SRV, if anyone wanted to use ECS with them, because they do the same trick of putting A/AAAA in Additional, and Additional can't use ECS.

enygren: For what it's worth, there are also plenty of places where ECS will get used today
on CNAME records that end up pointing to SVCB records.  For example
the following will be part of a very common deployment path that is widely used today:

```
www.foo.example. CNAME IN 600 foo.georedir.example.net.
foo.georedir.example.net CNAME IN 600 foo.cdn.example.
foo.cdn.example. HTTPS IN 600 ...
foo.cdn.example. A IN 600 ...
foo.cdn.example. AAAA IN 600 ...
```

In this case, "foo.georedir.example.net." may be doing multi-CDN splitting
based on client geo/asn and using ECS as part of this.  
As such, having a "SHOULD NOT" for ECS with SVCB seems likely
to be problematic since the authority here issuing the SVCB has no control
over whether a CNAME pointing to it is using ECS.

Just changing the A/AAAA case won't help when the protocols may need 
to vary (eg, "ISP ___ in the US breaks QUIC by randomly dropping UDP
and thus we don't want to offer it for clients in their ASN").

rs: > DNS steering with priority is allowed within this rule. It would look like
> 
> ```
> ; static
> www.example.com HTTPS IN 600 1 .
>                              2 fallback.example.com.
> ; ECS-controlled
> www.example.com AAAA ...
> fallback.example.com AAAA ...
> ```

That will requires two round trips from the resolver to the auth, assuming the resolver supports filling the iphints or worse, RTT from the client when resolver support is missing. We are losing a latency improvement opportunity here, where the auth could just provide a self contained response to minimize traffic and latency while doing steering. 

I think iphint from auth is especially important in the context of DNS steering as those A/AAAA have limited TTL and thus, those additional round trips won't be the exception.

To me the only reason for adding this clause is economical on the resolver side. Having a new popular qtype with ECS cache split can be expensive, especially for a huge resolver like quad8. If that's the motivation, it should be stated clearly, as I see no purely technical reason for this change.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +661,36 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+queries, to indicate that ECS is supported but is disabled for this query.
+If the resolver does not plan to perform this resolution, it SHOULD send the
+same ECS option that it would use for an A/AAAA query.
+
+According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
+Additional section\] MUST NOT be tied to a network".  Accordingly,
+resolvers SHOULD assume that any A/AAAA records in the Additional section
+are network-invariant.  When responding to a SVCB query with an ECS
+option (even one whose SOURCE PREFIX-LENGTH is zero), Authoritative servers
+must omit such records if they wish to serve subnet-specific addresses.
```
enygren, line 691:
```suggestion
MUST omit such records if they wish to serve subnet-specific addresses.
```

???

draft-ietf-dnsop-svcb-https.md
bemasc: @enygren
> For what it's worth, there are also plenty of places where ECS will get used today on CNAME records

I think ECS-variant CNAMEs should probably have been banned in the ECS spec, and this seems like a good opportunity to start phasing them out.  However, if we decide that they should be well-supported, then you're right that we need ECS on the SVCB queries.  I've drawn up #309 to show what I think that would look like.

> Just changing the A/AAAA case won't help when the protocols may need
> to vary (eg, "ISP ___ in the US breaks QUIC by randomly dropping UDP
> and thus we don't want to offer it for clients in their ASN").

Yuck.  You can't rely on ECS, so clients need to handle this anyway.  Any network of this kind is horribly broken, so we shouldn't be trying to put in performance optimizations for it.

@rs
> That will requires two round trips from the resolver to the auth

The number of roundtrips actually has nothing to do with whether the resolver is allowed to disable ECS.  I made a slight text update that I hope makes the reasoning clearer; you can also look at #309.

> To me the only reason for adding this clause is economical on the resolver side. Having a new popular qtype with ECS cache split can be expensive

Yes, the motivation is cache performance, but the resolver can operate with any cache size it likes, so it's not exactly an "expense".  Even with an infinite cache size, unnecessary use of ECS harms performance for users, by making cache hits less likely.

bemasc:
Thanks for the feedback, everyone.  There have been several requests for a simpler version that permits more creative use of ECS, so I've drawn up #309 as an alternative.  Please take a look.

This is a confusing topic.  Hopefully, #309 will help make it clearer that
1. The ECS-conditional behavior for authoritative servers is unavoidable because Additional records always have an implicit SCOPE PREFIX-LENGTH of zero.
2. The key question is whether full-service resolvers MUST, SHOULD, MAY, or SHOULD NOT send the same ECS on SVCB as A/AAAA.  This is essentially a tradeoff between cache efficiency (SHOULD NOT) and usage flexibility (SHOULD/MUST).

draft-ietf-dnsop-svcb-https.md
bemasc: "MUST if you wish" in all-caps seems strange to me.

draft-ietf-dnsop-svcb-https.md
puneetsood: > @enygren
> 
> > For what it's worth, there are also plenty of places where ECS will get used today on CNAME records
> 
> I think ECS-variant CNAMEs should probably have been banned in the ECS spec, and this seems like a good opportunity to start phasing them out. However, if we decide that they should be well-supported, then you're right that we need ECS on the SVCB queries. I've drawn up #309 to show what I think that would look like.

FWIW CNAME processing is mentioned in RFC 7871 (section 7.2.1). Quote:
   Since some queries can result in multiple RRsets being added to the
   response, there is an unfortunate ambiguity from the original
   specification as to how SCOPE PREFIX-LENGTH would apply to each
   individual RRset.  For example, multiple types in response to an ANY
   metaquery could all have different applicable SCOPE PREFIX-LENGTH
   values, but this protocol only has the ability to signal one.  The
   response SHOULD therefore, include the longest relevant PREFIX-LENGTH
   of any RRset in the answer, which could have the unfortunate side
   effect of redundantly caching some data that could be cached more
   broadly.  For the specific case of a Canonical Name (CNAME) chain,
   the Authoritative Nameserver SHOULD only place the initial CNAME
   record in the Answer section, to have it cached unambiguously and
   appropriately. Most modern Recursive Resolvers restart the query
   with the CNAME, so the remainder of the chain is typically ignored
   anyway.  For message-focused resolvers, rather than RRset-focused
   ones, this will mean caching the entire CNAME chain at the longest
   PREFIX-LENGTH of any RRset in the chain.

There is no mention of the scope to use for the CNAME response which would imply a non-zero scope is allowed. It does however recommend splitting the response across CNAME, requested RR if the prefix-length scope is different (especially if the CNAME scope is /0) for more efficient caching. It will be good to have reference to this section and a similar recommendation to split CNAME, SVCB alias, SVCB target RRs with different ECS scopes.

> 
> > Just changing the A/AAAA case won't help when the protocols may need
> > to vary (eg, "ISP ___ in the US breaks QUIC by randomly dropping UDP
> > and thus we don't want to offer it for clients in their ASN").
> 
> Yuck. You can't rely on ECS, so clients need to handle this anyway. Any network of this kind is horribly broken, so we shouldn't be trying to put in performance optimizations for it.
> 
> @rs
> 
> > That will requires two round trips from the resolver to the auth
> 
> The number of roundtrips actually has nothing to do with whether the resolver is allowed to disable ECS. I made a slight text update that I hope makes the reasoning clearer; you can also look at #309.
> 
> > To me the only reason for adding this clause is economical on the resolver side. Having a new popular qtype with ECS cache split can be expensive
> 
> Yes, the motivation is cache performance, but the resolver can operate with any cache size it likes, so it's not exactly an "expense". Even with an infinite cache size, unnecessary use of ECS harms performance for users, by making cache hits less likely.

puneetsood: Suggested wording to put the ECS RFC text first and the recommendations for auth, recursive after that. Keeping this here instead of in #309 since all the discussion is here.

```suggestion
SVCB records may contain IP addresses (in ipv*hint SvcParams),
or direct users to a TargetName. In these cases, a recursive
resolver and authoritative server can use ECS [expand] to request and
receive answers suitable for a particular client IP range.

According to Section 7.3.1 of RFC7871, "Any records from \[the
Additional section\] MUST NOT be tied to a network". In addition,
Section 7.2.1 states "For the specific case of a Canonical Name (CNAME) chain,
the Authoritative Nameserver SHOULD only place the initial CNAME
record in the Answer section, to have it cached unambiguously and
appropriately.". This is meant to allow caching different RRs at the
appropriate scope prefix-length by a resolver.

Accordingly, authoritative servers SHOULD not include additional RRs
in a response to an HTTPS query if their prefix-length scope is
different from the first RR in the response (which could be a CNAME or an SVCB record). A recursive resolver SHOULD include the same ECS option in SVCB queries as in A/AAAA queries.
```

bemasc:
It seems that we have consensus to support non-address variation with ECS (e.g. ECS-dependent CNAME), so I've merged #309, which simplifies this PR by removing the cache optimization.

draft-ietf-dnsop-svcb-https.md
bemasc: Thanks for the suggestion.  I've incorporated these references in https://github.com/MikeBishop/dns-alt-svc/pull/308/commits/9e0fe746a458fdd2bf7f63d0c357009efcfdb428.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +661,25 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+resolvers to request IP addresses that are suitable for a particular client
+IP range.  SVCB records may contain IP addresses (in ipv*hint SvcParams),
+or direct users to a subnet-specific TargetName, so recursive resolvers
+SHOULD include the same ECS option in SVCB queries as in A/AAAA queries.
+
+According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
+Additional section\] MUST NOT be tied to a network".  Accordingly,
+resolvers SHOULD treat any records in the Additional section as having
+SCOPE PREFIX-LENGTH zero, and MAY cache them on this basis.  Authoritative
+servers MUST omit such records if they are not suitable for global use.
```
bemasc, line 676: Here's an alternative interpretation of "MUST NOT be tied to a network".  This might allow more use of Additional records, resulting in lower latency.

@puneetsood FYI

```suggestion
SOURCE PREFIX-LENGTH zero and SCOPE PREFIX-LENGTH as specified
in the ECS option, and MAY cache them on this basis.  Authoritative
servers MUST omit such records if they are not suitable
for use by any stub resolvers that set SOURCE PREFIX-LENGTH to zero.
```

bemasc: On reflection, I'm leaning toward this interpretation, because it makes it more likely that we can revise RFC 7871 in the future to make these records ECS-enabled without creating a compatibility problem.

draft-ietf-dnsop-svcb-https.md
bemasc: I've reworked the text since this comment.  It now discusses this topic (line 679).

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +661,27 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+resolvers to request IP addresses that are suitable for a particular client
+IP range.  SVCB records may contain IP addresses (in ipv*hint SvcParams),
+or direct users to a subnet-specific TargetName, so recursive resolvers
+SHOULD include the same ECS option in SVCB queries as in A/AAAA queries.
+
+According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
+Additional section\] MUST NOT be tied to a network".  Accordingly,
+resolvers SHOULD treat any records in the Additional section as having
+SOURCE PREFIX-LENGTH zero and SCOPE PREFIX-LENGTH as specified
+in the ECS option, and MAY cache them on this basis.  Authoritative
```
puneetsood, line 676: https://tools.ietf.org/html/rfc2181#section-5.4.1 recommends not forwarding Additional section records to clients.

_**Unauthenticated RRs received and cached from the least trustworthy of
those groupings, that is data from the additional data section, and
data from the authority section of a non-authoritative answer, should
not be cached in such a way that they would ever be returned as
answers to a received query.  They may be returned as additional
information where appropriate.  Ignoring this would allow the
trustworthiness of relatively untrustworthy data to be increased
without cause or excuse.**_

Therefore the recommendation:
_**Accordingly,
resolvers SHOULD treat any records in the Additional section as having
SOURCE PREFIX-LENGTH zero and SCOPE PREFIX-LENGTH as specified
in the ECS option, and MAY cache them on this basis.**_

is unlikely to have an impact in the real world - popular DNS resolvers/implementations (e.g. Quad{189}, OpenDNS) do not propagate the ECS information to their clients (stubs mentioned here). This can be seen in the handling of MX records.

$ dig @ns1.google.com -t MX google.com +nostats
<snip>
;; ANSWER SECTION:
google.com.		600	IN	MX	20 alt1.aspmx.l.google.com.
google.com.		600	IN	MX	30 alt2.aspmx.l.google.com.
google.com.		600	IN	MX	50 alt4.aspmx.l.google.com.
google.com.		600	IN	MX	10 aspmx.l.google.com.
google.com.		600	IN	MX	40 alt3.aspmx.l.google.com.

;; ADDITIONAL SECTION:
alt1.aspmx.l.google.com. 293	IN	A	173.194.216.26
alt1.aspmx.l.google.com. 293	IN	AAAA	2607:f8b0:400c:c12::1a
alt2.aspmx.l.google.com. 293	IN	A	64.233.186.26
alt2.aspmx.l.google.com. 293	IN	AAAA	2800:3f0:4003:c00::1a
alt4.aspmx.l.google.com. 293	IN	A	172.253.120.26
alt4.aspmx.l.google.com. 293	IN	AAAA	2a00:1450:400c:c01::1a
aspmx.l.google.com.	293	IN	A	142.250.123.26
aspmx.l.google.com.	293	IN	AAAA	2607:f8b0:4023:1402::1a
alt3.aspmx.l.google.com. 293	IN	A	209.85.203.26
alt3.aspmx.l.google.com. 293	IN	AAAA	2a00:1450:400b:c03::1a

$ dig @9.9.9.9 -t MX google.com +nostats
<snip>
;; ANSWER SECTION:
google.com.		600	IN	MX	10 aspmx.l.google.com.
google.com.		600	IN	MX	20 alt1.aspmx.l.google.com.
google.com.		600	IN	MX	40 alt3.aspmx.l.google.com.
google.com.		600	IN	MX	50 alt4.aspmx.l.google.com.
google.com.		600	IN	MX	30 alt2.aspmx.l.google.com.

bemasc: Thanks for investigating that.  I didn't know about RFC 2181.

I think the text you quoted is (surprisingly?) in line with this draft.

> **Unauthenticated** RRs ... from the additional data section ... should
> not be cached in such a way that they would ever be returned as
> answers

SVCB Section 4.1:
> If the zone is signed, the server SHOULD also include positive or negative DNSSEC responses for these records in the Additional section.

So these Additional records could actually end up cached and used for general purposes.

RFC 2181
> They may be returned as additional information where appropriate.

SVCB Section 4.2:
> ... adding all useful RRs from the response to the Additional section

So even without authentication, we're using these records as recommended in RFC 2181.

Thanks for showing an easy way to check the real world behavior.
> ... popular DNS resolvers/implementations (e.g. Quad{189}, OpenDNS) do not propagate the ECS information to their clients

I think they do not propagate any Additional data to their clients.  That seems like a compliant, but perhaps not highly optimized behavior.  A highly optimized resolver could comply with RFC 2181 by maintaining a separate cache for records returned in the Additional section, to be used only when populating the Additional section of responses (when appropriate).  If that cache assumes SCOPE /0, only global records can go in the Additional section, whereas if it assumes SOURCE /0, authoritative servers can safely return ECS-tailored Additional records.

As a practical matter this may be moot in current implementations, but as a matter of requirements I think the text here still makes sense.

puneetsood: To provide context to a reader, will be be good to have a reference to https://tools.ietf.org/html/rfc2181#section-5.4.1 after the phrase "and MAY cache them on this basis".

puneetsood (approved):

draft-ietf-dnsop-svcb-https.md
bemasc: OK.  The caching question is mostly independent of ECS, so I moved it to the preceding section and added the full quote from RFC 2181.

enygren (approved):

draft-ietf-dnsop-svcb-https.md
```diff
@@ -659,6 +661,35 @@ each RRSet in the Additional section with the same DNSSEC-related records
...
+
+The EDNS Client Subnet option (ECS, {{!RFC7871}}) allows recursive
+resolvers to request IP addresses that are suitable for a particular client
+IP range.  SVCB records may contain IP addresses (in ipv*hint SvcParams),
+or direct users to a subnet-specific TargetName, so recursive resolvers
+SHOULD include the same ECS option in SVCB queries as in A/AAAA queries.
+
+According to Section 7.3.1 of {{!RFC7871}}, "Any records from \[the
+Additional section\] MUST NOT be tied to a network".  Accordingly,
+when processing a response whose QTYPE is SVCB-compatible,
```
bemasc, line 682: @enygren @MikeBishop I added this line to scope the normative requirement more explicitly.

bemasc:
Thanks everyone for your input and reviews.  It seems like we've found a broadly acceptable solution here, but please continue to provide feedback if there's something you'd like to change.

PR #312, opened 2021-04-07
Test vectors with failure cases

wtoorop:
This includes Pieter Lexis' test vectors and has some additional failure cases added.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1899,18 +1899,57 @@ backslash in two presentation formats.
 
 ## Failure cases
 
-In this subsection example resource records are shown which are not compliant
-with this document for various reasons explained with each example.
+In this subsection, example resource records are shown which are not
+compliant with this document. The various reasons for non-compliance
+are explained with each example.
 
-### Mandatory key missing
 
-    example.com.   SVCB   1 foo.example.com. mandatory=key123 key456=abc
+This example has mutilple instances of the same SvcParamKey.
 
-### Multiple instances of the same mandatory key
+    example.com.   SVCB   1 foo.example.com. (
+                           key123=abc key123=abc
+                           )
+
+The next example the SvcParams are missing the comma-separated values.
+
+    example.com.   SVCB   1 foo.example.com. (
+                           alpn port ipv4hint echconfig ipv6hint
+                           )
+
+In this record a mandatory SvcParam is missing.
+
+    example.com.   SVCB   1 foo.example.com. mandatory=key123
+
+The "mandatory" SvcParamKey should not be included in mandatory
+list ({{mandatory}}).
+
+    example.com.   SVCB   1 foo.example.com. mandatory=mandatory
+
+Here there are multiple instances of the same SvcParamKey in
+the mandatory list ({{mandatory}}).
 
     example.com.   SVCB   1 foo.example.com. ( 
                           mandatory=key123,key123 key123=abc
                           )
+
+With HTTPS records the SvcParamKeys "port" and "no-default-alpn" are
+automatically mandatory and therefore are required in the
+record ({{value-list}}).
+
+    example.com.   HTTPS   1 foo.example.com. (
+                           mandatory=key123 key123=123
+                           )
+
+
+In the last example, the automatically mandatory SvcParamKeys 
+"port" and "no-default-alpn" should not appear in the 
+mandatory list ({{mandatory}}).
+
+    example.com.   HTTPS   1 foo.example.com. (
+                           mandatory=key123,no-default-alpn,port
+                           key123=123 port=1 no-default-alpn
+                           )
+
 
 # Change history
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1911,11 +1911,15 @@ SvcParamKey {{presentation}}.
                            key123=abc key123=def
                            )
 
-The next example the SvcParams are missing their values.
+In the next examples the SvcParamKeys are missing their values.
 
-    example.com.   SVCB   1 foo.example.com. (
-                           alpn port ipv4hint echconfig ipv6hint
-                           )
+    example.com.   SVCB   1 foo.example.com. mandatory
+    example.com.   SVCB   1 foo.example.com. alpn
+    example.com.   SVCB   1 foo.example.com. port
+    example.com.   SVCB   1 foo.example.com. ipv4hint
+    example.com.   SVCB   1 foo.example.com. echconfig
+    example.com.   SVCB   1 foo.example.com. ipv6hint
+    example.com.   SVCB   1 foo.example.com. keyXXX
 
 The "no-default-alpn" SvcParamKey value MUST be empty ({{alpn-key}}).
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1904,7 +1904,7 @@ compliant with this document. The various reasons for non-compliance
 are explained with each example.
 
 
-This example has mutilple instances of the same
+This example has multiple instances of the same
 SvcParamKey {{presentation}}.
 
     example.com.   SVCB   1 foo.example.com. (
@@ -1917,7 +1917,7 @@ The next example the SvcParams are missing their values.
                            alpn port ipv4hint echconfig ipv6hint
                            )
 
-The "no-default-alpn" SvcParamKey value must be empty ({{alpn-key}}).
+The "no-default-alpn" SvcParamKey value MUST be empty ({{alpn-key}}).
 
     example.com.   SVCB   1 foo.example.com. no-default-alpn=abc
 
@@ -1925,7 +1925,7 @@ In this record a mandatory SvcParam is missing ({{mandatory}}).
 
     example.com.   SVCB   1 foo.example.com. mandatory=key123
 
-The "mandatory" SvcParamKey should not be included in mandatory
+The "mandatory" SvcParamKey MUST not be included in mandatory
 list ({{mandatory}}).
 
     example.com.   SVCB   1 foo.example.com. mandatory=mandatory
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1710,7 +1710,7 @@ long, it is broken into several lines.
 
 ## AliasForm
 
-    0 foo.example.com
+    0 foo.example.com.
 
     \# 19 (
     00 00                                              ; priority
@@ -1735,7 +1735,7 @@ The first form is the simple "use the ownername".
 
 This vector only has a port.
 
-    16 foo.example.com port=53
+    16 foo.example.com. port=53
 
     \# 25 (
     00 10                                              ; priority
@@ -1753,7 +1753,7 @@ This vector only has a port.
 
 This example has a key that is not registered, its value is unquoted.
 
-    1 foo.example.com key667=hello
+    1 foo.example.com. key667=hello
 
     \# 28 (
     00 01                                              ; priority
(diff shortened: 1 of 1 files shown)


wtoorop:
More commits with more failure cases (from @TCY16) will follow shortly...

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,259 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+Here there are multiple instances of the same SvcParamKey in
+the mandatory list ({{mandatory}}).
+
+    example.com.   SVCB   1 foo.example.com. ( 
+                          mandatory=key123,key123 key123=abc
+                          )
+
+With HTTPS records the SvcParamKeys "port" and "no-default-alpn" are
+automatically mandatory and therefore are required in the
+record ({{value-list}}).
```
bemasc, line 1942: No, this record is perfectly valid.  "automatically mandatory" means "if these keys are present, clients must act as if they were in the `mandatory=...` list".

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,259 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+
+    example.com.   HTTPS   1 foo.example.com. (
+                           mandatory=key123 key123=123
+                           )
+
+
+In the last example, the automatically mandatory SvcParamKeys 
+"port" and "no-default-alpn" should not appear in the
```
bemasc, line 1950: As your text notes, this is a "should not", so the record is not invalid.

draft-ietf-dnsop-svcb-https.md
TCY16: Removed

draft-ietf-dnsop-svcb-https.md
TCY16: and removed, thanks for the feedback Ben!

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,240 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+
+
+## Failure cases
+
+In this subsection, example resource records are shown which are not
+compliant with this document. The various reasons for non-compliance
+are explained with each example.
+
+
+This example has mutilple instances of the same
```
wtoorop, line 1907:
```suggestion
This example has multiple instances of the same
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,240 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+                           key123=abc key123=def
+                           )
+
+The next example the SvcParams are missing their values.
+
+    example.com.   SVCB   1 foo.example.com. (
+                           alpn port ipv4hint echconfig ipv6hint
+                           )
+
+The "no-default-alpn" SvcParamKey value must be empty ({{alpn-key}}).
```
wtoorop, line 1920:
```suggestion
The "no-default-alpn" SvcParamKey value MUST be empty ({{alpn-key}}).
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,240 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+
+    example.com.   SVCB   1 foo.example.com. no-default-alpn=abc
+
+In this record a mandatory SvcParam is missing ({{mandatory}}).
+
+    example.com.   SVCB   1 foo.example.com. mandatory=key123
+
+The "mandatory" SvcParamKey should not be included in mandatory
```
wtoorop, line 1928:
```suggestion
The "mandatory" SvcParamKey MUST not be included in mandatory
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,240 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+SvcParamKey {{presentation}}.
+
+    example.com.   SVCB   1 foo.example.com. (
+                           key123=abc key123=def
+                           )
+
+The next example the SvcParams are missing their values.
+
+    example.com.   SVCB   1 foo.example.com. (
+                           alpn port ipv4hint echconfig ipv6hint
```
marka63, line 1917: These should be done in individual records rather than 1 record to ensure proper error detection.

marka63, line 1917: Add keyXXX forms as well.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,240 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    \x20\x01\x0d\xb8\x00\x00\x00\x00
+         \x00\x00\x00\x00\x00\x00\x00\x01              # first address
+    \x20\x01\x0d\xb8\x00\x00\x00\x00
+         \x00\x00\x00\x00\x00\x53\x00\x01              # second address
+
+
+This example shows a single IPv6 hint in IPv4 mapped IPv6 presentation
+format.
+
+    1 example.com ipv6hint="2001:db8:ffff:ffff:ffff:ffff:198.51.100.100"
```
marka63, line 1817: This may need to done in multi-line format to allow it to fit nicely into .txt format.

draft-ietf-dnsop-svcb-https.md
marka63, line 1917: add mandatory to the list.

bemasc:
Do we have an example of a published RFC with such a detailed catalog of invalid test vectors?  If not, perhaps it's inappropriate for a standards document.  After all, we are not trying to ship a complete test suite.

wtoorop:
Hi @bemasc , I can't think of a RFC from head, but I know that two ZONEMD implementations completely missed that they should not try and *fail* verification with two ZONEMD RR's with the same algorithm. A test-vector for the failure case would probably have prevented that, as implementations tend to copy and use those in their test suits. What would be a good reason to not include them? Do you want me to ask the workgroup again about whether they think it should be included or not?

Also, @pieterlexis' test-vectors already payed off! Because both the ldns implementation and the Net::DNS implementation missed that alpn values have to be escaped and unescaped twice!

draft-ietf-dnsop-svcb-https.md
TCY16: resolved

TCY16: resolved

TCY16: resolved

draft-ietf-dnsop-svcb-https.md
TCY16: This _should_ fit the .txt file as it is lined out to be on the exact maximum

TCY16:
Hi @bemasc,

I do get what you’re saying about the complete test suite and agree that some are slightly triviaal, though I think there is still added value of including at least some of the tests. 
I started this list mainly for the NSD test suite, but as both you and @marka64 pointed out, there were some mistakes/improvements. By adding a subset of these tests me and @wtoorop hope that other implementers are less prone to incorrect implementations.

Do you disagree?

moonshiner:
I agree with @TCY16 - we should more inclusive with the examples.  They seem to trivial to us, but we are not the people whom will reading this and attempting to updating their DNS zone data.   Trivial examples are the turtles folks will use to reach the next set of turtles. 

(speaking only for myself)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,244 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+
+In the next examples the SvcParamKeys are missing their values.
+
+    example.com.   SVCB   1 foo.example.com. mandatory
+    example.com.   SVCB   1 foo.example.com. alpn
+    example.com.   SVCB   1 foo.example.com. port
+    example.com.   SVCB   1 foo.example.com. ipv4hint
+    example.com.   SVCB   1 foo.example.com. echconfig
+    example.com.   SVCB   1 foo.example.com. ipv6hint
+    example.com.   SVCB   1 foo.example.com. keyXXX
```
wtoorop, line 1922:
```suggestion
    example.com.   SVCB   1 foo.example.com. key123
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,244 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+This example has multiple instances of the same
+SvcParamKey {{presentation}}.
+
+    example.com.   SVCB   1 foo.example.com. (
+                           key123=abc key123=def
+                           )
+
+In the next examples the SvcParamKeys are missing their values.
+
+    example.com.   SVCB   1 foo.example.com. mandatory
```
pieterlexis, line 1916: Another failure case would be:

```
     example.com.   SVCB   1 foo.example.com. mandatory=
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,244 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    example.com.   SVCB   1 foo.example.com. alpn
+    example.com.   SVCB   1 foo.example.com. port
+    example.com.   SVCB   1 foo.example.com. ipv4hint
+    example.com.   SVCB   1 foo.example.com. echconfig
```
bemasc, line 1920: This is a tricky one.  The param indicates a valid SVCB `echconfig` parameter whose value is "" (which is what you get when you base64-decode "").

Of course, "" is currently not a valid ECHConfigList, and is unlikely to be, but the ECH specification is not final.  In any event, I don't think there's an obligation for zone file parsers to check whether the ECHConfigList is valid, which may be difficult if they're not linked to an ECH-capable TLS library.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1702,6 +1702,244 @@ the authoritative.  Thus, splitting the RR type would double, or in
...
+    example.com.   SVCB   1 foo.example.com. ipv6hint
+    example.com.   SVCB   1 foo.example.com. key123
```
bemasc, line 1922: This is definitely allowed.  `key123` could someday be registered as a key that uses an empty value.  Until it's registered, zone files can't apply any limits.

bemasc (commented):
OK, it sounds like there is a lot of interest in including invalid RDATA.

draft-ietf-dnsop-svcb-https.md
pieterlexis: I know an earlier draft mentioned that empty generic keys were not allowed and only no-default-alpn was defined as having no value?

bemasc: I don't recall a limit like that.  Perhaps an early draft implied that this should be spelled `key123=""`, but there is no such requirement now.

draft-ietf-dnsop-svcb-https.md
wtoorop, line 1920:
```suggestion
```
Allow for empty echconfig lists. Because ECH spec is not final.

wtoorop (commented):
Suggestions based on review comments from @bemasc

draft-ietf-dnsop-svcb-https.md
wtoorop: Removed that one.

draft-ietf-dnsop-svcb-https.md
wtoorop: Removed this line too.

bemasc (approved):
Waiting for coauthor approval.

Habbie:
> Do we have an example of a published RFC with such a detailed catalog of invalid test vectors? If not, perhaps it's inappropriate for a standards document. After all, we are not trying to ship a complete test suite.

The few positive examples in CSYNC really helped me in implementing the type, and I wouldn't have minded some 'negative' examples in there too.

In general, I feel RFCs generally have too few test vectors (often zero), not too many.

I think shipping a complete test suite would be an excellent way to lead by example!

enygren:
Opened issue #315  to test the test vectors.  It would be great if people could confirm there if they've tested the test vectors.

PR #313, opened 2021-04-08
Relax requirements for resolvers and SvcParams

bemasc:
It's probably better for everyone if resolvers don't inspect the
SvcParams, but it's not really a requirement, and it has been a
controversial point due to implementation details and RDATA-based
filtering.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -649,10 +649,11 @@ See {{incomplete-response}} for possible optimizations of this procedure.
 
 ## General requirements
 
-Recursive resolvers MAY treat the SvcParams portion of the SVCB RR
-as opaque.  No part of this specification requires recursive resolvers
-to alter their behavior based on its contents, even if the contents are
-invalid.
+Recursive resolvers MUST be able to convey SVCB records with unrecognized
+SvcParamKeys or malformed SvcParamValues.  Resolvers MAY treat the entire
+SvcParams portion of the record as opaque.  No part of this specification requires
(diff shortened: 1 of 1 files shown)


vcunat (approved):

Habbie (approved):

enygren:
This seems like entering into risky territory.  A big part of the current SHOULD NOT text is that there is much more ossification risk if recursive resolvers start looking at parameters.  It's critical that we be able to add new parameters without recursive resolvers looking at them.  (SHOULD NOT does not prevent them from doing so if they have a good reason to do so.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -649,9 +649,10 @@ See {{incomplete-response}} for possible optimizations of this procedure.
 
 ## General requirements
 
-Recursive resolvers SHOULD treat the SvcParams portion of the SVCB RR
-as opaque and SHOULD NOT try to alter their behavior based
-on its contents.
+Recursive resolvers MAY treat the SvcParams portion of the SVCB RR
+as opaque.  No part of this specification requires recursive resolvers
+to alter their behavior based on its contents, even if the contents are
+invalid.
```
bemasc, line 655:
```suggestion
Recursive resolvers MUST be able to convey SVCB records with unrecognized
SvcParamKeys or malformed SvcParamValues.  Resolvers MAY treat the entire
SvcParams portion of the record as opaque.  No part of this specification requires
recursive resolvers to alter their behavior based on its contents, even if the contents
are invalid.
```

PR #314, opened 2021-04-09
Relax recommendation to have a default set

bemasc:
This change trades normative language for more explanation about how to
choose an appropriate default set for a scheme.


bemasc:
See https://github.com/bemasc/svcb-dns/pull/2

Habbie (approved):

enygren:
Given that this is guidance for specification-writers this makes sense.

PR #316, opened 2021-04-16
Move ECH codepoint

bemasc:
This simplifies the compatibility questions related to changing the name
of "echconfig" to "ech".


bemasc:
It looks like we don't have consensus for this change.

PR #319, opened 2021-04-21
Adjust requirements for invalid SvcParamValues

bemasc:
This change relaxes the requirement not to validate SvcParamValues, while encouraging resolvers to distinguish lightweight validation from deep parsing of complex value types.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -655,9 +655,10 @@ Recursive resolvers MUST be able to convey SVCB records with unrecognized
 SvcParamKeys, and MAY treat the entire
 SvcParams portion of the record as opaque.  No part of this specification requires
 recursive resolvers to alter their behavior based on its contents, even if the contents
-are invalid.  To ensure compatibility with complex SvcParam specifications,
-recursive resolvers MAY validate the values of recognized SvcParamKeys, but
-MUST NOT reject the record on this basis unless a value is obviously invalid.
+are invalid.  Recursive resolvers MAY validate the values of recognized
+SvcParamKeys and reject records containing invalid values.  However, for
+complex value types whose interpretation might differ between implementations
+(e.g. URIs), resolvers SHOULD limit validation to basic sanity checks.
 
 When responding to a query that includes the DNSSEC OK bit ({{!RFC3225}}),
 DNSSEC-capable recursive and authoritative DNS servers MUST accompany
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -657,7 +657,7 @@ SvcParams portion of the record as opaque.  No part of this specification requir
 recursive resolvers to alter their behavior based on its contents, even if the contents
 are invalid.  Recursive resolvers MAY validate the values of recognized
 SvcParamKeys and reject records containing values 
-which are invalid according to the SvsParam specification.
+which are invalid according to the SvcParam specification.
 For complex value types whose interpretation might differ 
 between implementations or have additional future
 allowed values added (e.g. URIs or "alpn"), resolvers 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -652,10 +652,12 @@ See {{incomplete-response}} for possible optimizations of this procedure.
 ## General requirements
 
 Recursive resolvers MUST be able to convey SVCB records with unrecognized
-SvcParamKeys or malformed SvcParamValues.  Resolvers MAY treat the entire
+SvcParamKeys, and MAY treat the entire
 SvcParams portion of the record as opaque.  No part of this specification requires
 recursive resolvers to alter their behavior based on its contents, even if the contents
-are invalid.
+are invalid.  To ensure compatibility with complex SvcParam specifications,
```
enygren, line 658:
```suggestion
are invalid.  To ensure compatibility with complex and/or future SvcParam specifications,
```

bemasc: Why isn't this covered by the preceding requirement to pass unrecognized keys?

enygren: I guess it mostly is, but if we're adding justification does it hurt to remind people over and over?

bemasc: I don't think it's relevant to this sentence, unless you're imagining changes that formally alter the range of allowed values for an existing key.

enygren: An example might be future alpn values.  Perhaps there's a better way to work?

bemasc: "evolving"?

bwelling: This spec identifies ALPN IDs as "a sequence of 1-255 octets", and in no way implies that anyone should restrict validation to currently defined identifiers.  If some future ALPN ID is more than 255 octets, there are larger problems, in that the existing parameter format wouldn't be able to encode them.

davidben: (TLS uses a 1-byte length prefix, so it's limited to 255 octets there too.)

bemasc:
@enygren I've reworded this based on feedback from @paulehoffman.  Please take another look.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -652,10 +652,13 @@ See {{incomplete-response}} for possible optimizations of this procedure.
...
 Recursive resolvers MUST be able to convey SVCB records with unrecognized
-SvcParamKeys or malformed SvcParamValues.  Resolvers MAY treat the entire
+SvcParamKeys, and MAY treat the entire
 SvcParams portion of the record as opaque.  No part of this specification requires
 recursive resolvers to alter their behavior based on its contents, even if the contents
-are invalid.
+are invalid.  Recursive resolvers MAY validate the values of recognized
+SvcParamKeys and reject records containing invalid values.  However, for
+complex value types whose interpretation might differ between implementations
+(e.g. URIs), resolvers SHOULD limit validation to basic sanity checks.
```
enygren, line 661: We should use something more descriptive/specific here.  
This wording is confusing.  (The term "sanity checks"
is also one that should avoid using from an inclusive
language perspective, if better and more descriptive terms
are available.)

Perhaps:

```suggestion
SvcParamKeys and reject records containing values 
which are invalid according to the SvsParam specification.
For complex value types whose interpretation might differ 
between implementations or have additional future
allowed values added (e.g. URIs or "alpn"), resolvers 
SHOULD limit validation to specified constraints.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -652,10 +652,16 @@ See {{incomplete-response}} for possible optimizations of this procedure.
...
 
 Recursive resolvers MUST be able to convey SVCB records with unrecognized
-SvcParamKeys or malformed SvcParamValues.  Resolvers MAY treat the entire
+SvcParamKeys, and MAY treat the entire
 SvcParams portion of the record as opaque.  No part of this specification requires
 recursive resolvers to alter their behavior based on its contents, even if the contents
-are invalid.
+are invalid.  Recursive resolvers MAY validate the values of recognized
+SvcParamKeys and reject records containing values 
+which are invalid according to the SvsParam specification.
```
enygren, line 660:
```suggestion
which are invalid according to the SvcParam specification.
```

enygren (approved):
Good other than typo

PR #320, opened 2021-04-30
Use the correct prefix for v4 mapped v6

pieterlexis:
RFC 3513 section 2.5.5 tells us IPv4-mapped IPv6 lives in the ::ffff
prefix.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1849,25 +1849,25 @@ Here, two IPv6 hints are quoted in the presentation format.
          \x00\x00\x00\x00\x00\x53\x00\x01              # second address
 
 
-This example shows a single IPv6 hint in IPv4 mapped IPv6 presentation
-format.
+This example shows a single IPv6 hint in IPv4-mapped IPv6 presentation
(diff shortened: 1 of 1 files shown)


bemasc (approved):

enygren:
What is the use-case for having a v4 mapped v6 address in a ipv6hint?  This seems like an invalid use-case and something we should prohibit?  Using v4 mapped v6 addresses in AAAA records used to be a way to break sites and is usually considered broken.

bemasc:
It seems like using a v4-mapped address in ipv6hint would break the site if served on the public internet, but could semi-plausibly be permissible in a closed network environment.

This is theoretically separate from the question of whether the syntax is allowed.

I'm happy to revert this change if you think it's less confusing to use the documentation prefix in this example.  We would also need to change the text not to say "IPv4-mapped", since it would merely (weirdly) be using the "decimal notation" syntax for a generic IPv6 address.

pieterlexis:
Do note that converting the documentation format mapped addresses `2001:db8:ffff:[..]:192.0.2.1` to wire format and calling `inet_ntop` will not yield the mapped address but a 'true' v6 address in presentation format.

PR #322, opened 2021-05-07
Test Vector presentation format full dns records

moonshiner:
Updated the Test Vectors in the appendix so they appear as DNS records, similar to the failure cases below them.


moonshiner:
ARRGH thought I caught them all.  thanks

PR #323, opened 2021-05-11
Document the rationale for Params in the RDATA

bemasc:


enygren (commented):
Do we want this in an appendix?  It seems distracting up-top.  (Otherwise looks fine?)

bemasc:
I proposed this language in response to a conversation on the mailing list, but it looks like that conversation ended up going in a different direction.  Maybe we don't need this at all.

I've changed the status to "draft".  I'm happy to put it in an appendix if anyone thinks that would help.

PR #325, opened 2021-05-20
Make comma escaping support optional if it's not needed

bemasc:
This change makes support for comma escaping in value lists mandatory
only after an ALPN ID is registered that contains a special character.
This gives zone file implementers the option to avoid implementation of
the value-list escaping now, and perhaps indefinitely.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -786,7 +786,9 @@ ALPNs are identified by their registered "Identification Sequence"
     alpn-id = 1*255OCTET
 
 The presentation `value` SHALL be a comma-separated list ({{value-list}})
-of one or more `alpn-id`s.
+of one or more `alpn-id`s.  So long as there are no registered protocol
+identifiers containing "," or "\\", zone file implementations MAY disallow
+these characters instead of implementing the `value-list` escaping procedure.
 
 The wire format value for "alpn" consists of at least one
 `alpn-id` prefixed by its length as a single octet, and these length-value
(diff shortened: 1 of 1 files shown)


enygren (commented):
Mark points out (and I was going to make the same comment) that we should also mention that such implementations may choose to display such entries as key1="..." instead of alpn="...".

Rather than referencing the ALPN registry, we could just have the MAY and say that in this case it is still possible to use key1="..." with these implementations for any ALPN values needing special characters.

bemasc:
OK, I've removed the conditional related to the IANA registry and explained the workaround.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1625,8 +1628,9 @@ although the output length in this document is not limited to 255 octets.
 ## Decoding a comma-separated list {#value-list}
 
 In order to represent lists of items in zone files, this specification uses
-comma-separated lists.  When "," is not escaped (by a preceding "\\"), it
-separates items in the list.  (For simplicity, empty items are not allowed.)
+comma-separated lists.  When the allowed items in the list cannot contain ","
+or "\\", this is trivial.  (For simplicity, empty items are not allowed.)
```
enygren, line 1632:
```suggestion
or "\\", this is trivial.  If items in the list are allowed to contain ",",
it can be specified by escaping as "\\,"  or by relying on the 
relying on the opaque key format (e.g. `key1=\002h2`) for 
implementations that do not support escaping.
(For simplicity, empty items are not allowed.)
```

It seems worth including an example here for here for potential zone file authors, not just implementors?

enygren (changes requested):
Seems fine, but we may want to include this example?

draft-ietf-dnsop-svcb-https.md
bemasc: I think it's OK to mention the opaque key format here (although I doubt many zone file authors will be reading the appendix), but I would prefer not to duplicate the example.

bemasc: I expanded the text here in a different way.  Please review.

enygren (approved):

Issue #326, opened 2021-05-20
ECH and AltSvc


martinthomson:
I find the text in the current draft a little too light when it comes to ECH and AltSvc.  I find that I can't work out exactly what to do based on the text that is present.

Each is fine in isolation:

* Section 9.1 makes it clear that when a client supports ECH it makes SVC queries always.  If it *requires* ECH then it will insist on receiving an ECH config.  In either case, if there is an ECH config, not getting ECH results in hard failure.

* Section 8.3 makes it clear that for AltSvc, you just use A/AAAA to find the server.  But that means that you won't have an ECH config.

The only text about the combination is this text in Section 8.3:

> Clients that implement support for ECH MUST perform the HTTPS RR query first, and MUST only make use of Alt-Svc when operating in SVCB-optional mode (see Section 9.1).

So, what name is queried for?  I *think* that this refers to the origin.  That is, the client needs to confirm whether the origin has an ECH config in SVCB.  If the ECH config exists, then the client needs to use it (and so it can't use AltSvc).  If the ECH config doesn't exist, then it is free to use AltSvc.

This is a valid conclusion, but I haven't seen any real discussion on this point (to be fair, I could have missed it, the volume of messages on this draft means that I've not read all of them).

I can see why you would prohibit the use of AltSvc if ECH is enabled for the origin.  AltSvc requires that the SNI indicate the origin in the TLS handshake, the origin has indicated (using SVCB) that it wants that information protected, and no means has been provided to protect that information.  I can see other options:

1. We might allow the origin to indicate that its ECH config applies to alternative services.  That could be a sign that the ECH config at the origin equally applies to all alternatives.  This would be a good outcome in my opinion, but I would need to walk through the implications for distributed deployments.  I think that it's OK because you only get AltSvc after successfully talking to a server and presumably one that has used the ECH config successfully.

2. We might allow the alternative service to offer its own ECH config and so offer that up through SVCB.  This might need some sort of indication from the origin (as part of AltSvc perhaps, as part of SVCB, or as part of its own ECH config) to delegate this.  In this case, the client requests SVCB for the alternative.  I have some reservations about taking this path; I haven't considered all the implications fully and it seems like there might be a lot of them.

Whatever the outcome is here, I'm uncomfortable with the level of treatment of this problem in the draft.  It needs more text and - ideally - more debate.

[...]

PR #327, opened 2021-05-21
Expand explanation of Alt-Svc/ECH interaction

bemasc:
Addresses #326

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1082,22 +1082,28 @@ single-client granularity.
 
 ## Interaction with Alt-Svc
 
-Clients that do not implement support for Encrypted ClientHello MAY
-skip the HTTPS RR query
+This specification does not alter the DNS records used when connecting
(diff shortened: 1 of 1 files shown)


martinthomson (changes requested):
This doesn't really say all it needs to say.  Well, at least it is a pretty user-unfriendly way of presenting the information.

To start with, you have to go off and read the other section to understand that SVCB-optional mode is something that is implicitly disabled when an ECH-enabled client encounters an ECH attribute.  You also have to appreciate understand that Alt-Svc operates by including the origin name in SNI and not the alternative service name.  Then you need to connect the dots from there to realize that Alt-Svc as specified in RFC 7838 is incompatible with ECH.

It would also be nice to lay out how server operators deal with this mess.  As it stands, server operators currently need to choose between the two with the understanding that ECH-enabled clients might not follow Alt-Svc, that clients without ECH might follow Alt-Svc, AND that all of this might change if Alt-Svc is ever revised (we can't promise that it will change, though we really need to - at least in my opinion).

bemasc:
Converted to draft as #329 seems to have more support.

Issue #252, opened 2020-09-30
Self-target


squarooticus:
§ 2.5.2 says:

```
For ServiceMode SVCB RRs, if TargetName has the value ".", then the
owner name of this record MUST be used as the effective TargetName.
```

This makes sense for HTTPS and any future child RRTYPES that eschew Attrleafs, but it doesn't make a lot of sense for owner names with Attrleafs. Wouldn't it make sense for `.` to refer to the bare domain with the Attrleafs stripped?

bemasc:
I think this is a plausible improvement.  I believe @marka63 previously suggested something similar.

However, I think it may already be too late for this change.  iOS 14 has shipped an implementation of draft-01, and this change would break compatibility with those devices.

squarooticus:
What would be the downside? Fallback to authority? That seems like an acceptable outcome from shipping a draft implementation without a clean escape hatch.

bemasc:
True, it's possible that this would be handled transparently via fallback.  ECH will remove the fallback, but (luckily?) ECH isn't defined yet.

marka63:
This will only become a issue for clients that are not aware of this expected behaviour and the operator uses "." for the target name.

Apple has implicitly committed to shipping a fix for this if it is changed by releasing code against a draft.  Additionally iOS users are relatively quick to apply updates.  I would not be worried about making a change like this.

enygren:
Per discussion, we're proposing to close this as a wontfix.  It is a nice feature but adds complexity and corner cases and is not necessary.  (There was a little WG support, but only a little.)

PR #328, opened 2021-05-23
Remove underscore prefix for SVCB TargetName resolving

Habbie:
Hello authors,

I'm putting this in as a PR because I suspect -not- having these words was an unintentional oversight; I talked to a few people familiar with this draft (and the svcb-dns) one and several of them thought I was wrong and the words are just missing.

There's one FIXME in this PR; there might be other places where a few words should also be added.

If you tell me 'this makes sense' I will clean up this PR. Otherwise, if you prefer I take this to dnsop, I'll do that.

Thanks!

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -480,7 +480,8 @@ without the use of SVCB.
 ### ServiceMode
 
 For ServiceMode SVCB RRs, if TargetName has the value ".", then the
-owner name of this record MUST be used as the effective TargetName.
+owner name of this record, with a possible underscore prefix removed,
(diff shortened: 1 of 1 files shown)


bemasc:
See #252 by @squarooticus, which was discussed in [this WG thread](https://mailarchive.ietf.org/arch/msg/dnsop/K6uyh2B5uFpNgNcgLc8PtOK9uEg/).  At the time, the authors concluded that there was insufficient support for this change in the working group.

Habbie:
Thanks, managed to miss that in my searches. In that case, the draft needs to clarify that underscore prefixes are -not- stripped. I might PR that later.

PR #329, opened 2021-05-26
Alternative Alt-Svc+HTTPS-RR integration

bemasc:
This version places the burden of consistency on domain owners instead
of client implementors.

Addresses #326.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1085,7 +1085,7 @@ single-client granularity.
 Clients that implement support for both Alt-Svc and HTTPS records SHOULD
 retrieve any HTTPS records for the Alt-Svc alt-authority, and ensure that
 their connection attempts are consistent with both the Alt-Svc parameters
-and any received HTTPS SvcParams.  If present, the HTTPS record's target
+and any received HTTPS SvcParams.  If present, the HTTPS record's TargetName
 and port override the alt-authority.  For example, suppose that
 "https://example.com" sends an Alt-Svc field value of:
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1103,7 +1103,7 @@ The client would retrieve the following HTTPS records:
 Based on these inputs, the following connection attempts would always be
 allowed:
 
-* HTTPS over TLS to `alt.example:443` (Consistent with both Alt-Svc and
+* HTTPS over TCP to `alt.example:443` (Consistent with both Alt-Svc and
   its HTTPS record)
 * HTTP/3 to `alt3.example:9443` (Consistent with both Alt-Svc and its HTTPS
   record)
@@ -1116,13 +1116,13 @@ The following connection attempts would not be allowed:
 * HTTP/3 to `alt.example:443` (not consistent with Alt-Svc)
 * Any connection to `alt2b.example` (no ALPN consistent with both the HTTPS
   record and Alt-Svc)
-* HTTPS over TLS to any port on `alt3.example` (not consistent with Alt-Svc)
+* HTTPS over TCP to any port on `alt3.example` (not consistent with Alt-Svc)
 
 The following connection attempts would be allowed only if the client does
 not support ECH, as they rely on SVCB-optional fallback behavior that is
 disabled when the "ech" SvcParam is present ({{ech-client-behavior}}):
 
-* HTTPS over TLS to `alt2.example:443` (Alt-Svc only)
+* HTTPS over TCP to `alt2.example:443` (Alt-Svc only)
 * HTTP/3 to `example.com:8443` (Alt-Svc only)
 
 Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1082,22 +1082,28 @@ single-client granularity.
 
 ## Interaction with Alt-Svc
 
-Clients that do not implement support for Encrypted ClientHello MAY
-skip the HTTPS RR query
+This specification does not alter the DNS records used when connecting
+to an Alt-Svc endpoint (typically A and/or AAAA only) or the data contained
+in Alt-Svc field values.  Origins can publish HTTPS records without any
+need to modify their existing Alt-Svc arrangements.  Future specifications
+may enable more integration of Alt-Svc and HTTPS records, such as querying
+HTTPS records based on an Alt-Svc field value.
+
+In general, clients MAY skip the HTTPS RR query
 if a usable Alt-Svc value is available in the local cache.
 If Alt-Svc connection fails, these clients SHOULD fall back to the HTTPS RR
 client connection procedure ({{client-behavior}}).
 
-Clients that implement support for ECH MUST perform the HTTPS RR query first,
-and MUST only make use of Alt-Svc when operating in SVCB-optional mode (see
-{{ech-client-behavior}}).  This rule allows origins to send Alt-Svc
-advertisements (which do not currently support ECH) without compromising the
-privacy assurances of ECH configurations they have published in the DNS.
-
-This specification does not alter the DNS records used when connecting
-to an Alt-Svc hostname (typically A and/or AAAA only) or the data contained
-in Alt-Svc field values.  Future specifications may enable more use of
-Alt-Svc with HTTPS RR queries and/or ECH.
+Clients that support ECH are subject to additional requirements.  If all HTTPS
+records for the origin have the "ech" SvcParamKey, ECH-capable clients MUST
+use ECH on all connections for this origin (as mandated in
+{{ech-client-behavior}}).  To comply with this requirement, clients will
+typically have to retrieve any HTTPS records for the origin before making use
+of any cached Alt-Svc field values (which currently cannot provide an
+ECHConfigList), and ignore any Alt-Svc field value whose use would require
+disabling ECH.  (The "ech" SvcParamValue is only for connections consistent
+with its Service Binding, and does not automatically apply to all Alt-Svc
+field values for the origin.)
 
 ## Requiring Server Name Indication
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1092,6 +1092,7 @@ and port override the alt-authority.  For example, suppose that
     Alt-Svc: h2="alt.example.com:443", h3=":8443"
 
 The client would retrieve the following HTTPS records:
+
     alt.example.com. IN HTTPS 1 . alpn=h2,h3 ech=...
     _8443._https.example.com. IN HTTPS 1 alt3.example.com. (
         port=9443 alpn=h2,h3 ech=... )
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1082,28 +1082,28 @@ single-client granularity.
 
 ## Interaction with Alt-Svc
 
-This specification does not alter the DNS records used when connecting
-to an Alt-Svc endpoint (typically A and/or AAAA only) or the data contained
-in Alt-Svc field values.  Origins can publish HTTPS records without any
-need to modify their existing Alt-Svc arrangements.  Future specifications
-may enable more integration of Alt-Svc and HTTPS records, such as querying
-HTTPS records based on an Alt-Svc field value.
+Clients that implement support for both Alt-Svc and HTTPS records SHOULD
+retrieve any HTTPS records for the Alt-Svc alt-authority, and ensure that
+their connection attempts are consistent with both the Alt-Svc parameters
+and any received HTTPS SvcParams.  For example, suppose
+"https://example.com" sends an Alt-Svc field value of:
 
-In general, clients MAY skip the HTTPS RR query
-if a usable Alt-Svc value is available in the local cache.
-If Alt-Svc connection fails, these clients SHOULD fall back to the HTTPS RR
-client connection procedure ({{client-behavior}}).
+    Alt-Svc: h3="alt.example.com:8443"; ma=...
 
-Clients that support ECH are subject to additional requirements.  If all HTTPS
-records for the origin have the "ech" SvcParamKey, ECH-capable clients MUST
-use ECH on all connections for this origin (as mandated in
-{{ech-client-behavior}}).  To comply with this requirement, clients will
-typically have to retrieve any HTTPS records for the origin before making use
-of any cached Alt-Svc field values (which currently cannot provide an
-ECHConfigList), and ignore any Alt-Svc field value whose use would require
-disabling ECH.  (The "ech" SvcParamValue is only for connections consistent
-with its Service Binding, and does not automatically apply to all Alt-Svc
-field values for the origin.)
+The client would retrieve the following HTTPS record:
+
+    _8443._https.alt.example.com. ... 1 alt.example.com. alpn=h2,h3 ech=...
+
+The client could then attempt an HTTP/3 connection to `alt.example.com:8443`
+with ECH, as this is consistent with both the Alt-Svc field value and the
+HTTPS record.
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any Alt-Svc hostnames.  Otherwise,
+clients may reveal the unencrypted ClientHello during Alt-Svc connections.
+Similar consistency considerations may apply to future SvcParamKeys, so
+alt-authorities SHOULD carry the same SvcParams as the origin unless
+a deviation is specifically known to be safe.
 
 ## Requiring Server Name Indication
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1104,7 +1104,7 @@ interpretation of Alt-Svc's `protocol-id` field.
 Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
 also publish an "ech" SvcParam for any Alt-Svc hostnames.  Otherwise,
 clients may reveal the unencrypted ClientHello during Alt-Svc connections.
-Similar consistency considerations may apply to future SvcParamKeys, so
+Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
 a deviation is specifically known to be safe.
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1110,9 +1110,9 @@ also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
 clients might reveal the name of the server in an unencrypted ClientHello.
 Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
-a deviation is specifically known to be safe.  Clients MAY impose additional
-conditions on Alt-Svc connections, e.g. requiring ECH if the origin's
-SvcParams offered ECH.
+a deviation is specifically known to be safe.  Clients MAY impose their
+own consistency conditions on Alt-Svc connections, e.g. requiring ECH if the
+origin's SvcParams offered ECH.
 
 ## Requiring Server Name Indication
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1132,7 +1132,7 @@ Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
 a deviation is specifically known to be safe.
 
-As noted in Section 2.4 of {{AltSvc}}, clients MAY disallow any Alt-Svc
+As noted in {{Section 2.4 of AltSvc}}, clients MAY disallow any Alt-Svc
 connection according to their own criteria, e.g. disallowing Alt-Svc
 connections that lack ECH support when there is an active ECH-protected
 connection for this origin.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1089,7 +1089,7 @@ and any received HTTPS SvcParams.  If present, the HTTPS record's target
 and port override the alt-authority.  For example, suppose that
 "https://example.com" sends an Alt-Svc field value of:
 
-~~~ http-message
+~~~ HTTP
 Alt-Svc: h2="alt.example.com:443", h3=":8443"
 ~~~
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1105,8 +1105,9 @@ with ECH, or an HTTP over TLS connection to `alt.example:443` with ECH,
 as these options are consistent with both an Alt-Svc field value and its
 HTTPS record.  An ECH-capable client would not connect to `alt2.example`,
 as there is no possible connection that is consistent with both ALPN values
-("h2" from Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam triggers
-SVCB-reliant client behavior ({{ech-client-behavior}}).  Operators SHOULD
+("h2" from Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam prevents
+the client from falling back to a plain A/AAAA-based connection (i.e. the
+client is SVCB-reliant, {{ech-client-behavior}}).  Operators SHOULD
 avoid such inconsistencies, and clients SHOULD treat them as Alt-Svc
 connection failures, falling back to non-Alt-Svc connection when appropriate
 as described in Section 2.4 of {{!AltSvc}}.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1089,10 +1089,11 @@ and any received HTTPS SvcParams.  If present, the HTTPS record's target
 and port override the alt-authority.  For example, suppose that
 "https://example.com" sends an Alt-Svc field value of:
 
-    Alt-Svc: h3="alt.example.com:8443"; ...
+    Alt-Svc: h3="alt.example.com:8443"; h2="alt.example.com:443"; ...
 
-The client would retrieve the following HTTPS record:
-
+The client would retrieve the following HTTPS records:
+    alt.example.com. ... 1 alt2.example.com. (
+            port=9443 alpn=h2,h3 ech=... )
     _8443._https.alt.example.com. ... 1 alt2.example.com. (
         port=9443 alpn=h2,h3 ech=... )
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1107,10 +1107,12 @@ interpretation of Alt-Svc's `protocol-id` field.
 
 Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
 also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
-clients may reveal the unencrypted ClientHello during Alt-Svc connections.
+clients might reveal the name of the server in an unencrypted ClientHello.
 Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
-a deviation is specifically known to be safe.
+a deviation is specifically known to be safe.  Clients MAY impose additional
+conditions on Alt-Svc connections, e.g. requiring ECH if the origin's
+SvcParams offered ECH.
 
 ## Requiring Server Name Indication
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1096,34 +1096,44 @@ Alt-Svc: h2="alt.example:443", h2="alt2.example:443", h3=":8443"
 The client would retrieve the following HTTPS records:
 
     alt.example.              IN HTTPS 1 . alpn=h2,h3 ech=...
-    alt2.example.             IN HTTPS 1 . alpn=h3 ech=...
+    alt2.example.             IN HTTPS 1 alt2b.example. alpn=h3 ech=...
     _8443._https.example.com. IN HTTPS 1 alt3.example. (
         port=9443 alpn=h2,h3 ech=... )
 
-The client could then attempt an HTTP/3 connection to `alt3.example:9443`
-with ECH, or an HTTP over TLS connection to `alt.example:443` with ECH,
-as these options are consistent with both an Alt-Svc field value and its
-HTTPS record.  An ECH-capable client would not connect to `alt2.example`,
-as there is no possible connection that is consistent with both ALPN values
-("h2" from Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam prevents
-the client from falling back to a plain A/AAAA-based connection (i.e. the
-client is SVCB-reliant, {{ech-client-behavior}}).  Operators SHOULD
-avoid such inconsistencies, and clients SHOULD treat them as Alt-Svc
-connection failures, falling back to non-Alt-Svc connection when appropriate
-as described in Section 2.4 of {{!AltSvc}}.
+Based on these inputs, the following connection attempts would always be
+allowed:
 
-This specification does not alter the interpretation of Alt-Svc's
-`protocol-id` field.
+* HTTPS over TLS to `alt.example:443` (Alt-Svc and HTTPS record)
+* HTTP/3 to `alt3.example:9443` (Alt-Svc and HTTPS record)
+* Fallback to the the client's non-Alt-Svc connection behavior
+
+ECH-capable clients would use ECH when establishing any of these connections.
+
+The following connection attempts would not be allowed:
+
+* HTTP/3 to `alt.example:443` (not consistent with Alt-Svc)
+* Any connection to `alt2b.example` (no ALPN consistent with both the HTTPS
+  record and Alt-Svc)
+* HTTPS over TLS to any port on `alt3.example` (not consistent with Alt-Svc)
+
+The following connection attempts would be allowed only if the client does
+not support ECH, as they rely on SVCB-optional fallback behavior that is
+disabled when the "ech" SvcParam is present ({{ech-client-behavior}}):
+
+* HTTPS over TLS to `alt2.example:443` (Alt-Svc only)
+* HTTP/3 to `example.com:8443` (Alt-Svc only)
 
 Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
 also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
 clients might reveal the name of the server in an unencrypted ClientHello.
 Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
-a deviation is specifically known to be safe.  As noted in Section 2.4 of
-{{AltSvc}}, clients MAY disallow any Alt-Svc connection according to their
-own criteria, e.g. disallowing Alt-Svc connections that lack ECH support
-if ECH is supported on the origin.
+a deviation is specifically known to be safe.
+
+As noted in Section 2.4 of {{AltSvc}}, clients MAY disallow any Alt-Svc
+connection according to their own criteria, e.g. disallowing Alt-Svc
+connections that lack ECH support when there is an active ECH-protected
+connection for this origin.
 
 ## Requiring Server Name Indication
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1090,10 +1090,14 @@ client connection procedure ({{client-behavior}}).
 
 Clients that implement support for ECH MUST perform the HTTPS RR query first,
 and MUST only make use of Alt-Svc when operating in SVCB-optional mode (see
-{{ech-client-behavior}}).
+{{ech-client-behavior}}).  This rule allows origins to send Alt-Svc
+advertisements (which do not currently support ECH) without compromising the
+privacy assurances of ECH configurations they have published in the DNS.
 
 This specification does not alter the DNS records used when connecting
-to an Alt-Svc hostname (typically A and/or AAAA only).
+to an Alt-Svc hostname (typically A and/or AAAA only) or the data contained
+in Alt-Svc field values.  Future specifications may enable more use of
+Alt-Svc with HTTPS RR queries and/or ECH.
 
 ## Requiring Server Name Indication
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1086,7 +1086,7 @@ Clients that implement support for both Alt-Svc and HTTPS records SHOULD
 retrieve any HTTPS records for the Alt-Svc alt-authority, and ensure that
 their connection attempts are consistent with both the Alt-Svc parameters
 and any received HTTPS SvcParams.  If present, the HTTPS record's target
-and port override the alt-authority.  For example, suppose
+and port override the alt-authority.  For example, suppose that
 "https://example.com" sends an Alt-Svc field value of:
 
     Alt-Svc: h3="alt.example.com:8443"; ...
@@ -1102,7 +1102,7 @@ HTTPS record.  This specification does not alter or clarify the
 interpretation of Alt-Svc's `protocol-id` field.
 
 Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
-also publish an "ech" SvcParam for any Alt-Svc hostnames.  Otherwise,
+also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
 clients may reveal the unencrypted ClientHello during Alt-Svc connections.
 Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1089,16 +1089,16 @@ and any received HTTPS SvcParams.  If present, the HTTPS record's target
 and port override the alt-authority.  For example, suppose that
 "https://example.com" sends an Alt-Svc field value of:
 
-    Alt-Svc: h3="alt.example.com:8443"; h2="alt.example.com:443"; ...
+    Alt-Svc: h2="alt.example.com:443", h3=":8443"
 
 The client would retrieve the following HTTPS records:
-    alt.example.com. ... 1 alt2.example.com. (
-            port=9443 alpn=h2,h3 ech=... )
-    _8443._https.alt.example.com. ... 1 alt2.example.com. (
+    alt.example.com. IN HTTPS 1 . alpn=h2,h3 ech=...
+    _8443._https.example.com. IN HTTPS 1 alt3.example.com. (
         port=9443 alpn=h2,h3 ech=... )
 
-The client could then attempt an HTTP/3 connection to `alt2.example.com:9443`
-with ECH, as this is consistent with both the Alt-Svc field value and the
+The client could then attempt an HTTP/3 connection to `alt3.example.com:9443`
+with ECH, or an HTTP over TLS connection to `alt.example.com:443` with ECH,
+as these options are consistent with both an Alt-Svc field value and its
 HTTPS record.  This specification does not alter or clarify the
 interpretation of Alt-Svc's `protocol-id` field.
 
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1103,13 +1103,13 @@ The client would retrieve the following HTTPS records:
 The client could then attempt an HTTP/3 connection to `alt3.example:9443`
 with ECH, or an HTTP over TLS connection to `alt.example:443` with ECH,
 as these options are consistent with both an Alt-Svc field value and its
-HTTPS record.  It would not connect to `alt2.example`, as there is no
-possible connection that is consistent with both ALPN values ("h2" from
-Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam triggers SVCB-reliant
-client behavior ({{ech-client-behavior}}).  Operators SHOULD avoid
-such inconsistencies, and clients SHOULD treat them as Alt-Svc connection
-failures, falling back to non-Alt-Svc connection when appropriate as
-described in Section 2.4 of {{!RFC7838}}
+HTTPS record.  An ECH-capable client would not connect to `alt2.example`,
+as there is no possible connection that is consistent with both ALPN values
+("h2" from Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam triggers
+SVCB-reliant client behavior ({{ech-client-behavior}}).  Operators SHOULD
+avoid such inconsistencies, and clients SHOULD treat them as Alt-Svc
+connection failures, falling back to non-Alt-Svc connection when appropriate
+as described in Section 2.4 of {{!RFC7838}}
 
 This specification does not alter the interpretation of Alt-Svc's
 `protocol-id` field.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1085,18 +1085,21 @@ single-client granularity.
 Clients that implement support for both Alt-Svc and HTTPS records SHOULD
 retrieve any HTTPS records for the Alt-Svc alt-authority, and ensure that
 their connection attempts are consistent with both the Alt-Svc parameters
-and any received HTTPS SvcParams.  For example, suppose
+and any received HTTPS SvcParams.  If present, the HTTPS record's target
+and port override the alt-authority.  For example, suppose
 "https://example.com" sends an Alt-Svc field value of:
 
-    Alt-Svc: h3="alt.example.com:8443"; ma=...
+    Alt-Svc: h3="alt.example.com:8443"; ...
 
 The client would retrieve the following HTTPS record:
 
-    _8443._https.alt.example.com. ... 1 alt.example.com. alpn=h2,h3 ech=...
+    _8443._https.alt.example.com. ... 1 alt2.example.com. (
+        port=9443 alpn=h2,h3 ech=... )
 
-The client could then attempt an HTTP/3 connection to `alt.example.com:8443`
+The client could then attempt an HTTP/3 connection to `alt2.example.com:9443`
 with ECH, as this is consistent with both the Alt-Svc field value and the
-HTTPS record.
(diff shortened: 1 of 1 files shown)


enygren:
I like this direction better.  Some things we should cover:

* alt-authorities may not have HTTPS records, just A/AAAA records.  
* What to do if an alt-authority has no HTTPS records, just A/AAAA, and a client is SVCB-required.  (Answer would be to ignore the Alt-Svc.)
* How to handle ports in the HTTPS record vs Alt-Svc record.  Should they be required to match?  We should be explicit.
* How to handle compatible ALPNs where negotiation happens at TLS-time (eg, h2 vs http/1.1).  We have better logic here and this is something Alt-Svc-bis will want to fix.  In the meantime, do we want to allow these to be considered compatible or not?

bemasc:
OK, please suggest text on those counts.

Personally, I think your list looks like a list of reasons not to do this.  We are attempting to delicately stitch together two different data models, with an N^2-like complexity of interacting features, which seems much more difficult than just declaring that you do one or the other.

bemasc:
I've updated the text to address some of those concerns.

> * alt-authorities may not have HTTPS records, just A/AAAA records.

I think this case is clear in the text.

> * What to do if an alt-authority has no HTTPS records, just A/AAAA, and a client is SVCB-required.  (Answer would be to ignore the Alt-Svc.)

"Clients for pre-existing protocols (e.g.  HTTPS) SHALL implement SVCB-optional behavior", so I don't think we need to worry about that.

> * How to handle ports in the HTTPS record vs Alt-Svc record.  Should they be required to match?  We should be explicit.

OK, added a sentence.

> * How to handle compatible ALPNs where negotiation happens at TLS-time (eg, h2 vs http/1.1).  We have better logic here and this is something Alt-Svc-bis will want to fix.  In the meantime, do we want to allow these to be considered compatible or not?

The endless debate is about this text in RFC 7838

> If the connection to the alternative service does not negotiate the expected protocol (for example, ALPN fails to negotiate h2, or an Upgrade request to h2c is not accepted), the connection to the alternative service MUST be considered to have failed.

This text is ambiguous, but I think "fails to negotiate" means that clients MAY send just the specified ALPN, not that servers MUST prefer it.  Regardless, I am hesitant to mess with this here, due to the risk of needing to Update RFC 7838.  I added a sentence alluding to that.

There is an alternative formulation in which we says that the Alt-Svc ALPN is added to the SVCB ALPN set, but this seems like a bigger change to Alt-Svc, and it would not improve security.

ericorth:
> > * How to handle compatible ALPNs where negotiation happens at TLS-time (eg, h2 vs http/1.1).  We have better logic here and this is something Alt-Svc-bis will want to fix.  In the meantime, do we want to allow these to be considered compatible or not?
> 
> The endless debate is about this text in RFC 7838
> 
> > If the connection to the alternative service does not negotiate the expected protocol (for example, ALPN fails to negotiate h2, or an Upgrade request to h2c is not accepted), the connection to the alternative service MUST be considered to have failed.
> 
> This text is ambiguous, but I think "fails to negotiate" means that clients MAY send just the specified ALPN, not that servers MUST prefer it. Regardless, I am hesitant to mess with this here, due to the risk of needing to Update RFC 7838. I added a sentence alluding to that.
> 
> There is an alternative formulation in which we says that the Alt-Svc ALPN is added to the SVCB ALPN set, but this seems like a bigger change to Alt-Svc, and it would not improve security.

I agree that HTTPS's looser ALPN renegotiation is generally better, but I think Ben has it right that we can't mess with that here without updating RFC7838.  (Doesn't make a difference either way for Chrome's current implementations where we only support h3 Alt-Svc which as far as I know can't be renegotiated.)

ericorth:
And in general, yes, I like this direction and feel it does a good job capturing my earlier proposal.

ericorth:
I think we might need to add a sentence clarifying SVCB-optional behavior.

When, per the SVCB-optional behavior the client acts as if there's an extra endpoint at the end without any SvcParams, I assume we want to be able to connect to that using the Alt-Svc protocol.  But I think things could be misinterpreted into technically blocking fallback to that endpoint because it is specified as an endpoint with "no SvcParams", which is just the default ALPN and thus not consistent with the Alt-Svc if the Alt-Svc specifies a different protocol.

Should this text clarify that the extra SVCB-optional endpoint is treated as having the necessary SvcParams to be consistent with Alt-Svc parameters to allow a normal connection that follows SVCB aliases? Or is there a desire to not allow SVCB aliasing with Alt-Svc, and thus only allow fallback to pure non-SVCB connection?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,31 @@ single-client granularity.
...
+        port=9443 alpn=h2,h3 ech=... )
+
+The client could then attempt an HTTP/3 connection to `alt2.example.com:9443`
+with ECH, as this is consistent with both the Alt-Svc field value and the
+HTTPS record.  This specification does not alter or clarify the
+interpretation of Alt-Svc's `protocol-id` field.
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any Alt-Svc hostnames.  Otherwise,
+clients may reveal the unencrypted ClientHello during Alt-Svc connections.
```
martinthomson, line 1106: Is it the case that the client will use the alternative service without ECH, or that it won't use the alternative service?

Or do we want to leave that unspecified?  I know that we allow some HTTPS/SVCB records to include ECH and some to omit it and that client policy governs which are chosen.  Is this just another case where client policy applies?

Personally, I would very much prefer saying that if you use ECH to connect, you MUST NOT use an alternative that doesn't have an ECH config.  "SHOULD NOT" would be acceptable on the grounds that we aren't dictating client policy.  But it's a pretty bad leak, and one that could be the result of a mistake in server configuration.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,31 @@ single-client granularity.
...
+Similar consistency considerations may apply to future SvcParamKeys, so
```
martinthomson, line 1107:
```suggestion
Similar consistency considerations could apply to future SvcParamKeys, so
```

martinthomson (commented):
I like this a lot more than the other one.

I have my reservations about the squishiness here.

I would be interested in the opinion of @ekr, @huitema, @chris-wood, @carickdb, @cjpatton, and @davidben on this issue as this relates directly to ECH.

draft-ietf-dnsop-svcb-https.md
bemasc: > Or do we want to leave that unspecified?

FWIW, that's what I mean in this text (equivalent to normative MAY).

> Personally, I would very much prefer saying that if you use ECH to connect, you MUST NOT use an alternative that doesn't have an ECH config.

A client that wanted to implement that behavior would have to query _both_ the origin and the alt-authority for HTTPS records before it could proceed with an Alt-Svc connection.  If we want clients to take responsibility for ensuring ECH isn't lost, I think #327 makes more sense.

P.S. Typo: @carrickdb

ericorth: > > Personally, I would very much prefer saying that if you use ECH to connect, you MUST NOT use an alternative that doesn't have an ECH config.
> 
> A client that wanted to implement that behavior would have to query _both_ the origin and the alt-authority for HTTPS records before it could proceed with an Alt-Svc connection. If we want clients to take responsibility for ensuring ECH isn't lost, I think #327 makes more sense.

I think we're sufficiently covered due to the assumption (supported by a SHOULD here in the text) that if an origin publishes ECH for the origin endpoints, it also publishes ECH for the Alt-Svc endpoints.  Means that with a single DNS query, the client can find whether the server supports ECH unless the domains are stupidly configured (a similar level of stupidity as just forgetting to publish ECH keys on any other endpoint).

The current requirement is already that only if all endpoints found have ECH, the client MUST use the the ECH endpoints.  If a server is misconfigured to give non-ECH endpoints, I think it's best left to client choice how to prioritize ECH vs other factors like h3 (e.g. a non-privacy-sensitive client may choose to prioritize using an h3 endpoint if only h2 endpoints have ECH, or an extra-privacy-sensitive client may choose to not allow Alt-Svc at all unless ECH endpoints are found or may want to take the performance hit to query all Alt-Svc names and the origin to find any possible ECH endpoints before connecting to anything).

No matter what the client chooses as endpoint priorities, the proper configuration for an origin to force ECH is clear (and clearly recommended by section 9.2): Include ECH on all endpoints and section 9.1 requires ECH-capable clients to only connect via those endpoints.  This just now would apply to endpoints given on Alt-Svc names too.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,31 @@ single-client granularity.
...
+    _8443._https.alt.example.com. ... 1 alt2.example.com. (
+        port=9443 alpn=h2,h3 ech=... )
+
+The client could then attempt an HTTP/3 connection to `alt2.example.com:9443`
+with ECH, as this is consistent with both the Alt-Svc field value and the
+HTTPS record.  This specification does not alter or clarify the
+interpretation of Alt-Svc's `protocol-id` field.
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any Alt-Svc hostnames.  Otherwise,
```
enygren, line 1105:
```suggestion
also publish an "ech" SvcParam for any Alt-Svc alt-authority hostnames.  Otherwise,
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,31 @@ single-client granularity.
...
+their connection attempts are consistent with both the Alt-Svc parameters
+and any received HTTPS SvcParams.  If present, the HTTPS record's target
+and port override the alt-authority.  For example, suppose
+"https://example.com" sends an Alt-Svc field value of:
+
+    Alt-Svc: h3="alt.example.com:8443"; ...
+
+The client would retrieve the following HTTPS record:
+
+    _8443._https.alt.example.com. ... 1 alt2.example.com. (
```
enygren, line 1096: Do we also want to add an example in the appendix without the port prefix?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,31 @@ single-client granularity.
...
+        port=9443 alpn=h2,h3 ech=... )
+
+The client could then attempt an HTTP/3 connection to `alt2.example.com:9443`
+with ECH, as this is consistent with both the Alt-Svc field value and the
+HTTPS record.  This specification does not alter or clarify the
+interpretation of Alt-Svc's `protocol-id` field.
```
enygren, line 1102: Do we also want to add a:
"Clients MAY choose to ignore individual returned HTTPS RRs,
or the Alt-Svc entry entirely, based on mismatches between the 
Alt-Svc record and HTTPS RR SvcParams."
This gives flexibility here for client behavior based on ECS
as well as for cases like the Chrome behavior of wanting
to match h3-to-h3 and h2-to-tcp but without over-specifying.

enygren (changes requested):
This is looking generally good.  I'll leave the question as to whether we want to be stricter about ECH to others.  Perhaps we merge this and then look at the ECH section to see if we want some tightening or clarification there?

draft-ietf-dnsop-svcb-https.md
bemasc: It seems clear enough to me as-is, but if you think it's confusing, we can add it.

draft-ietf-dnsop-svcb-https.md
bemasc: I'm not sure I understand, but I don't think this is necessary.  If a record cannot be used in a way that is consistent with the Alt-Svc Field Value, then the record is effectively ignored when attempting to use that Alt-Svc Field Value.

bemasc:
> When, per the SVCB-optional behavior the client acts as if there's an extra endpoint at the end without any SvcParams, I assume we want to be able to connect to that using the Alt-Svc protocol. But I think things could be misinterpreted into technically blocking fallback to that endpoint because it is specified as an endpoint with "no SvcParams", which is just the default ALPN and thus not consistent with the Alt-Svc if the Alt-Svc specifies a different protocol.

As you noted, this only applies to "Alias-only" HTTPS record setups on the alt-authority.  For that weird case, I think either client behavior is acceptable.  If you really want Alt-Svc clients to follow that alias, you should really just update your Alt-Svc hostname to point to the target ... or use a CNAME.  (Alt-authorities aren't user-visible, so there's no need for them to be at a zone apex.)

draft-ietf-dnsop-svcb-https.md
bemasc: Fixed!

ericorth:
> > When, per the SVCB-optional behavior the client acts as if there's an extra endpoint at the end without any SvcParams, I assume we want to be able to connect to that using the Alt-Svc protocol. But I think things could be misinterpreted into technically blocking fallback to that endpoint because it is specified as an endpoint with "no SvcParams", which is just the default ALPN and thus not consistent with the Alt-Svc if the Alt-Svc specifies a different protocol.
> 
> As you noted, this only applies to "Alias-only" HTTPS record setups on the alt-authority. For that weird case, I think either client behavior is acceptable. If you really want Alt-Svc clients to follow that alias, you should really just update your Alt-Svc hostname to point to the target ... or use a CNAME. (Alt-authorities aren't user-visible, so there's no need for them to be at a zone apex.)

I think you're right that enough alternatives exist here for that the server should be able to easily do what it needs to do either way.  I'm just worried about whether or not the client should have to worry about making sure something specific happen or just let it get treated however the client implementation would otherwise treat it.  I think a simple added "MAY" could simplify implementations quite a bit to make it explicit that either client behavior is acceptable and non-important.  Otherwise, I'm worried that things technically work out into an implicit MUST that the client has to make sure to block the extra endpoint in some scenarios, and I'm honestly not sure which way things will work out easiest to implement.

How about:
"SVCB-optional clients MAY make use of AliasMode HTTPS records with Alt-Svc parameters by attempting connections to the final alias query name with the Alt-Svc parameters."?

bemasc:
@ericorth This whole section is only at "SHOULD" strength, so I don't think we need to worry about an "implicit MUST".

The client behavior for Alt-Svc (in RFC 7838) is specified loosely, with the understanding that if it doesn't work clients will normally just fall back to regular connection.  Clients can pretty much try, or not try, whatever they want.  From that perspective, I think adding more text on every corner case is likely to obscure the essentials more than it helps.

ericorth:
> @ericorth This whole section is only at "SHOULD" strength, so I don't think we need to worry about an "implicit MUST".
> 
> The client behavior for Alt-Svc (in RFC 7838) is specified loosely, with the understanding that if it doesn't work clients will normally just fall back to regular connection. Clients can pretty much try, or not try, whatever they want. From that perspective, I think adding more text on every corner case is likely to obscure the essentials more than it helps.

Maybe some of the other text elsewhere around SVCB-optional client algorithms and such are bordering on over-specified, because it feels to me that they mandate a specific behavior that doesn't quite make sense here.  But if nobody else feels that way and everybody is happy with the practical interpretation of clients doing reasonable things falling back to using Alt-Svc parameters with SVCB alias-only or non-alias, then I guess I'm fine with that.

bemasc:
I believe this is the last remaining PR blocking the post-WGLC revision.  Please review this version (especially if you have thoughts on ECH) and indicate whether it looks acceptable.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,31 @@ single-client granularity.
...
+their connection attempts are consistent with both the Alt-Svc parameters
+and any received HTTPS SvcParams.  If present, the HTTPS record's target
+and port override the alt-authority.  For example, suppose that
+"https://example.com" sends an Alt-Svc field value of:
+
+    Alt-Svc: h3="alt.example.com:8443"; ...
+
+The client would retrieve the following HTTPS record:
+
+    _8443._https.alt.example.com. ... 1 alt2.example.com. (
```
enygren, line 1096:
```suggestion
    Alt-Svc: h3="alt3.example.com:8443"; h2="alt2.example.com:443"; ...

The client would retrieve the following HTTPS records:

    _8443._https.alt3.example.com. ... 1 target3.example.com. (
    alt2.example.com. ... 1 target2.example.com. (
```

draft-ietf-dnsop-svcb-https.md
enygren: It seems good to add just to be crystal clear that h3="alt.example.com:443" only looks up the HTTPS RR for "alt.example.com".
Just showing both forms here as below might be one way to do this.

draft-ietf-dnsop-svcb-https.md
enygren: This would be giving the client permission to do filtering.  In the above example, h3 would only be used for the HTTPS RR at _8443._https.alt.example.com and h2 would only be used for the HTTPS RR at alt.example.com

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I added your suggested example and tweaked it slightly to (1) simplify, (2) cover the case where there is no `port` SvcParam, and (3) cover the case where the Alt-Svc hostname is implicit.

draft-ietf-dnsop-svcb-https.md
bemasc: In this example, neither RR is "ignored".  They're "constrained" or "filtered", perhaps.

I've expanded the text to match the new example.  I think the new text conveys the idea pretty clearly.

enygren (approved):
The new example covers things fairly well.  It seems like more clear clarification can be in a future Alt-Svc update.  Hopefully this is not under-specified?

enygren:
@martinthomson  and @ericorth : are you good with what we have here now?

ericorth:
> @martinthomson and @ericorth : are you good with what we have here now?

Looks good to me.  I think we've hit all the main necessary points.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,32 @@ single-client granularity.
...
+retrieve any HTTPS records for the Alt-Svc alt-authority, and ensure that
+their connection attempts are consistent with both the Alt-Svc parameters
+and any received HTTPS SvcParams.  If present, the HTTPS record's target
+and port override the alt-authority.  For example, suppose that
+"https://example.com" sends an Alt-Svc field value of:
+
+    Alt-Svc: h2="alt.example.com:443", h3=":8443"
+
+The client would retrieve the following HTTPS records:
+    alt.example.com. IN HTTPS 1 . alpn=h2,h3 ech=...
```
martinthomson, line 1095:
```suggestion

    alt.example.com. IN HTTPS 1 . alpn=h2,h3 ech=...
```
I think that you need vertical space here for this to render correctly.

Using `~~~` around figures is good hygiene and that allows you to [annotate](https://www.rfc-editor.org/materials/sourcecode-types.txt) them correctly.  (Like using `~~~http-message` for the example above.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,32 @@ single-client granularity.
...
+
+The client could then attempt an HTTP/3 connection to `alt3.example.com:9443`
+with ECH, or an HTTP over TLS connection to `alt.example.com:443` with ECH,
+as these options are consistent with both an Alt-Svc field value and its
+HTTPS record.  This specification does not alter or clarify the
+interpretation of Alt-Svc's `protocol-id` field.
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
+clients may reveal the unencrypted ClientHello during Alt-Svc connections.
```
martinthomson, line 1107: I would prefer:

> Clients that connect to an alt-authority would reveal the name of the server in an unencrypted ClientHello.  Clients SHOULD NOT use an alt-authority if the original connection uses ECH and no "ech" SvcParam is provided for the alt-authority.

This is not saying that the client could (or may) reveal identity, but saying that if it did connect then it would.  THEN I want a recommendation against doing the stupid thing.  I would settle for just the first sentence, but I do really prefer both.

martinthomson (changes requested):

draft-ietf-dnsop-svcb-https.md
bemasc: Just to be clear, you're proposing that implementations should add an indication to each Alt-Svc cache entry about whether ECH was in use on the connection over which that entry was received, so that ECH consistency can be enforced when that cache entry is used (perhaps several weeks later)?

martinthomson: No, I'm suggesting that client implementations that start using ECH keep using ECH.  The decision to use an alternative is made in the moment and can be based on what is active. That's consistent with the other text you have.  

I hadn't fully considered the point about caching.  That is, if we use ECH on Tuesday, but fetch and reject an Alt-Svc advertisement for lack of ECH, do you cache that entry?  

This is really something that an Alt-Svc revision needs to deal with, so we might need to be careful here.  I can imagine that the decision about whether to insist on ECH would be made anew on Friday.  My preference is that yes, if you do cache the Alt-Svc advertisement you also remember that it was disqualified for lack of ECH or just that it was not acceptable at the time it was retrieved.  For me, that marker would be cause to avoid connecting immediately to the alternative without first going to the primary.  Which gets back to the SHOULD.

ericorth: I certainly hope most clients would get smart about not changing away from an ECH connection to a non-ECH connection, but I think this is hard to create any actual rules about without over-specifying things and I suggest we not get into it.  E.g., how long after a successful ECH connection should a client be rejecting non-ECH Alt-Svc?

Plus, I think leaving this up to the client to make a good decision is consistent with the rest of the spec.  If the server gives some ECH endpoints and some non-ECH endpoints, it's up to the client to make the decisions over how/if to prefer endpoints that support ECH.  So the spec is saying if a server really wants to require ECH for ECH-capable clients, they need to include ECH on all endpoints, including Alt-Svc endpoints, and this is the recommended/should'ed server configuration.  If the server doesn't configure that, most clients will ideally still choose to prefer ECH, but that's up to the client, and some clients may have reason to prefer any route to h3, thus potentially disabling ECH.

bemasc: > My preference is that yes, if you do cache the Alt-Svc advertisement you also remember that it was disqualified for lack of ECH or just that it was not acceptable at the time it was retrieved.

Alt-Svc field values don't say anything about ECH.  That's a property of the HTTPS record for the alt-authority, which is not known (and may change) until the client attempts the Alt-Svc connection, which may not be until much later.

It sounds like we want to support this behavior, but the details are too unsettled to make a normative recommendation.  I've adjusted the explanation based on @martinthomson's suggestion, and added the following sentence:
> Clients MAY impose their own consistency conditions on Alt-Svc connections, e.g. requiring ECH if the origin's SvcParams offered ECH.

Please review.

enygren (approved):

draft-ietf-dnsop-svcb-https.md
ericorth: Current text still looks good to me.  The important thing here is that we give clients a technical path to ECH and make clear to domains the strongly recommended way to configure it to make sure clients can use it.  I think we have that.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,37 @@ single-client granularity.
...
+
+~~~ HTTP
+Alt-Svc: h2="alt.example.com:443", h3=":8443"
+~~~
+
+The client would retrieve the following HTTPS records:
+
+    alt.example.com. IN HTTPS 1 . alpn=h2,h3 ech=...
+    _8443._https.example.com. IN HTTPS 1 alt3.example.com. (
+        port=9443 alpn=h2,h3 ech=... )
```
davidben, line 1100: Probably good to have an incompatible example too. Something like...

Alt-Svc example:
```
Alt-Svc: h2="alt.example.com:443", h2="alt2.example.com:443", h3=":8443"
```

DNS example:
```
    alt.example.com. IN HTTPS 1 . alpn=h2,h3 ech=...
    alt2.example.com. IN HTTPS 1 . alpn=h3 ech=...
    _8443._https.example.com. IN HTTPS 1 alt3.example.com. (
        port=9443 alpn=h2,h3 ech=... )
```

New text:

> The client could then attempt an HTTP/3 connection to `alt3.example.com:9443` with ECH, or an HTTP over TLS connection to `alt.example.com:443` with ECH, as these options are consistent with both an Alt-Svc field value and its HTTPS record. It, however, could not attempt an HTTP over TLS connection to `alt2.example.com:443` as this option is not compatible with the Alt-Svc value's ALPN protocol.

Perhaps it's also worth something like:

> An incompatible SVCB option for an Alt-Svc value is considered to have failed. If all SVCB options for an Alt-Svc value fail, an SVCB-optional client ({{client-behavior}}) MAY fall back to a direct connection to the Alt-Svc value, unless prohibited by {{client-failures}} or {{ech-client-behavior}}. If all permitted SVCB or fallback options for an Alt-Svc value fail, the Alt-Svc connection is considered to have failed. The client MAY fall back to using the origin or another alternative service, as described in {{Section 2.4 of RFC7838}}.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,37 @@ single-client granularity.
...
+interpretation of Alt-Svc's `protocol-id` field.
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
+clients might reveal the name of the server in an unencrypted ClientHello.
+Similar consistency considerations could apply to future SvcParamKeys, so
+alt-authorities SHOULD carry the same SvcParams as the origin unless
+a deviation is specifically known to be safe.  Clients MAY impose their
+own consistency conditions on Alt-Svc connections, e.g. requiring ECH if the
+origin's SvcParams offered ECH.
```
davidben, line 1115: This sentence is unclear to me. Is the idea that the client would still fetch the origin's HTTPS record? Or is it that the client maintains some state from the last time it fetched the HTTPS record, which may have been quite some time ago? I assume it's not the former. But the latter is very stateful and probably deserves more specific text if that's the intent.

davidben (commented):
This works for me. Specifically, it meets these goals:

* Origins should not have to choose between their current h3 deployments and ECH. It should be easy for an origin to reliably enable ECH after having previously deployed h3 with Alt-Svc, and without sacrificing that deployment.
* It should not be a requirement to stick ECH keys inside Alt-Svc. Alt-Svc only applies to the next HTTP request, which means it needs a qualitatively longer lifetime than HTTPS/SVCB. ECH keys in Alt-Svc may well be something to explore later, but it should be possible to deploy ECH without committing yourself to that lifetime of key.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I've added that example and incorporated various elements of this text.

draft-ietf-dnsop-svcb-https.md
bemasc: We don't have consensus to do any specific version of this, so there's no need to write it out in great detail.  I've made this vaguer, which I hope is clearer: "Clients MAY impose their own consistency conditions on Alt-Svc connections, e.g. requiring ECH **if the origin is known to support** ECH.".

I wouldn't recommend Alt-Svc ECH consistency enforcement to a client implementor, but I think it's permissible within the very lax rules of Alt-Svc, and some people clearly want to implement it.  A correct and reliable implementation would have to re-query the origin's HTTPS record along with the alt-authority's in order to have them both fresh for a comparison (your first option).  However, a client that only wants partial consistency enforcement could gate this check on a cached hint (your second option), or apply the enforcement only when a connection to the origin is still open, or when the origin's HTTPS record is still in the local cache.

But this is all complicated and speculative, and I don't think it goes in the draft.

@martinthomson Please review this version.

bemasc:
I suspect that any detailed specification of how these state machines are integrated is likely to be unimplementable for some clients, so I would prefer to avoid excessive detail here.  I also think it's unnecessary: Alt-Svc always falls back to the origin, so apart from security issues, we don't have to standardize the corner cases.

That said, please suggest any extra text that you would like to see added.

draft-ietf-dnsop-svcb-https.md
ericorth: New text looks good to me.  Adds some nice clarity.

Some of the text on fallback to non-Alt-Svc gets a little bit close to impeding/redefining territory on RFC7838, but I think it's sufficiently on the safe side of the line, especially with the "as described in ..." reference.

draft-ietf-dnsop-svcb-https.md
ericorth: It is still my opinion that it would be very difficult to get any more specific on how to enforce matching origin ECH usage without getting way over-specified.

We shouldn't require fresh querying the origin HTTPS because the performance cost will be unacceptable to many clients.  And anything involving caching ECH state of previous connections or looking at what's going on in parallel connections from the same client would involve a lot of finicky details where the right thing to do really depends on what the client wants to do and details of how the client is implemented.

I think it is the right approach to ensure clients are capable of adding such enforcement if desired without mandating a specific approach, and I think the current text handles this well.

(And the overall main case is that an ECH-supporting server should have ECH config on all endpoints, origin and Alt-Svc, and then it is very clearly defined that ECH-supporting clients will use ECH.  All this other discussion is just regarding what to do in the very discouraged case where the domains have a very silly configuration of only offering ECH on some endpoints.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,46 @@ single-client granularity.
...
+    _8443._https.example.com. IN HTTPS 1 alt3.example. (
+        port=9443 alpn=h2,h3 ech=... )
+
+The client could then attempt an HTTP/3 connection to `alt3.example:9443`
+with ECH, or an HTTP over TLS connection to `alt.example:443` with ECH,
+as these options are consistent with both an Alt-Svc field value and its
+HTTPS record.  It would not connect to `alt2.example`, as there is no
+possible connection that is consistent with both ALPN values ("h2" from
+Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam triggers SVCB-reliant
+client behavior ({{ech-client-behavior}}).  Operators SHOULD avoid
```
martinthomson, line 1109: This reads like "the ech SvcParam triggers SVCB-reliant client behavior" is a reason for not choosing alt2.  But it can't be, because alt2 has a SVCB record.

Is there an error in the example?  Should alt2 not have the ech SvcParam, with the ech SvcParam on other records engaging the SVCB-reliant client behaviour?

draft-ietf-dnsop-svcb-https.md
davidben: Oh, my comment was more general: I wasn't sure what the text was trying to say in the first place. Whatever prescriptiveness we want here, the text should still convey that clearly! Intentionally trying not to be prescriptive isn't a reason to be unclear. :-)

I see the new text says something slightly looser. I read this to mean "if you have some out-of-band information that some origin supports ECH, you can go act on that". Is that the intent here, or was this meant to capture some other scenario?

bemasc: I'm not sure what you mean by "out of band", but that sounds basically right.  If you know the origin is FOO, and the alt-authority is not FOO, you MAY decide not to use the Alt-Svc, for any value of FOO.  How you know that the origin is FOO is not specified here, but presumably involves retrieving the origin's HTTPS record at some point.

In my view, this instruction is normatively null, because clients can always decide not to use the Alt-Svc: RFC 7838 says "By their nature, alternative services are OPTIONAL".  However, several comments on this PR have asked for text on client enforcement of ECH consistency, so this text serves as a reminder that it is allowed.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, adjusted.

draft-ietf-dnsop-svcb-https.md
ericorth: I wonder if maybe "impose their own consistency conditions" implies a specific reason/algorithm for doing this, when what we're really trying to do is clarify that clients can choose not to use any Alt-Svc names or any HTTPS endpoints within them for really whatever reason the client wants.  Would this be clearer: "Clients MAY impose their own consistency conditions on Alt-Svc connections, e.g. ..." -> "Clients MAY choose to disallow connections to any HTTPS endpoints, e.g. disallow endpoints without an "ech" SvcParam if the client is aware that the origin's HTTPS endpoints support ECH."

bemasc: OK, changed to

> As noted in Section 2.4 of
> {{AltSvc}}, clients MAY disallow any Alt-Svc connection according to their
> own criteria, e.g. disallowing Alt-Svc connections that lack ECH support
> if ECH is supported on the origin.

draft-ietf-dnsop-svcb-https.md
martinthomson: Oh, I completely missed the part where you connect based on the A/AAAA records.  Maybe again:

> an ECH-capable client would not attempt a direct h2 connection to "alt2.example" (that is, using the address listed in A/AAAA) because the "ech" SvcParam triggers SVCB-reliant client behavior ({{ech-client-behavior}})

If "direct" is a term of art, then the clarification might not be needed, but this is hardly "direct" by any reckoning familiar to me.

bemasc: OK, changed to

> An ECH-capable client would not connect to `alt2.example`,
> as there is no possible connection that is consistent with both ALPN values
> ("h2" from Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam prevents
> the client from falling back to a plain A/AAAA-based connection (i.e. the
> client is SVCB-reliant, {{ech-client-behavior}}).

Please review.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,47 @@ single-client granularity.
...
+    _8443._https.example.com. IN HTTPS 1 alt3.example. (
+        port=9443 alpn=h2,h3 ech=... )
+
+The client could then attempt an HTTP/3 connection to `alt3.example:9443`
+with ECH, or an HTTP over TLS connection to `alt.example:443` with ECH,
+as these options are consistent with both an Alt-Svc field value and its
+HTTPS record.  An ECH-capable client would not connect to `alt2.example`,
+as there is no possible connection that is consistent with both ALPN values
+("h2" from Alt-Svc, "h3" from SvcParams), and the "ech" SvcParam triggers
+SVCB-reliant client behavior ({{ech-client-behavior}}).  Operators SHOULD
```
davidben: This sentence ("An ECH-capable client [...]") seems to contain two steps at once. It's also misleading because non-ECH clients can also have see Alt-Svc vs ServiceForm incompatibilities. There's a difference between using ServiceMode records vs. A/AAAA if the alt2.example fetch gives back another TargetName. How about replacing:
```
    alt2.example.             IN HTTPS 1 . alpn=h3 ech=...
```
with:
```
    alt2.example.             IN HTTPS 1 alt2b.example. alpn=h3 ech=...
```

And then replace this sentence with:

> The client could not connect to `alt2b.example` because there is no possible connection that is consistent with both ALPN values ("h2" from Alt-Svc, "h3" from SvcParams). A SVCB-reliant client could not use the `alt2.example` Alt-Svc entry, but an SVCB-optional client could instead connect to `alt2.example` without the use of ServiceMode records. In an ECH-capable client, the "ech" SvcParam triggers SVCB-reliant client behavior ({{ech-client-behavior}}).

draft-ietf-dnsop-svcb-https.md
davidben: Hrm, I think this still doesn't quite capture it. The sentence ("An ECH-capable client would not connect [...]") seems to contain two steps at once. It's also misleading because non-ECH clients can also have see Alt-Svc vs ServiceForm incompatibilities. There's a difference between using ServiceMode records vs. A/AAAA if the alt2.example fetch gives back another TargetName. How about replacing:
```
    alt2.example.             IN HTTPS 1 . alpn=h3 ech=...
```
with:
```
    alt2.example.             IN HTTPS 1 alt2b.example. alpn=h3 ech=...
```

And then replace this sentence with:

> The client could not connect to `alt2b.example` because there is no possible connection that is consistent with both ALPN values ("h2" from Alt-Svc, "h3" from SvcParams). A SVCB-reliant client could not use the `alt2.example` Alt-Svc entry, but an SVCB-optional client could instead connect to `alt2.example` without the use of ServiceMode records. In an ECH-capable client, the "ech" SvcParam triggers SVCB-reliant client behavior ({{ech-client-behavior}}).

draft-ietf-dnsop-svcb-https.md
davidben: I think this is the least clear of the three so far. :-P How is the client supposed to know that ECH is supported on the origin, when Alt-Svc causes them not to query the origin?

It's especially problematic because the "ech" consistency requirement is merely a SHOULD rather than a MUST. At the end of the day, we need to define enough of what this means so that we can clearly say what is and isn't _required_ of servers to deploy ECH. The combination of merely making it a SHOULD and this clairvoyant client MAY means it's quite plausible to read this to mean that servers needn't worry about consistency since the client could just repair it.

davidben:
Oh gosh GitHub is the worst. It got confused and duplicated one of my comments, but in trying to delete the duplicate, it deleted my entire set of comments. Let me try to recreate them...

davidben:
Hrm, or maybe it didn't delete it, just the outer "code review" grouping?

draft-ietf-dnsop-svcb-https.md
ericorth: I'm fine with "direct" or "plain A/AAAA".  Either seems clear enough to me, but maybe my understanding has just been tainted by informal usage of "direct" for this meaning in these various discussions.  

Related note: If we don't like "direct" to refer to plain A/AAAA, might also want to reword Section 9.1: 

> The SVCB-optional client behavior specified in Section 3 permits clients to fall back to a direct connection if all SVCB options fail.

draft-ietf-dnsop-svcb-https.md
ericorth: > I think this is the least clear of the three so far. :-P How is the client supposed to know that ECH is supported on the origin, when Alt-Svc causes them not to query the origin?

I would do it either by adding more data to the Alt-Svc info cache to specify that the original connection (that provided the Alt-Svc header) used ECH, or I'd query the QUIC connection pool for a still-active connection to the origin and leave info there about whether or not ECH was used.  But I imagine this would all be very dependent on the client-specifics of how Alt-Svc info is stored and how other connections are kept track of, so I don't know if there's a reasonable way to get any more specific in the draft.  Overall, I assume any client that wanted to do something like this could find a way.

> It's especially problematic because the "ech" consistency requirement is merely a SHOULD rather than a MUST. At the end of the day, we need to define enough of what this means so that we can clearly say what is and isn't required of servers to correctly deploy ECH. The combination of merely making it a SHOULD and this clairvoyant client MAY means it's quite plausible to read this to mean that servers needn't worry about consistency since the client could just repair it.

The server requirements are only a SHOULD because servers are not required to support ECH and not required to have a reasonable configuration to ensure clients use ECH.

But if a server wants to do the right thing and have a good configuration, I think all the text is already pretty clear how to do that (by including "ech=" on all endpoints in the origin and Alt-Svc).  Section 9.2 states pretty clearly that an RRSet with mixed ECH configuration is vulnerable to downgrade attacks.  And the current draft text here says servers with ECH on the origin should also include it on Alt services or the client might make non-ECH-protected connections.

I think it would be getting redundant, but do you think we need a final clarifying "servers SHOULD NOT rely on clients doing this to avoid leaking unencrypted Client Hello" after the "clients MAY..."?

draft-ietf-dnsop-svcb-https.md
ericorth: Could break up the 2 things completely, and stop trying to contain the SVCB-reliant stuff just to the alt2 example.  One sentence about how the client could not use the HTTPS info from alt2.example (or state this via the alt2b.example stuff).  Then another sentence saying something along the lines of "For all 3 of these names, an ECH-capable client could not fall back to plain A/AAAA-based connection because the "ech" SvcParam triggers SVCB-reliant client behavior ({{...}})."

draft-ietf-dnsop-svcb-https.md
davidben: Well, I guess this goes back to what this text is trying to say in the first place. Again, the problem isn't the prescriptiveness but clarity. If the intent is about adding bits to the Alt-Svc cache, or looking at live connections (which are not guaranteed to be present), that doesn't check if "ECH is supported on the origin" but "if the origin was previously known to support ECH". If that's the intent (I'm still not sure, since the text is unclear), something like that may be better.

I don't think not being required to support ECH has any bearing on SHOULD/MUST. The text already only applies to "Origins that publish an 'ech' SvcParam in their HTTPS record". But, sure, a MUST may be impractical for other reasons. I'm not saying that it should be a MUST, but that the combination of the SHOULD and the client mitigation coming immediately afterwards in the same paragraph is misleading.

Perhaps simply remove it from the server paragraph, and instead integrate it into the paragraph about the client's rules?

draft-ietf-dnsop-svcb-https.md
davidben: Ah yeah, good point. The A/AAAA fallback case is relevant for both incompatible Alt-Svc/SVCB and plain connection error. I think that means we should also incorporate some more text from my original suggestion. The aim was not to be prescriptive about the order, but to translate these new notions of incompatibility and expanded Alt-Svc connection attempts into the existing notions of failure and fallback. I think we should spell out the general rule here.

`alt2b.example` isn't as important, but without it, there is no difference for an ECH-incapable client between the A/AAAA version and the ServiceForm version. Introducing a TargetName makes them different, which makes it clearer that this actually makes a difference.

draft-ietf-dnsop-svcb-https.md
davidben: > if the origin was previously known to support ECH

I guess that's also slightly tricky because "previously" encompasses some random stale fetch you did 5 months ago, even though you have more recent information. "if the most recent fetch of the origin's HTTPS record supported ECH".

Though, thinking about this some more, maybe we're better off just dropping this. If the problem is a stale Alt-Svc record, the next time you visit the origin, when you learn about ECH support, you'll also get a new Alt-Svc record:

>   When an Alt-Svc response header field is received from an origin, its
>   value invalidates and replaces all cached alternative services for
>   that origin.

If the problem is a server misconfiguration, that basically the same as this one, which we haven't suggested clients repair:

> An HTTPS RRSet containing some RRs with "ech" and some without is vulnerable to a downgrade attack. This configuration is NOT RECOMMENDED.

Removing the text also doesn't change whether it's allowed because:

> As noted in Section 2.4 of {{AltSvc}}, clients MAY disallow any Alt-Svc connection according to their own criteria

What we're doing by calling it out is emphasizing it. That is, we actually want to suggest clients do this. But given how unwilling we are to write down what "this" is, and the inconsistency with the other analogous error, it's unclear to me that's actually true.

Finally, on SHOULD vs. MUST, I'll note that we already have other text where we have MUST-level requirements to capture things that, if you fail to do, things will go wrong.

> When publishing a record containing an "ech" parameter, the publisher MUST ensure that all IP addresses of TargetName correspond to servers that have access to the corresponding private key or are authoritative for the public name.

draft-ietf-dnsop-svcb-https.md
bemasc: I added the `alt2b.example` TargetName and restructured this section as an enumeration instead of prose.  Please review.

draft-ietf-dnsop-svcb-https.md
bemasc: I broke this sentence into its own paragraph and changed the example to

> e.g. disallowing Alt-Svc connections that lack ECH support when there is an active ECH-protected connection for this origin.

That's a specific example of an allowed, plausible behavior, and I don't think it will give any server operator the impression that they can safely skip ECH on their alt-authority.

draft-ietf-dnsop-svcb-https.md
bemasc: I've restructured this section to avoid the need for this terminology.  Please review.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,58 @@ single-client granularity.
...
-Clients that implement support for ECH MUST perform the HTTPS RR query first,
-and MUST only make use of Alt-Svc when operating in SVCB-optional mode (see
-{{ech-client-behavior}}).
-
-This specification does not alter the DNS records used when connecting
-to an Alt-Svc hostname (typically A and/or AAAA only).
+Clients that implement support for both Alt-Svc and HTTPS records SHOULD
+retrieve any HTTPS records for the Alt-Svc alt-authority, and ensure that
+their connection attempts are consistent with both the Alt-Svc parameters
+and any received HTTPS SvcParams.  If present, the HTTPS record's target
```
enygren, line 1088: Is there a way to make this more clear that the consistency constraint
is between a selected Alt-Svc record and the selected HTTPS RR 
corresponding to the alt-authority?  The word "any" in 
"any received HTTPS SvcParams" makes this feel a little more 
ambiguous.   Maybe removing "any" would help?

enygren, line 1088:
```suggestion
and any received HTTPS SvcParams.  If present, the HTTPS record's TargetName
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,58 @@ single-client granularity.
...
+* HTTPS over TLS to `alt2.example:443` (Alt-Svc only)
+* HTTP/3 to `example.com:8443` (Alt-Svc only)
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
+clients might reveal the name of the server in an unencrypted ClientHello.
+Similar consistency considerations could apply to future SvcParamKeys, so
+alt-authorities SHOULD carry the same SvcParams as the origin unless
+a deviation is specifically known to be safe.
+
```
enygren, line 1132: Should we point out that there may be areas of transition where there may be inconsistencies?  For example, when origins are adding ECH?  (Due to the lack of an SVCB-Used or equivalent, the server returning the Alt-Svc has no way to know which version of an HTTPS RR was returned.  As such, there is may be mismatch during transitions.  This is probably OK, but clients MUST NOT break when there is a mismatch.

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
enygren: Perhaps append to "consistent with the Alt-Svc ALPN and parameters" :
" for the selected Alt-Svc record"
?

draft-ietf-dnsop-svcb-https.md
bemasc: Clients would have to go out of their way to create brokenness here, so I don't think we need this warning.

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
bemasc: OK, updated.

draft-ietf-dnsop-svcb-https.md
ericorth: As long as we're just talking about Alt-Svc, I think it's acceptable for a client to completely "break" and disallow the entire Alt-Svc for whatever reason the client chooses.  Just the nature of Alt-Svc as an optional mechanism.

enygren: Ignoring the Alt-Svc entirely in this case is fine, as long as the client falls back to the standard origin.  I guess that is clear in Alt-Svc so we don't need to spell it out here.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,58 @@ single-client granularity.
...
+    alt.example.              IN HTTPS 1 . alpn=h2,h3 ech=...
+    alt2.example.             IN HTTPS 1 alt2b.example. alpn=h3 ech=...
+    _8443._https.example.com. IN HTTPS 1 alt3.example. (
+        port=9443 alpn=h2,h3 ech=... )
+
+Based on these inputs, the following connection attempts would always be
+allowed:
+
+* HTTPS over TLS to `alt.example:443` (Alt-Svc and HTTPS record)
```
ericorth, line 1106: "HTTPS over TLS" -> "HTTP/2 over TLS"?

Also, I don't understand what "(Alt-Svc and HTTPS record)" means here.  Is it "(Consistent with both Alt-Svc and HTTPS record)"? Otherwise, I start wondering if this is trying to say something around SVCB vs direct connection.

Is it that

bemasc: It says "HTTPS over TLS" because this is about what connections the client may attempt, not about what ALPN the server will choose on that connection.

> Is it "(Consistent with both Alt-Svc and HTTPS record)"?

Yes.  I've changed the text to that phrasing.

ericorth: > It says "HTTPS over TLS" because this is about what connections the client may attempt, not about what ALPN the server will choose on that connection.

Good point, but in that case, shouldn't it be "TLS over TCP"? I believe that would be the more accurate description of the connection itself and is more consistent with the couple other places in the draft that refer to the connection type.

ericorth: (Or maybe "HTTPS over TCP"?)

bemasc: OK, changed to "HTTPS over TCP".

davidben (approved):
LGTM.

The "disallowing Alt-Svc connections that lack ECH support when there is an active ECH-protected connection for this origin" example seems like something that'll rare never fire, but whatever. I've probably quibbled here enough. :-) (If h2 or h3, why make a new connection when you've already got one? Seems the new connection cases are much more likely to be when you don't have a connection already open, but potentially some cached Alt-Svc values in play.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1082,18 +1082,60 @@ single-client granularity.
...
+* HTTP/3 to `example.com:8443` (Alt-Svc only)
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
+also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
+clients might reveal the name of the server in an unencrypted ClientHello.
+Similar consistency considerations could apply to future SvcParamKeys, so
+alt-authorities SHOULD carry the same SvcParams as the origin unless
+a deviation is specifically known to be safe.
+
+As noted in Section 2.4 of {{AltSvc}}, clients MAY disallow any Alt-Svc
```
davidben, line 1135: Minor nitpick, in the spirit of https://github.com/MikeBishop/dns-alt-svc/pull/331, this can be written `{{Section 2.4 of AltSvc}}` and the section will be linkified. It's *amazing*.

bemasc: Done.  Note to other contributors: this required me to update to the latest kramdown-rfc2629 (`sudo gem install kramdown-rfc2629`).

bemasc:
Thanks to everyone for helping to shape this section.  If there are no more comments in 24 hours, I'm going to regard this text as approved and publish a new draft version.

ericorth:
All looks good to me.

PR #334, opened 2021-08-12
Address review comments from Mark Nottingham

bemasc:
See https://mailarchive.ietf.org/arch/msg/dnsop/JIJ6qJ53k8k4YnP4MH-AQ6kQqMA/

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -776,21 +776,33 @@ alter their presentation or wire formats.
 The "alpn" and "no-default-alpn" SvcParamKeys together
 indicate the set of Application Layer Protocol Negotiation (ALPN)
 protocol identifiers {{!ALPN=RFC7301}}
-and associated transport protocols supported by this service endpoint.
+and associated transport protocols supported by this service endpoint (the
+"SVCB ALPN set").
 
-As with Alt-Svc {{AltSvc}}, the ALPN protocol identifier is used to
+As with Alt-Svc {{AltSvc}}, each ALPN protocol identifier is used to
 identify the application protocol and associated suite
 of protocols supported by the endpoint (the "protocol suite").
+The presence of an ALPN protocol in the SVCB ALPN set indicates that this
+service endpoint, described by TargetName and the other parameters (e.g.
+"port") offers service with the protocol suite associated with this ALPN
+protocol.
+
 Clients filter the set of ALPN identifiers to match the protocol suites they
 support, and this informs the underlying transport protocol used (such
-as QUIC-over-UDP or TLS-over-TCP).
+as QUIC-over-UDP or TLS-over-TCP).  ALPN protocol names that do not uniquely
+identify a protocol suite (e.g. an Identification Sequence that
+can be used with both TLS and DTLS) are not compatible with this
+SvcParamKey and MUST NOT be included in the SVCB ALPN set.
+
+### Representation
 
 ALPNs are identified by their registered "Identification Sequence"
 (`alpn-id`), which is a sequence of 1-255 octets.
 
     alpn-id = 1*255OCTET
 
-The presentation `value` SHALL be a comma-separated list ({{value-list}})
+For "alpn", the presentation `value` SHALL be
+a comma-separated list ({{value-list}})
 of one or more `alpn-id`s.  Zone file implementations MAY disallow the
 "," and "\\" characters instead of implementing the `value-list` escaping
 procedure, relying on the opaque key format (e.g. `key1=\002h2`) in the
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -776,21 +776,33 @@ alter their presentation or wire formats.
...
 protocol identifiers {{!ALPN=RFC7301}}
-and associated transport protocols supported by this service endpoint.
+and associated transport protocols supported by this service endpoint (the
+"SVCB ALPN set").
 
-As with Alt-Svc {{AltSvc}}, the ALPN protocol identifier is used to
+As with Alt-Svc {{AltSvc}}, each ALPN protocol identifier is used to
 identify the application protocol and associated suite
 of protocols supported by the endpoint (the "protocol suite").
+The presence of an ALPN protocol in the SVCB ALPN set indicates that this
```
MikeBishop, line 785:
```suggestion
The presence of an ALPN protocol identifier in the SVCB ALPN set indicates that this
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -776,21 +776,33 @@ alter their presentation or wire formats.
...
+service endpoint, described by TargetName and the other parameters (e.g.
+"port") offers service with the protocol suite associated with this ALPN
+protocol.
```
MikeBishop, line 788:
```suggestion
"port") offers service with the protocol suite associated with this ALPN identifier.
```
In both cases, there are no ALPN protocols, just identifiers that reference protocols.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -776,21 +776,33 @@ alter their presentation or wire formats.
...
+
 Clients filter the set of ALPN identifiers to match the protocol suites they
 support, and this informs the underlying transport protocol used (such
-as QUIC-over-UDP or TLS-over-TCP).
+as QUIC-over-UDP or TLS-over-TCP).  ALPN protocol names that do not uniquely
```
MikeBishop, line 792:
```suggestion
as QUIC-over-UDP or TLS-over-TCP).  ALPN protocol identifiers that do not uniquely
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1158,8 +1161,10 @@ the origin, not the TargetName.
 ## HTTP Strict Transport Security {#hsts}
 
 By publishing a usable HTTPS RR, the server operator indicates that all
-useful HTTP resources on that origin are reachable over HTTPS, similar to
-HTTP Strict Transport Security {{HSTS}}.
+useful HTTP resources on this Host ({{!I-D.draft-ietf-httpbis-semantics,
+Section 7.2}}) are reachable over HTTPS, similar to HTTP Strict Transport
```
MikeBishop, line 1165:
```suggestion
origins with an "http" scheme governed by this host ({{!I-D.draft-ietf-httpbis-semantics,
Section 4.3.2}}) have an equivalent origin with an "https" scheme ({{!I-D.draft-ietf-httpbis-semantics,
Section 4.3.3}}), similar to HTTP Strict Transport
```

I think Mark's point is that we're making a statement about mapping between one set of origins and another, not individual resources or a single origin.  The origin transformation rules imported from HSTS below reflect this.

MikeBishop (approved):

draft-ietf-dnsop-svcb-https.md
bemasc: I'm sure your terminology is an improvement, but I think we need to tweak this.

My concern with this phrasing is that "equivalent origin" suggests that the resources on both origins should be the same, but that's not actually what we want.  Ideally, the HTTP origin will be blank except for a redirect to HTTPS.  That's the rationale for the current language about "all useful HTTP resources".

Can you think of a way to represent this idea without implying the that origins have the same content?

moonshiner: "By publishing a usable HTTPS RR, the server operator indicates that all origins with an "http" scheme governed by this host have HTTPS resources have an origin with the correct "https" scheme similar to HTTP Strict Transport"

no that sucks too.   I find "equivalent" an issue as well after reading it over a few trimes.

MikeBishop: Seems like what we want is to communicate that you're going to end up with the same content either way.  That might be because the http:// endpoint serves content, or it might be because it sends you a 3XX telling you that you should have done https://.

If the issue is with the term "equivalent," how about "associated"?  The origins are cooperating, they will get you to the same content (though the mechanism may vary), but they may not return identical responses.

moonshiner: I like associated.

bemasc: I've tried to rewrite this, borrowing more text from the HSTS RFC and some of @MikeBishop's language.  It now reads

>An HTTPS RR directs the client to communicate with this host only over a
> secure transport, similar to HTTP Strict Transport Security {{HSTS}}.

...

> If the "http" origin exists and contains important resources that have no
> equivalent at the "https" origin, the operator MUST NOT publish an HTTPS RR.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1177,6 +1173,9 @@ the client SHOULD construct a corresponding "https" URL as follows:
 
 This construction is equivalent to {{Section 8.3 of HSTS}}, point 5.
 
+If the "http" origin exists and contains important resources that have no
```
MikeBishop, line 1176: I feel like "important" may be a sticking point here.  I infer what you mean, but there's no definition of "important" that someone could point to confirm that they are compliant with this MUST.  Is it sufficient to simply say "contains resources"?  The content of a 301 isn't a resource, it's merely the payload of a redirect.

bemasc: Here's another variation:

>If this redirection would result in a loss of functionality (e.g. important
resources that are only available on the "http" origin), the operator MUST
NOT publish an HTTPS RR.

enygren (approved):

PR #335, opened 2021-08-16
Respond to comments from Dale Worley

bemasc:


MikeBishop (approved):

draft-ietf-dnsop-svcb-https.md
```diff
@@ -189,7 +180,7 @@ intended destination to all entities along the network path.
 This document also defines a parameter for Encrypted ClientHello {{!ECH}}
 keys. See {{ech-param}}.
 
-## Terminology
+# Terminology
```
enygren, line 183:
```suggestion
## Terminology
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -395,8 +387,7 @@ resolvers SHOULD pick one at random.
 
 The primary purpose of AliasMode is to allow aliasing at the zone
 apex, where CNAME is not allowed.  In AliasMode, the TargetName will
-be the name of a domain that resolves to SVCB
-(or other SVCB-compatible record such as HTTPS),
+be the name of a domain that resolves to SVCB,
 AAAA, and/or A records.  The TargetName SHOULD NOT be equal
```
bemasc, line 391:
```suggestion
AAAA, and/or A records.  (See {{svcb-compatible}} for aliasing of SVCB-compatible RR types.) 
 The TargetName SHOULD NOT be equal
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -651,12 +648,10 @@ See {{incomplete-response}} for possible optimizations of this procedure.
...
 Recursive resolvers MUST be able to convey SVCB records with unrecognized
-SvcParamKeys, and MAY treat the entire
-SvcParams portion of the record as opaque.  No part of this specification requires
-recursive resolvers to alter their behavior based on its contents, even if the contents
-are invalid.  Recursive resolvers MAY validate the values of recognized
-SvcParamKeys and reject records containing values 
-which are invalid according to the SvcParam specification.
+SvcParamKeys, and MAY treat the entire SvcParams portion of the record as
+opaque, even if the contents are invalid.  Alternatively, recursive
+resolvers MAY report SERVFAIL to avoid returning a SvcParamValue that is
```
bemasc, line 654: @Habbie @vcunat This text change is based on the GENART review from Dale Worley.  Does this still work for your filtering requirements?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1720,19 +1734,17 @@ of three `item`s:
...
+|                                          |                                         |
+| ---------------------------------------- | --------------------------------------- |
+| **Mapped scheme**                        | "https"                                 |
+| **Other affected schemes**               | "http", "wss", "ws", (other HTTP-based) |
+| **RR type**                              | HTTPS (65)                              |
+| **Name prefix**                          | None for port 443, else `_$PORT._https` |
+| **Automatically Mandatory Keys**         | `port`, `no-default-alpn`               |
+| **SvcParam defaults**                    | `alpn`: \["http/1.1"\]                  |
+| **Special behaviors**                    | HTTP to HTTPS upgrade                   |
+| **Keys that clients must implement**     | None (if HTTP/1.1, default port only)   |
```
bemasc, line 1746: Maybe add "mandatory" here?

MikeBishop: That suggests a different type (which didn't specify "mandatory" here) would make "mandatory" an optional-to-implement param.  Which is... weird, verging on nonsensical.

In fact, I think this line is duplicative with "automatically mandatory keys."  Mandatory keys aren't required to be present in every record; a record is unusable if a mandatory type is present and not understood.  If a client MUST implement a key to be compliant with an RR type definition, how is that distinct from the set of automatically mandatory keys?

draft-ietf-dnsop-svcb-https.md
vcunat: Filtering?  Is there some context I've missed or forgot?  If it's filtering in sense of Response Policy Zones and similar, I don't see why consider it in this RFC.  That certainly isn't a case of "invalid according to the SvcParam's specification", and SERVFAIL isn't the usual filtering action (because caching).

Regardless of filtering, I wonder if SERVFAIL is the best thing to do in case the wrong record would be only non-mandatory (in the additional section).

bemasc: This is in reference to [this discussion](https://mailarchive.ietf.org/arch/msg/dnsop/DZKcP6tViymFl6axketzstO10i0/) from DNSOP.  I pinged you here because you commented on PR #313, which generated this text, and we want to make sure not regress on consensus.

vcunat: Ah, thanks... I confused the "filtering" meaning.  I think this new text is better.

Some implementations may like to check records to larger extent, some may not, but in any case extensibility needs to be kept in mind.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, removed.  I agree, this can be inferred from the other rows.

draft-ietf-dnsop-svcb-https.md
bemasc: I've relaxed the recommendation to use SERVFAIL specifically.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -763,6 +759,33 @@ in the initial response.  As a performance optimization, if some of the SVCB
...
+* RDATA presentation format
+* RDATA wire format
+* IANA registry used for SvcParamKeys
+* Authoritative server Additional Section processing
+* Recursive resolution process
+* Relevant Class (i.e. Internet ("IN") {{!RFC1035}})
+
+This allows authoritative and recursive DNS servers to apply identical
+processing to all SVCB-compatible RR types.
+
```
enygren, line 776: If we have a SVCB-compatible record like "NS2" that itself is used for recursive to authoritative logic, it's possible that this might not be identical in-terms of Recursive resolution process (or Auth server processing).  How do we allow for such in the future ?  Should those be "unless specified otherwise"?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -763,6 +759,33 @@ in the initial response.  As a performance optimization, if some of the SVCB
...
+All other behaviors described as applying to the SVCB RR also apply
+to all SVCB-compatible RR types unless explicitly stated otherwise.
+When following an AliasMode record of RR type `T` ({{alias-mode}}), the
+followup query to the TargetName MUST also be for type `T`.
+
+This document defines one SVCB-compatible RR type (other than SVCB itself):
+the HTTPS RR type ({{https}}), which avoids prefixes in order to improve
+compatibility with wildcards and CNAMEs, which are widely used with HTTPS.
```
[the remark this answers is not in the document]
enygren: Note in particular the s/HTTPS/HTTP/.
The new line after also helps readability.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -517,24 +509,30 @@ for a service, as performed by the client.  SVCB resolution is implemented as fo
...
 A client is called "SVCB-optional" if it can connect without the use of
 ServiceMode records, and "SVCB-reliant" otherwise.  Clients for pre-existing
 protocols (e.g. HTTPS) SHALL implement SVCB-optional behavior (except as
 noted in {{client-failures}} and {{ech-client-behavior}}).
 
 SVCB-optional clients SHOULD issue in parallel any other DNS queries that might
-be needed for connection establishment.  SVCB-optional clients SHALL append an
-alternative endpoint consisting of the final value of $QNAME, the authority
-endpoint's port number, and no SvcParams, to the list of alternative endpoints, which is
+be needed for connection establishment if the SVCB record is absent.
```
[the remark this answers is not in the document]
bemasc: OK, how about

```suggestion
be needed for connection establishment if the SVCB record is absent, in order to minimize delay
in that case and enable the optimizations discussed in {{optimizations}}.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -517,24 +509,30 @@ for a service, as performed by the client.  SVCB resolution is implemented as fo
...
+endpoint's port number, and no SvcParams.  (This endpoint will be
 attempted before falling back to non-SVCB connection modes.  This ensures that
 SVCB-optional clients will make use of an AliasMode record whose TargetName has
-A and/or AAAA records but no SVCB records.
+A and/or AAAA records but no SVCB records.)
+
+The client proceeds with connection establishment using the resolved list of
+endpoints.  Clients SHOULD try higher-priority alternatives first, with
+fallback to lower-priority alternatives.  Clients issue AAAA and/or A
+queries for the selected TargetName, and MAY choose between them using an
```
bemasc, line 531: Responding to https://github.com/MikeBishop/dns-alt-svc/pull/335#r723694121

It's true that these queries might not be "issued", but I'd rather avoid discussing caches here.  Maybe we can solve this with more precise language:
```suggestion
fallback to lower-priority alternatives.  Clients resolve AAAA and/or A
records for the selected TargetName, and MAY choose between them using an
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -517,24 +509,30 @@ for a service, as performed by the client.  SVCB resolution is implemented as fo
...
+endpoint consisting of the final value of $QNAME, the authority
+endpoint's port number, and no SvcParams.  (This endpoint will be
 attempted before falling back to non-SVCB connection modes.  This ensures that
 SVCB-optional clients will make use of an AliasMode record whose TargetName has
-A and/or AAAA records but no SVCB records.
+A and/or AAAA records but no SVCB records.)
+
+The client proceeds with connection establishment using the resolved list of
+endpoints.  Clients SHOULD try higher-priority alternatives first, with
+fallback to lower-priority alternatives.  Clients issue AAAA and/or A
```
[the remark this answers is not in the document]
bemasc: See below

draft-ietf-dnsop-svcb-https.md
```diff
@@ -763,6 +759,33 @@ in the initial response.  As a performance optimization, if some of the SVCB
...
+
+An RR type is called "SVCB-compatible" if it permits an implementation that is
+identical to SVCB in its:
+
+* RDATA presentation format
+* RDATA wire format
+* IANA registry used for SvcParamKeys
+* Authoritative server Additional Section processing
+* Recursive resolution process
+* Relevant Class (i.e. Internet ("IN") {{!RFC1035}})
```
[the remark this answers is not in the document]
bemasc: This section is about defining the word "SVCB-compatible", so the question is: is a definition that allows other classes more useful, or less useful?

I think it's probably less useful, because my goal is that resolvers and authorities should be able to add new SVCB-compatible RR types with a ~1 line change (just add another type number to the "SVCB-compatible types" list).  (And also because other classes no longer exist.)

draft-ietf-dnsop-svcb-https.md
bemasc: I think a SVCB-based type that requires some additional logic is "partially SVCB-compatible".  That may be a good choice in the future, but I don't think we need to define it here.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -763,6 +759,33 @@ in the initial response.  As a performance optimization, if some of the SVCB
...
+
+This allows authoritative and recursive DNS servers to apply identical
+processing to all SVCB-compatible RR types.
+
+All other behaviors described as applying to the SVCB RR also apply
+to all SVCB-compatible RR types unless explicitly stated otherwise.
+When following an AliasMode record of RR type `T` ({{alias-mode}}), the
+followup query to the TargetName MUST also be for type `T`.
```
bemasc, line 780: Responding to https://github.com/MikeBishop/dns-alt-svc/pull/335#r723713663

Is this clearer?
```suggestion
When following an AliasMode record ({{alias-mode}}) of RR type $T , the
followup query to the TargetName MUST also be for type $T.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -763,6 +759,33 @@ in the initial response.  As a performance optimization, if some of the SVCB
...
+* Authoritative server Additional Section processing
+* Recursive resolution process
+* Relevant Class (i.e. Internet ("IN") {{!RFC1035}})
+
+This allows authoritative and recursive DNS servers to apply identical
+processing to all SVCB-compatible RR types.
+
+All other behaviors described as applying to the SVCB RR also apply
+to all SVCB-compatible RR types unless explicitly stated otherwise.
+When following an AliasMode record of RR type `T` ({{alias-mode}}), the
```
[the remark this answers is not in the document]
bemasc: See below.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -990,16 +1013,6 @@ specific to the https and http schemes.  Clients MUST NOT
...
-The HTTPS RR wire format and presentation format are
-identical to SVCB, and both share the SvcParamKey registry.  SVCB
-semantics apply equally to HTTPS RRs unless specified otherwise.
-The presentation format of the record is:
-
-    Name TTL IN HTTPS SvcPriority TargetName SvcParams
-
-As with SVCB, the record is defined specifically within
-the Internet ("IN") Class {{!RFC1035}}.
-
```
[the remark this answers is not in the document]
bemasc: OK, I've reintroduced the zone file format, to jog the reader's memory.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -86,20 +86,11 @@ collection of consistent configuration parameters through the DNS
...
-to the HTTPS RR unless explicitly stated otherwise.
-{{https}} describes additional behaviors
-specific to the HTTPS RR.  Apart from {{https}}
-and introductory examples, much of this document refers only to the SVCB RR,
-but those references should be taken to apply to SVCB, HTTPS,
-and any future SVCB-compatible RR types.
+of services ({{svcb}}).  It also describes the rules for defining other
+SVCB-compatible RR types ({{svcb-compatible}}), starting with the HTTPS
+RR type ({{https}}), which provides improved efficiency and convenience
+with HTTPS by avoiding the need for an Attrleaf label {{?Attrleaf=RFC8552}}
```
enygren, line 92:
```suggestion
with HTTP by avoiding the need for an Attrleaf label {{?Attrleaf=RFC8552}}
```

Do we also want a normative reference to HTTP here?

bemasc: I don't think we need a reference.

PR #341, opened 2021-09-15
Adjust use of HTTPS vs. "https", etc.

### LINKED ISSUE #340: HTTPS as a protocol
The document pervasively refers to "HTTPS" as a protocol.  However, the current terminology appears to be that HTTP is a protocol, which can be used to access "http"-schemed and "https"-schemed URIs, as well as other schemes which might be transported.  We should do a pass over the document and ensure our use of HTTP/HTTPS/"http"/"https" is consistent with https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html.

bemasc:
Fixes #340

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -41,12 +41,12 @@ informative:
 
 This document specifies the "SVCB" and "HTTPS" DNS resource record (RR)
 types to facilitate the lookup of information needed to make connections
-to network services, such as for HTTPS origins.  SVCB records
+to network services, such as for "https" origins.  SVCB records
 allow a service to be provided from multiple alternative endpoints,
 each with associated parameters (such as transport protocol
 configuration and keys for encrypting the TLS ClientHello).  They also
 enable aliasing of apex domains, which is not possible with CNAME.
-The HTTPS RR is a variation of SVCB for HTTPS and HTTP origins.
+The HTTPS RR is a variation of SVCB for "https" and "http" origins.
 By providing more information to the client before it attempts to
 establish a connection, these records offer potential benefits to
 both performance and privacy.
@@ -67,9 +67,9 @@ default server, negotiating a preferred protocol, and providing relevant
 public keys.
 
 For example, when clients need to make a connection to fetch resources
-associated with an HTTPS URI, they currently resolve only A and/or AAAA
+associated with an "https" URI, they currently resolve only A and/or AAAA
 records for the origin hostname.  This is adequate for services that use
-basic HTTPS (fixed port, no QUIC, no {{!ECH=I-D.ietf-tls-esni}}), but
+basic HTTP/TLS (fixed port, no QUIC, no {{!ECH=I-D.ietf-tls-esni}}), but
 clients that learn additional information can gain privacy, performance, 
 and operational advantages.  It is highly desirable to minimize the 
 number of round-trips and lookups required to
@@ -87,7 +87,7 @@ collection of consistent configuration parameters through the DNS
 This document first describes the SVCB RR as a general-purpose resource
 record that can be applied directly and efficiently to a wide range
 of services ({{svcb}}).  The HTTPS RR is then defined as a special
-case of SVCB that improves efficiency and convenience for use with HTTPS
+case of SVCB that improves efficiency and convenience for use with HTTP
 ({{https}}) by avoiding the need for an Attrleaf label {{?Attrleaf=RFC8552}}
 ({{httpsnames}}).  Other protocols with similar needs may
 follow the pattern of HTTPS and assign their own
@@ -126,17 +126,17 @@ additional DNS RR in a way that:
   enables delegation of operational authority for an origin within the
   DNS to an alternate name.
 
-Additional goals specific to HTTPS RRs and the HTTPS use-case include:
+Additional goals specific to HTTPS RRs and the HTTP use-cases include:
 
-* Connect directly to HTTP3 (QUIC transport)
+* Connect directly to HTTP/3 (QUIC transport)
   alternative endpoints {{!HTTP3=I-D.ietf-quic-http}}
 * Obtain the Encrypted ClientHello {{!ECH}} keys associated with an
   alternative endpoint
 * Support non-default TCP and UDP ports
-* Enable SRV-like benefits (e.g. apex delegation, as mentioned above) for HTTP(S),
+* Enable SRV-like benefits (e.g. apex delegation, as mentioned above) for HTTP,
   where SRV {{?SRV=RFC2782}} has not been widely adopted
-* Provide an HSTS-like indication {{!HSTS=RFC6797}} signaling that the HTTPS
-  scheme should be used instead of HTTP for this request (see {{hsts}}).
+* Provide an HSTS-like indication {{!HSTS=RFC6797}} signaling that the "https"
+  scheme should be used instead of "http" for this request (see {{hsts}}).
 
 ## Overview of the SVCB RR
 
@@ -175,7 +175,7 @@ provide an extensible data model for describing alternative
 endpoints that are authoritative for the origin, along with
 parameters associated with each of these alternative endpoints.
 
-For the HTTPS use-case, the HTTPS RR enables many of the benefits of Alt-Svc
+For HTTP use-cases, the HTTPS RR enables many of the benefits of Alt-Svc
 {{?AltSvc=RFC7838}}
 without waiting for a full HTTP connection initiation (multiple roundtrips)
 before learning of the preferred alternative,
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -1180,7 +1183,7 @@ This construction is equivalent to {{Section 8.3 of HSTS}}, point 5.
 
 If an HTTPS RR query for this "https" URL returns any AliasMode HTTPS RRs,
 or any compatible ServiceMode HTTPS RRs (see {{mandatory}}), the client
-SHOULD act as if it has received an HTTP "307 Temporary Redirect" redirect
+SHOULD act as if it has received an HTTP 307 "Temporary Redirect"
 to this "https" URL.  (Receipt of an incompatible ServiceMode RR does not
```
MikeBishop, line 1187:
```suggestion
SHOULD behave as if it has received an HTTP 307 (Temporary Redirect) status code
with this "https" URL in the "Location" field.  (Receipt of an incompatible ServiceMode RR does not
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1190,7 +1193,7 @@ If this redirection would result in a loss of functionality (e.g. important
 resources that are only available on the "http" origin), the operator MUST
 NOT publish an HTTPS RR.
 
-When an HTTPS connection fails due to an error in the underlying secure
+When an "https" connection fails due to an error in the underlying secure
```
MikeBishop, line 1196:
```suggestion
When an HTTP connection fails due to an error in the underlying secure
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1557,13 +1560,13 @@ subregistry of the "Domain Name System (DNS) Parameters" registry:
...
 
-This document defines a new DNS RR type, HTTPS, whose value 65 has
+This document defines a new DNS RR type, "HTTPS", whose value 65 has
 been allocated by IANA from the "Resource Record (RR) TYPEs"
 subregistry of the "Domain Name System (DNS) Parameters" registry:
 
 * Type: HTTPS
 * Value: 65
-* Meaning: Service Binding for HTTPS
+* Meaning: Service Binding for "https"
```
MikeBishop, line 1569:
```suggestion
* Meaning: Service Binding for HTTP
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -41,12 +41,12 @@ informative:
 
 This document specifies the "SVCB" and "HTTPS" DNS resource record (RR)
 types to facilitate the lookup of information needed to make connections
-to network services, such as for HTTPS origins.  SVCB records
+to network services, such as for "https" origins.  SVCB records
```
MikeBishop, line 44:
```suggestion
to network services, such as for HTTP origins.  SVCB records
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -41,12 +41,12 @@ informative:
...
 allow a service to be provided from multiple alternative endpoints,
 each with associated parameters (such as transport protocol
 configuration and keys for encrypting the TLS ClientHello).  They also
 enable aliasing of apex domains, which is not possible with CNAME.
-The HTTPS RR is a variation of SVCB for HTTPS and HTTP origins.
+The HTTPS RR is a variation of SVCB for "https" and "http" origins.
```
MikeBishop, line 49:
```suggestion
The HTTPS RR is a variation of SVCB for use with HTTP {{!HTTP=I-D.ietf-httpbis-semantics}}.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -67,9 +67,9 @@ default server, negotiating a preferred protocol, and providing relevant
 public keys.
 
 For example, when clients need to make a connection to fetch resources
-associated with an HTTPS URI, they currently resolve only A and/or AAAA
+associated with an "https" URI, they currently resolve only A and/or AAAA
```
MikeBishop, line 70:
```suggestion
associated with a URI with an "https" or "http" scheme, they currently resolve only A and/or AAAA
```

MikeBishop (approved):

MikeBishop:
We may want @royfielding or @mnot to give this a once-over as well.

draft-ietf-dnsop-svcb-https.md
bemasc: This is an example, and the next part is specific to "https" origins, so I don't know if it makes sense to list both schemes here.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -979,14 +981,15 @@ This SvcParamKey is always automatically mandatory, and MUST NOT appear in its
 own value-list.  Other automatically mandatory keys SHOULD NOT appear in the
 list either.  (Including them wastes space and otherwise has no effect.)
 
-# Using SVCB with HTTPS and HTTP {#https}
+# Service Bindings for "https" {#https}
```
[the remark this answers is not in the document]
bemasc: My argument here is that a Service Binding mapping is specific to a scheme, not a protocol, and this section only defines a mapping for the "https" scheme.  Use of an "http" scheme with HTTPS or SVCB is actually forbidden!

Conversely, if you wanted to use some protocol that is not HTTP but somehow relies on the "https" scheme (maybe e.g. secure websockets?), this section would still apply.

draft-ietf-dnsop-svcb-https.md
bemasc: Given that HTTP connections don't generically have a secure transport, I find this version harder to understand.

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
MikeBishop: But you still have to do the lookup with "http"-schemed URIs, or you won't discover the HSTS-like directive.  So this does apply to all HTTP usage, just differently.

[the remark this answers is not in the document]
bemasc: It does apply to "http", but it doesn't create a service binding for "http".  The implications for "http" usage are a special behavior of the HTTPS RR type, separate from the Service Binding functionality, which is only for "https".

How about "Using Service Bindings with HTTP", or just "The HTTPS RR Type"?

draft-ietf-dnsop-svcb-https.md
bemasc: The next sentence is
```
This is adequate for services that use
basic HTTP/TLS (fixed port, no QUIC, no {{!ECH=I-D.ietf-tls-esni}}), but
clients that learn additional information can ...
```
Clients don't use HTTP/TLS when the URI has an "http" scheme, so this becomes a non sequitur ... unless we're talking about RFC 8164, which is another layer of complexity that I would prefer not to raise.

This is just supposed to be an example, so there's no need to make it comprehensive.  How about: "For example, when accessing an "https" URI, clients currently resolve only ..."?

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
MikeBishop: "Using Service Bindings with HTTP" sounds good.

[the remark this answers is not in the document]

draft-ietf-dnsop-svcb-https.md
MikeBishop: As before, I prefer the last option since it affects all HTTP requests.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, how about

> For example, HTTP clients currently resolve only A and/or AAAA records for the origin hostname, learning only its IP addresses.  If HTTP clients learn more information about the origin before connecting, they may be able to upgrade "http" URLs to "https", enable HTTP/3 or ECH, or switch to an operationally preferable endpoint.

MikeBishop: Works for me.

draft-ietf-dnsop-svcb-https.md

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1005,12 +1007,12 @@ HTTPS RRs.  The default set of ALPN IDs is the single value "http/1.1".
...
+Clients that restrict the destination port in "https" URIs
+(e.g. using the "bad ports" list from {{FETCH}}) SHOULD apply the
 same restriction to the "port" SvcParam.
 
 The presence of an HTTPS RR for an origin also indicates
-that all HTTP resources are available over HTTPS, as
+that clients should connect securely and use the "https" scheme, as
 discussed in {{hsts}}.  This allows HTTPS RRs to apply to
 pre-existing "http" scheme URLs, while ensuring that the client uses a
 secure and authenticated HTTPS connection.
```
enygren, line 1018:
```suggestion
secure and authenticated "https" scheme connection.
```

enygren (approved):

PR #344, opened 2021-09-27
Correct ALPN comparison

### LINKED ISSUE #343: ALPN text incorrectly implies something about Alt-Svc
> Unlike Alt-Svc Field Values, HTTPS RRs can contain multiple ALPN IDs, and clients are encouraged to offer additional ALPNs that they support.

From a grammatical perspective, the construction of this sentence is problematic.  It could mean:

1. HTTPS RRs differ from Alt-Svc in that they can convey multiple ALPN IDs with each entry.  Consequently, clients are encouraged to include multiple ALPN IDs when they attempt to make a connection.
2. HTTPS RRs differ from Alt-Svc in that they can convey multiple ALPN IDs with each entry.  They also differ in that clients are permitted to include additional ALPN IDs when they make a connection.

When it comes to the latter part, both interpretations are problematic.  Either way, it might be read to imply something that Alt-Svc (RFC 7838) does not say.  The relevant text in RFC 7838 says:

>  If the connection to the alternative service does not negotiate the expected protocol (for example, ALPN fails to negotiate h2, or an Upgrade request to h2c is not accepted), the connection to the alternative service MUST be considered to have failed.

This does not say anything about what clients offer when attempting to connect to a server, only what they ultimately accept.  However, the text first quoted here seems to imply that clients are somehow prohibited from offering additional ALPN IDs when connecting to an alternative service.

ALPN depends (critically) on clients representing the set of protocols that they are willing to use faithfully.  Implying that they might not, especially when using a medium with oft-questionable security properties to drive that decision, is a bad idea.

I think that this is simply editorial.  The procedures elsewhere in the document are good.  It's a little fiddly to get right, but I will attempt to propose a small change shortly that should fix this.

martinthomson:
This is the more verbose of the two options I considered.  It includes
more contextual information about what a client might choose to do with
HTTPS RRs.

In Alt-Svc, clients are required to consider an alternative "failed"
when the chosen ALPN is not negotiated by a server.  (This is something
I think we should remove, but I'll take that up in the revision.)  This
is a point of comparison with HTTPS RRs that is worth highlighting.

This could just say:

> Clients are not required to abort a connection attempt if a chosen
ALPN ID is not negotiated by the server.

The problem with that is that it doesn't really provide any information
on why that might be a problem.  Other text on the draft only really
talks about what to offer (which is absoltely the right thing to do).
That leaves readers with no context for the recommendation other than
text in RFC 7838.  This contextualizes it some.  It is perfectly
reasonable for clients to only use HTTPS RRs for what they perceive to
be protocol upgrades, using fallback behaviour for old protocols like
HTTP/1.1.

Closes #343.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1056,9 +1056,10 @@ differences in the intended client and server behavior.
 
 ### ALPN usage
 
-Unlike Alt-Svc Field Values, HTTPS RRs can contain multiple ALPN
-IDs, and clients are encouraged to offer additional ALPNs that they
-support.
+Unlike Alt-Svc Field Values, HTTPS RRs can contain multiple ALPN IDs.  Clients
+might choose to remove less-preferred ALPN IDs they offer when using HTTPS RRs,
+but they are not required to abort a connection attempt if a chosen ALPN ID is
+not negotiated by the server.
 
 ### Untrusted channel
 


draft-ietf-dnsop-svcb-https.md
```diff
@@ -1056,9 +1056,10 @@ differences in the intended client and server behavior.
 
 ### ALPN usage
 
-Unlike Alt-Svc Field Values, HTTPS RRs can contain multiple ALPN
-IDs, and clients are encouraged to offer additional ALPNs that they
-support.
+Unlike Alt-Svc Field Values, HTTPS RRs can contain multiple ALPN IDs.  Clients
+might choose to remove less-preferred ALPN IDs they offer when using HTTPS RRs,
+but they are not required to abort a connection attempt if a chosen ALPN ID is
+not negotiated by the server.
```
bemasc, line 1062:
```suggestion
Unlike Alt-Svc Field Values, HTTPS RRs can contain multiple ALPN IDs.
```

In my view, the text on ALPNs in RFC 7838 is ambiguous, and I would like to avoid attempting to clarify it in this draft.  (We can leave that for alt-svc-bis.)  I don't think "the client is required to abort if the indicated ALPN is not selected" is the only plausible reading of RFC 7838, and it's not the interpretation I prefer.

Since we don't seem to have agreement on the actual distinction between these drafts, I think we're best off simply removing this explanatory text.

enygren: I'm fine with Ben's latest proposal.  Martin?

martinthomson: Yes, I think that Ben nailed it.

enygren (approved):

PR #362, opened 2022-02-22
Make IPv4-mapped support optional and embedded support obligatory

### LINKED ISSUE #361: Why is IPv6 in IPv4-mapped format allowed?
`   example.com.   SVCB   1 example.com. ipv6hint="::ffff:198.51.100.100"` is allowed. I'd like to know the rationale behind this. Could you please give an example usecase?

DesWurstes:
Resolves #361

Looking forward for your comments @wtoorop

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -963,6 +963,9 @@ perform DNS64 ({{!RFC6147}}) on parameters within a SVCB record.
 For best performance, server operators SHOULD include an "ipv6hint" parameter
 whenever they include an "ipv4hint" parameter.
 
+For security reasons, clients MUST reject the RRSet if "ipv6hint" signifies
+an IPv4 mapped IPv6 address, such as "::ffff:198.51.100.100".
+
 These parameters are intended to minimize additional connection latency
 when a recursive resolver is not compliant with the requirements in
 {{server-behavior}}, and SHOULD NOT be included if most clients are using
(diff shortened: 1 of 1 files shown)


wtoorop:
> Resolves #361
> 
> Looking forward for your comments @wtoorop

:+1: for the corrected and added test vector.

Not sure about the added text in the _"ipv4hint" and "ipv6hint"_ section. I don't think you should just say "For security reasons" without saying what those reasons are. I suppose it would also be fine to just leave that text out. What do you think @DesWurstes ? Did you have a specific attack vector in mind?

davidben:
Agreed on "security reasons". But also, spelled out in the text or not, what is the motivation for requiring clients reject those values? I didn't think we did such things in, e.g., AAAA records. (@ericorth to confirm.)

bemasc:
Yeah, I don't think we should be trying to identify which addresses are and are not eligible to be included in a hint.  As much as possible, I think they should be just like `A` and `AAAA` records.  Clients might indeed reject certain values, but that rejection should not depend on whether the address arrived in an `A` record or an `ipv4hint`.

DesWurstes:
Updated. Now only embedded-v4 support is obligatory as @wtoorop . Or would you prefer to make it "SHOULD" as well?

bemasc (approved):

enygren:
Do we want an operational consideration recommending against using them?  Having IPv4-mapped addresses in AAAA records is a common operational mistake.

davidben:
Is there something about SVCB that changes the considerations here? No strong objections, but it seems to me we can assume this covered by existing stuff.

marka63:
Really there is nothing new here.  Nameservers already have to deal with mapped addresses in AAAA records.  If your acl / connection logic doesn't handle mapped address already it is broken.

enygren:
I'll mail v6ops for guidance.

bemasc:
I don't think this is specific to v4-mapped v6, or even v6 generally.  IPv4 and IPv6 both have various special ranges that probably should not appear in the DNS, but the DNS is nonetheless capable of conveying them.  I suspect that v4-mapped v6 is just one of many weird IP ranges that are commonly published due to an operational mistake.

DesWurstes:
Then I would suggest the following under ipv6hint:

> The same considerations apply as when interacting with AAAA: The behavior toward IPv4-mapped addresses in IPv6 syntax is implementation defined. Nevertheless IPv4-mapped addresses SHOULD NOT be used with ipv6hint instead of with ipv4hint.

Although these will be treated exactly the same way in practice AAAA is treated it's good practice to remind programmers.

bemasc:
I think this change is good as-is.  In very long documents such as this, adding more guidance that should be obvious can make it harder to find the guidance that is not so obvious.

enygren (approved):

PR #365, opened 2022-03-02
Editorial suggestions from Benjamin Kaduk's IESG review

kaduk:


draft-ietf-dnsop-svcb-https.md
```diff
@@ -946,10 +949,11 @@ geo-aware features and thereby degrade client performance.
 
 The presentation `value` SHALL be a comma-separated list ({{value-list}})
 of one or more IP addresses of the appropriate
-family in standard textual format {{!RFC5952}}.  To enable simpler parsing,
+family in standard textual format {{!RFC5952}}, {{!RFC4001}}.  To enable simpler parsing,
```
kaduk, line 952: 5952 only covers IPv6 textual format.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -984,7 +988,7 @@ mandatory", i.e. mandatory if they are present in an RR.  The SvcParamKey
 "mandatory" is used to indicate any mandatory keys for this RR, in addition to
 any automatically mandatory keys that are present.
 
-A ServiceMode RR is considered "compatible" with a client if the client
+A ServiceMode RR is considered "compatible" by a client if the client
```
kaduk, line 991: This one I'm not entirely sure about.
It is perhaps possible that it makes sense to think about a RR being "compatible with" a client, if the RR itself is the focus of the concept.  Yet the rest of the discussion here seems focused on client processing, and the phrasing is a little weird to me anyway, so I think this is probably better.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1004,7 +1008,7 @@ For example, the following is a valid list of SvcParams:
     ech=... key65333=ex1 key65444=ex2 mandatory=key65444,ech
 
 In wire format, the keys are represented by their numeric values in
-network byte order, concatenated in ascending order.
+network byte order, concatenated in strictly increasing numeric order.
```
kaduk, line 1011: we used the phrase "strictly increasing numeric order" earlier, so this seeks consistency.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1014,7 +1018,7 @@ list either.  (Including them wastes space and otherwise has no effect.)
 
 Use of any protocol with SVCB requires a protocol-specific mapping
 specification.  This section specifies the mapping for the "http" and "https"
-URI schemes {{!I-D.draft-ietf-httpbis-semantics}}.
+URI schemes {{!HTTP}}.
```
kaduk, line 1021: currently we list [HTTP] and [I-D.ietf-httpbis-semantics] as different reference entries.  I don't have a build toolchain set up, though, so I'm not 100% sure this is a complete fix.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1521,7 +1526,7 @@ with SVCB.
 
 # Security Considerations
 
-SVCB/HTTPS RRs are intended for distribution over untrusted
+SVCB/HTTPS RRs permit distribution over untrusted
```
kaduk, line 1529: just a couple paragraphs prior we said "intended for use over a privacy-preserving DNS transport", which is not exactly an untrusted channel (though it's also not a fully trusted channel).

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1899,7 +1906,7 @@ long, it is broken into several lines.
     \x02\x9b                                           # key 667
     \x00\x09                                           # length 9
     hello\xd2qoo                                       # value
-{: title="A generic key and quoted value with a decimal escape"}
+{: title="A generic key and quoted value with an octal escape"}
```
kaduk, line 1909: I'm not sure what "decimal escape" was intended to mean; this is a best guess

bemasc: Nope, it's actually a decimal!  RFC 1035 Section 5.1:

~~~
\DDD            where each D is a digit is the octet corresponding to
                the decimal number described by DDD.  The resulting
                octet is assumed to be text and is not checked for
                special meaning.
~~~

Zone files actually don't support hex, octal, or unicode escaping.

kaduk: Oops, reverted.
I ... am not really sure why my brain saw 1-255 and somehow got "octal" from that :-/

MikeBishop: I'd guess "resulting octet" got slurred in "resulting octal value."

draft-ietf-dnsop-svcb-https.md
```diff
@@ -370,7 +370,8 @@ any ServiceMode records in the set.
 RRSets are explicitly unordered collections, so the
 SvcPriority field is used to impose an ordering on SVCB RRs.
 SVCB RRs with a smaller SvcPriority value SHOULD be given
-preference over RRs with a larger SvcPriority value.
+preference over RRs with a larger SvcPriority value (the previous paragraph
+notwithstanding).
```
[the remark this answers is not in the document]
kaduk: My original thinking seems lost to the depths of time.
That said, it looks like both the "SHOULD have the same Mode" and "if AliasMode exists, MUST ignore ServiceMode" directives are potentially relevant here.  That is, the advice here remains good general advice, but is not grounds to ignore the specific directives already given.  In particular, ignoring this SHOULD (which is permitted) does not allow us to violate the previous MUST.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -370,7 +370,8 @@ any ServiceMode records in the set.
 RRSets are explicitly unordered collections, so the
 SvcPriority field is used to impose an ordering on SVCB RRs.
 SVCB RRs with a smaller SvcPriority value SHOULD be given
```
bemasc, line 372:
```suggestion
ServiceMode RRs with a smaller SvcPriority value SHOULD be given
```

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
bemasc: Does the proposed change above solve that problem?

bemasc (approved):

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
enygren: Incorporated Ben's proposed change.  With that can we remove the "(the previous paragraph notwithstanding)."

```suggestion
preference over RRs with a larger SvcPriority value.
```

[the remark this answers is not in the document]
kaduk: Apparently I missed this when it came in, sorry.  Yes, the now-merged fix is a good fix -- thanks!

draft-ietf-dnsop-svcb-https.md
```diff
@@ -946,10 +948,11 @@ geo-aware features and thereby degrade client performance.
...
 The presentation `value` SHALL be a comma-separated list ({{value-list}})
 of one or more IP addresses of the appropriate
-family in standard textual format {{!RFC5952}}.  To enable simpler parsing,
+family in standard textual format {{!RFC5952}}, {{!RFC4001}}.  To enable simpler parsing,
 this SvcParamValue MUST NOT contain escape sequences.
 
-The wire format for each parameter is a sequence of IP addresses in network
+The wire format for each parameter is a sequence of (the respective
+address-family) IP addresses in network
 byte order.  Like an A or AAAA RRSet, the list of addresses represents an
```
enygren, line 956:
```suggestion
The wire format for each parameter is a sequence of IP addresses in network
byte order (for the respective address-family).  
Like an A or AAAA RRSet, the list of addresses represents an
```

Moving the parenthetical to the end seems more readable?

draft-ietf-dnsop-svcb-https.md
enygren: I can also see this going either way.

bemasc: "compatible with" seems more natural to me, but on matters of comprehensibility I defer to the reader.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1123,7 +1126,8 @@ single-client granularity.
 
 ## Interaction with Alt-Svc
 
-Clients that implement support for both Alt-Svc and HTTPS records SHOULD
+Clients that implement support for both Alt-Svc and HTTPS records and
+receive an Alt-Svc header field in a response SHOULD
```
enygren, line 1130:
```suggestion
Clients that implement support for both Alt-Svc and HTTPS records and
are making a connection based on a cached Alt-Svc response SHOULD
```

Alt-Svc can come from an H2 frame not just a header.
The lookup also should come at request time, not at the time of receiving the Alt-Svc,
as these could be separated by hours or days.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1332,7 +1336,7 @@ The domain owner could add this record:
     @ 7200 IN HTTPS 1 . alpn=h3
 
 to indicate that https://simple.example supports QUIC
-in addition to TLS over TCP (the implicit default).
+in addition to HTTP/1.1 over TLS over TCP (the implicit default).
```
enygren, line 1339:
```suggestion
in addition to HTTP over TLS over TCP (the implicit default,
with HTTP/1.1 or HTTP/2 negotiated during the TLS handshake 
via ALPN).
```

It could be either as ALPN will negotiate.

draft-ietf-dnsop-svcb-https.md
enygren: makes sense

enygren (changes requested):
Looks generally good.  Some questions/comments still to resolve.

draft-ietf-dnsop-svcb-https.md
bemasc: Yeah, but the sentence is: "The domain owner could add this record ... to indicate that it ... supports ...".  The record indicates support for HTTP/3 and HTTP/1.1, not HTTP/2.  An HTTP/2-only would not attempt to use this record, because there is no indication that the endpoint supports HTTP/2.

enygren (approved):

PR #366, opened 2022-03-02
Highlight compatibility requirements for HSTS upgrade

bemasc:


tfpauly (approved):

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1036,7 +1037,7 @@ Clients that restrict the destination port in "https" URIs
 same restriction to the "port" SvcParam.
 
 The presence of an HTTPS RR for an origin also indicates
-that clients should connect securely and use the "https" scheme, as
+that clients should always connect securely and use the "https" scheme, as
```
kaduk, line 1040: "always" here is something I'm on the fence about.  I don't think it's safe to do a true HSTS-like pinning scheme based solely on an unauthenticated DNS record.  If DNSSEC is in play, then it looks better, but for regular HSTS we get the signal directly from the origin itself, and this is a more indirect mechanism that we may want to only use for the current batch of requests.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -127,7 +127,8 @@ Additional goals specific to HTTPS RRs and the HTTP use-cases include:
 * Enable SRV-like benefits (e.g. apex delegation, as mentioned above) for HTTP,
   where SRV {{?SRV=RFC2782}} has not been widely adopted
 * Provide an HSTS-like indication {{!HSTS=RFC6797}} signaling that the "https"
-  scheme should be used instead of "http" for this request (see {{hsts}}).
+  scheme should be used instead of "http" for all requests to this origin (see
+  {{hsts}}).
```
kaduk, line 131: An unqualified "origin" I would tend to treat as meaning "URI origin", but that is scheme-specific.  {{HTTP}} defines an "origin server" that may be closer to what we mean here.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1213,8 +1214,8 @@ Because HTTPS RRs are received over an often insecure channel (DNS),
 clients MUST NOT place any more trust in this signal than if they
 had received a 307 redirect over cleartext HTTP.
 If this redirection would result in a loss of functionality (e.g. important
-resources that are only available on the "http" origin), the operator MUST
-NOT publish an HTTPS RR.
+resources or behaviors that are only available when using the "http" origin),
+the operator MUST NOT publish an HTTPS RR.
```
kaduk, line 1218: If I held the pen I would have put more detail in here, but I don't have a particularly strong argument that more detail is clearly required.

kaduk (commented):
This seems overall okay, in that it shows that the authors/WG have thought about the topic I raised and considered it; as such, merging it would suffice to have me clear my DISCUSS (even though I would probably still comment that more could be said).
As a brainstorming exercise, I put an alternative strawman proposal at #374.
Since @martinthomson replied in email, I'd be interested to hear his thoughts here as well.

draft-ietf-dnsop-svcb-https.md
martinthomson: I would like to see this being more explicit.  If an action that was initiated for an HTTP resource is performed on an HTTPS resource instead, this might have unexpected consequences if the two URLs are not aliases of the same logical resource.

draft-ietf-dnsop-svcb-https.md
martinthomson: "always" is wrong.  It's only as long as the HTTPS resource records are within their TTL.

martinthomson (changes requested):
This is where you have to make a call, and I think that @kaduk is right to ask that we make this very consciously, because the call looks like this: are we OK with arbitrary third parties denying us access to HTTP websites?  A spoofed HTTPS RR is all it takes to lock in a commitment to HTTPS and deny a client access to an HTTP-only server.

The general concern that applies to this sort of commitment is that it leads to a denial of service.  If the commitment to provide a server (on HTTPS) lives longer than the service itself, clients will refuse to proceed (to HTTP).

We've enough operational experience with shipping HTTPS that shows it rarely needs to be rolled back.  HSTS makes the same commitment, so that part seems fine.

Where this differs from HSTS is that we essentially assume that DNS is unauthenticated.  Or at least for practical purposes, we cannot rely on it being authenticated enough that we would require clients to insist on having DNSSEC before they engage this behaviour.

Many of those clients that do not perform DNSSEC validation rely on their resolver to do that work, which leaves them vulnerable at a minimum to their resolver and often to on-path adversaries between them and their resolver.  And while some resolvers validate, not enough do.  Those that don't validate (or origins that don't sign) are open to spoofing of the DNS query to the authoritative.

All that doesn't matter that much to me.  What decides this for me is that denial of service is a better failure mode than having the same entities intercept cleartext HTTP, possibly surreptitiously.  An attacker in the local network that is in a position to spoof DNS resolution in this way can attack HTTP in far more sinister ways.

(The "request changes" is for the "always" addition, which is wrong.)

davidben:
Is the DoS any different from that same entity spoofing the wrong A/AAAA record for the site? The HTTPS-RR-triggered upgrade only lasts as long as the cached HTTPS record (TTL, DNS cache clearings, etc.), which seems to be the same power those same attackers would have with A/AAAA.

martinthomson:
Yep, it is no different from a spoofed (wrong) A/AAAA record.  This is also how you might jump from spoofing something outside the network (say the resolver to authoritative query) to an on-path attack.  Either way, the worst case for that attack is a full interception of an HTTP session.  Denying access to HTTPS is far less interesting if you can do that.

draft-ietf-dnsop-svcb-https.md
MikeBishop: IIRC, we've circled and redrawn this line several times.  The breaking point for saying that the resources must be identical is that that HTTP origin might return redirects to the HTTPS origin anyway -- in which case you're *not* finding the same things over HTTPS as over HTTP, but you're winding up at the same resource.  Hence describing it in terms of loss of functionality, rather than identity of resources or responses.

davidben: The upgrade works by simulating a redirect, so that also lines up nicely: publishing such a record, just like setting HSTS, means you're serving a redirect at the http origin. If you wanted to do that, do that. If you didn't want to do that, don't.

ericorth: Seems like the "need" for this language keeps coming up because we're afraid a domain operator will do something stupid and add the record when a redirect will break their website.  Do we really think anybody will add the record when the redirect is not desired (or at least anybody that has actually read through enough of the RFC to see the redirect happens, which should be roughly a superset of anybody reading whatever statement we put here)? If not, any additional details will just add confusion, so I'd be happy if we just left the current language or simplified it.

ericorth: And by leave the "current language", I'm fine with the text in this change or before.  I'm just against adding more on top of that to try to add any more details on when a redirect would break a domain.

martinthomson: @MikeBishop, my objection was to the extreme economy of the phrase "resources or behaviors".  I was just looking for this being a little clearer by making the "behaviors" bit a little more explicit.  Something like,

> Where the HTTP resource does not redirect to the HTTPS resource or reference the same underlying resource, as is commonplace, requests directed to the HTTP resource will be processed by the HTTPS resource with potentially unexpected results.

kaduk: > Seems like the "need" for this language keeps coming up because we're afraid a domain operator will do something stupid and add the record when a redirect will break their website

I would rather characterize my primary concern as that the description in the document should accurately reflect its technical contents.  We are deviating from what the HTTP protocol specifies, and should clearly document the nature of the deviation.  (Ideally, document the motivation for the deviation as well.)

> which should be roughly a superset of anybody reading whatever statement we put here

Right, which is why I want to make it more visible in the Introduction that there are scenarios where the mechanism is not applicable (see #374 for thoughts on how)

> my objection was to the extreme economy of the phrase "resources or behaviors"

I think that also describes my concerns.

MikeBishop: @martinthomson, I like that text.

draft-ietf-dnsop-svcb-https.md
bemasc: "Origin server" doesn't seem to quite capture this either.  Changed to "host and port".

draft-ietf-dnsop-svcb-https.md
bemasc: I meant "always" in the sense of "unconditionally", but evidently this is confusing.  Removed.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, I've incorporated Martin's text.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1211,10 +1212,10 @@ with this "https" URL in the "Location" field.  (Receipt of an incompatible Serv
...
 Because HTTPS RRs are received over an often insecure channel (DNS),
 clients MUST NOT place any more trust in this signal than if they
-had received a 307 redirect over cleartext HTTP.
-If this redirection would result in a loss of functionality (e.g. important
-resources that are only available on the "http" origin), the operator MUST
-NOT publish an HTTPS RR.
+had received a 307 redirect over cleartext HTTP.  Where the "http" resource does
+not redirect to the "https" resource or reference the same underlying resource,
+as is commonplace, requests directed to the "http" resource will be processed by
+the "https" resource with potentially unexpected results.
```
enygren, line 1218:
```suggestion
had received a 307 redirect over cleartext HTTP.  

Publishing an HTTPS RR has the potential to have unexpected results 
or a loss in functionality in cases where the "http" resource neither
redirects to the "https" resource nor reference the same underlying resource.
This is because requests directed to the "http" resource will be processed by
the "https" resource.
```

Rewording to be more clear.  I'm also not sure we should talk about "commonplace" 
since there may also be commonplace cases in the other direction.

I'm not sure if we want to give operator guidance.?

enygren (changes requested):
Proposed some alternate wording that puts the risk up-front.

(Note that this is going to be a common challenge --- as an example, 
CDNs will likely need to provide customer choice over the presence of 
HTTPS RRs as unless HSTS or redirects are configured there is no way
to know that breakage would occur.)

That might be something we could also include as another way of framing this
(ie, encouraging operational care if the origin isn't known to be using redirects
or HSTS already.)

draft-ietf-dnsop-svcb-https.md
bemasc: I'm not sure the last sentence is a correct description of HTTP redirects.  I would just remove it.

In general, I think this is all extremely obvious, so I don't know how to clarify it.

enygren: How about this?

```suggestion
had received a 307 redirect over cleartext HTTP.  

Publishing an HTTPS RR has the potential to have unexpected results 
or a loss in functionality in cases where the "http" resource neither
redirects to the "https" resource nor references the same underlying resource.
```

bemasc: That's fine with me.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1212,10 +1212,11 @@ with this "https" URL in the "Location" field.  (Receipt of an incompatible Serv
 trigger the redirect behavior.)
 Because HTTPS RRs are received over an often insecure channel (DNS),
```
MikeBishop, line 1213:
```suggestion
Because HTTPS RRs are received over an often-insecure channel (DNS),
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1212,10 +1212,11 @@ with this "https" URL in the "Location" field.  (Receipt of an incompatible Serv
...
 clients MUST NOT place any more trust in this signal than if they
-had received a 307 redirect over cleartext HTTP.  Where the "http" resource does
-not redirect to the "https" resource or reference the same underlying resource,
-as is commonplace, requests directed to the "http" resource will be processed by
-the "https" resource with potentially unexpected results.
+had received a 307 redirect over cleartext HTTP.
```
MikeBishop, line 1215:
```suggestion
had received a 307 (Temporary Redirect) over cleartext HTTP.
```

MikeBishop (approved):
Nits are incidental to the point of this PR.

enygren (approved):

tungmeoo (commented):
good

PR #374, opened 2022-03-04
Add more details about HSTS scenarios

kaduk:
A strawman alternative to #366 to illustrate some topics that we may or may not want to cover.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1223,7 +1223,7 @@ sites either only provide "https" access or offer equivalent content across
 the two schemes, and the benefits of providing a signal to always use TLS in
 the common case outweigh the cost of some rare incompatibilities.
 Accordingly, if the redirection from http to https would result in a loss of
-functionality (e.g., there are important resources or behaviorsthat are only
+functionality (e.g., there are important resources or behaviors that are only
 available on the "http" origin), the operator MUST NOT publish an HTTPS RR.
 
 Note that in contrast to {{HSTS}}, the signal from the HTTPS RR is not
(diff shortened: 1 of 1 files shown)


kaduk:
The revisions in #366 do a decent job at explaining the conditions when the HSTS behavior has bad side effects and what those side effects are.  What I see as the key differentiating factors of this PR over that one is that it (1) clearly calls out in the introduction that there exist (rare) cases where the HSTS behavior renders the entire functionality unusable for a given service, and (2) attempts to justify the design decision to provide a protocol mechanism that is unusable for a minority of services.  These are both aspects that I see as part of responsible protocol design and specification, and I do feel that it is important to provide them in the specification.
That said, as I commented early on in #366, it is factually accurate and does provide clear documentation of the protocol's behavior, so I would not have a DISCUSS-level objection to the document once it is merged, only a COMMENT-level objection.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1212,9 +1214,23 @@ trigger the redirect behavior.)
...
+the common case outweigh the cost of some rare incompatibilities.
+Accordingly, if the redirection from http to https would result in a loss of
+functionality (e.g., there are important resources or behaviors that are only
+available on the "http" origin), the operator MUST NOT publish an HTTPS RR.
+
+Note that in contrast to {{HSTS}}, the signal from the HTTPS RR is not
+always received in-band from the authoritative origin, and when DNSSEC is
+not used may be tampered with in transit.  So the "strict transport
+security" property of the HTTPS RR should only apply to the current https
+request unless it is authenticated in some other way.
```
bemasc, line 1233: This sentence opens up the possibility of extending the HSTS behavior in some unspecified fashion if the DNS record is authenticated, which strikes me as very strange.  Also, "the current https request" is not really a defined thing.

If you think this needs exposition, I think it makes more sense to focus on why this is safe, rather than why it wouldn't be safe to do it some other way.

bemasc (commented):
I'm happy to add more discussion of the design rationale, if you think it's helpful and not obvious.  Let's merge #366 and then rebase this PR on top for further discussion.

enygren:
It seems like #366 as-merged covers most of this.  Do we need this extra level of detail or can we skip/close this?

bemasc:
@kaduk Please let us know if you want to revise this text given the merge conflict.  Otherwise, I think we'll stick with the (updated) text and work to finish the new draft.

enygren:
Rejecting/closing as the draft is post IESG review and in the RFC editor queue.
Any future changes need to happen in a subsequent draft 
or need a very very compelling reason to re-open the dumpster of worms.

PR #375, opened 2022-04-05
Avoid any ambiguity regarding handling wildcards

### LINKED ISSUE #356: Make sure we have the right meaning of "." with wildcard owner names
When TargetName is ".", and the record's owner name is a wildcard, it's clear what we want this to mean, but I'm not sure the text formally captures it.

bemasc:
Resolves #356

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -472,6 +472,7 @@ without the use of SVCB.
 
 For ServiceMode SVCB RRs, if TargetName has the value ".", then the
 owner name of this record MUST be used as the effective TargetName.
+For wildcard owner names, the synthesized owner name is used.
 
 For example, in the following example "svc2.example.net"
 is the effective TargetName:
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -472,7 +472,8 @@ without the use of SVCB.
 
 For ServiceMode SVCB RRs, if TargetName has the value ".", then the
 owner name of this record MUST be used as the effective TargetName.
-For wildcard owner names, the synthesized owner name is used.
+If the record has a wildcard owner name in the zone file, the recipient
+SHALL use the response's synthesized owner name as the effective TargetName.
 
 For example, in the following example "svc2.example.net"
 is the effective TargetName:


draft-ietf-dnsop-svcb-https.md
```diff
@@ -472,6 +472,7 @@ without the use of SVCB.
 
 For ServiceMode SVCB RRs, if TargetName has the value ".", then the
 owner name of this record MUST be used as the effective TargetName.
+For wildcard owner names, the synthesized owner name is used.
```
enygren, line 475:
```suggestion
Wildcard owner names only exist in zone files, so clients will see
and use the synthesized owner name.
```

I think we need more context for readability.

enygren (changes requested):

draft-ietf-dnsop-svcb-https.md
bemasc: But is this true with DNSSEC?

moonshiner: Yes, I believe it is.  And an example for more context may help.

bemasc: In general, validating DNSSEC "clients" (what's a client?) are fully aware of wildcard names, as wildcard status is explicit in DNSSEC.  Otherwise, how would it be possible to statically sign a zone containing wildcards?  So I find the idea that "wildcard owner names only exist in zone files" a bit confusing.

enygren: I guess they do leak out some in DNSSEC in ways I'm not an expert in, so maybe there is some other phrasing.  It's that wildcard owner names are used neither in queries nor in SVCB RR responses (unless literally
queried for).  (The only time the wildcard might show up would be in the DNSSEC signature and in logic around NSEC/NSEC3.)  We might want to reference rfc4592 ?

But as such, perhaps my proposal is also confusing just in different ways.
But I'm not sure that the original proposed wording is right either, and it is fairly confusing
from a client implementor perspective as they will generally never see the wildcard.

moonshiner: *.centralbit.com is a cname to the datatracker.ietf.org and the domain is signed.  Only way I can get the wildcard out of a resolver is by querying for the wildcard. I may be missing something.

bemasc: The validator [reconstructs the wildcard label](https://medium.com/nlnetlabs/the-peculiar-case-of-nsec-processing-using-expanded-wildcard-records-ae8285f236be#:~:text=This%20is%20an,is%20*.nlnetlabs.nl.) during RRSIG validation.

I've expanded the description here to try to capture @enygren's clarification, and added a normative SHALL.

enygren (approved):
I think this looks fine unless we also want an example for the wildcard case?  I'm not sure that's needed...

bemasc:
I'm having difficulty thinking of a good example here because I'm not sure how this would actually be used.  I'm inclined to leave it out unless someone asks.

PR #376, opened 2022-04-08
Adjust uses of SHOULD

### LINKED ISSUE #355: Overuse of should/SHOULD
See this mailing list thread which recommends that we convert some SHOULD to MUST and which also points out some uses of "should" that would be better to fix:
https://mailarchive.ietf.org/arch/msg/dnsop/EhP4nvJV1WEa4wiEIPPZIdLvIhc/

bemasc:
Resolves #355

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -226,11 +226,11 @@ for the case of "https" origins as described in {{https}}.
 
 SVCB RRs are extensible by a list of SvcParams, which are pairs consisting of a
 SvcParamKey and a SvcParamValue. Each SvcParamKey has a presentation name and a
-registered number. Values are in a format specific to the SvcParamKey. Their
-definition must specify both their presentation format (used in zone files) and
+registered number. Values are in a format specific to the SvcParamKey.  Each
+SvcParam has a specified presentation format (used in zone files) and
 wire encoding
 (e.g., domain names, binary data, or numeric values). The initial SvcParamKeys
-and formats are defined in {{keys}}.
+and their formats are defined in {{keys}}.
 
 ## Zone file presentation format {#presentation}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -227,7 +227,7 @@ for the case of "https" origins as described in {{https}}.
 SVCB RRs are extensible by a list of SvcParams, which are pairs consisting of a
 SvcParamKey and a SvcParamValue. Each SvcParamKey has a presentation name and a
 registered number. Values are in a format specific to the SvcParamKey. Their
-definition should specify both their presentation format and wire encoding
+definition must specify both their presentation format and wire encoding
```
MikeBishop, line 230: Is a different lower-case keyword an improvement here?  Given that the rest of this section is descriptive rather than normative, perhaps "SvcParam definitions specify...."

draft-ietf-dnsop-svcb-https.md
```diff
@@ -424,7 +426,7 @@ request.  This limit MUST NOT be zero, i.e. implementations MUST be able to
 follow at least one AliasMode record.  The exact value of this limit
 is left to implementations.
 
-For compatibility and performance, zone owners SHOULD NOT configure their zones
+For compatibility and performance, zone owners should not configure their zones
```
MikeBishop, line 429: I feel like lower-casing the keyword doesn't really address the issue here.  This is a legitimate recommendation.  Either we keep the SHOULD NOT -- but when might someone want to do this? -- or we rephrase this as a statement of fact:  "A zone that requires following multiple AliasMode records could encounter compatibility and performance issues."

MikeBishop (approved):

draft-ietf-dnsop-svcb-https.md

draft-ietf-dnsop-svcb-https.md

MikeBishop (approved):

PR #377, opened 2022-04-08
Clarify expectation for HSTS and proxies

### LINKED ISSUE #359: Interaction between HTTPS upgrade and HTTP(S) proxies is unclear
How and whether HTTP(S) proxies interact with SVCB/HTTPS records is a little unclear. On the one hand, they do speak HTTP, and it would be nice for for SVCB/HTTPS-based features to apply. ECH is generically applicable, and this allows proxies to negotiate HTTP/3 cleanly.

But, HTTP(S) proxies are weird. [draft-ietf-httpbis-semantics-latest](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#routing.proxy) says:
> Proxy configuration is implementation-dependent, but [...], and the proxy itself is usually identified by an "http" or "https" URI.

This doesn't quite match Chrome's implementation. Not sure about others, though PAC scripts use a [custom non-URL serialization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_PAC_file#return_value_format). While HTTP(S) proxies only sort of have URLs. Only the scheme, host, and port are meaningful. And the host is used differently from normal HTTP(S) URLs because the proxy is never sent its own name in the `Host` header, only that of the origin URL.

Still, scheme, host, and port is enough to construct a SVCB/HTTPS query name. If we do that, most features naturally apply, except the upgrade. The spec says:
> the client SHOULD behave as if it has received an HTTP 307 (Temporary Redirect) status code with this "https" URL in the "Location" field
https://github.com/MikeBishop/dns-alt-svc/blob/main/draft-ietf-dnsop-svcb-https.md#http-strict-transport-security-hsts

At least in Chrome's implementation, HTTP(S) proxies do not redirect:
* When the origin URL is an http:// URL and we get just send `GET http://example.com/whatever HTTP/1.1` to the proxy, there is no way for the proxy to utter a redirect at all. Any redirect is interpreted as coming from the origin.
* When the origin URL is an https:// URL, or we're doing WebSockets, we use `CONNECT example.com:443`, which is at least less ambiguous. But (at least in Chrome; not sure about other impls and I can't find spec text) a 3xx response in response to a `CONNECT` is an error. Indeed it's unclear what it'd mean since the path component isn't meaningful.

This seems problematic. The former is clearly wrong here. I'd argue the latter is wrong too, since it means deploying an HTTPS record breaks any clients with a stale config, rather than fixing them up.

Not sure what the right text is here. I can see the following options:
1. Switch it to an HTTPS proxy via non-redirect-based prose. This matches the spirit of the upgrade, but since proxies don't currently have a notion of redirects, this entails defining a whole new operation for proxies.
2. Don't apply this upgrade for HTTP proxies. One could argue that, unlike URLs in a browser's address bar, or links in hand-authored HTML, proxies are less likely to have the "wrong" scheme, and maybe this is less important. But it's somewhat disappointing.
3. Make it impl-defined, since so much of proxies are impl-defined anyway. ((2) is mostly the same as saying the implementation supports SVCB/HTTPS for HTTPS proxies but not HTTP proxies.)

Whichever we do, the existing text, if it applies to proxies (does it?), doesn't seem quite right.

bemasc:
Resolves #359

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1236,9 +1236,9 @@ opening handshake from the W3C Fetch specification {{FETCH}} SHOULD use HTTPS RR
 for the `requestURL`.
 
 When HTTP is used in a context where URLs or redirects are not applicable
-(e.g. Proxy Auto-Configuration), clients MUST implement transport security
-equivalent to the HSTS behavior specified in {{hsts}} or forego the use of
-HTTPS records.
+(e.g. connections to an HTTP proxy), clients that find a corresponding HTTPS RR
+SHOULD implement a security upgrade behavior equivalent to the one specified in
+{{hsts}}.
 
 Such protocols MAY define their own SVCB mappings, which MAY
 be defined to take precedence over HTTPS RRs.
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1228,11 +1228,17 @@ Section 12.1 of HSTS}}.
 
 ## Use of HTTPS RRs in other protocols
 
-All protocols employing "http://" or "https://" URLs SHOULD respect HTTPS RRs.
-For example, clients that
+All HTTP connections to named origins are eligible to use HTTPS RRs, even
+when HTTP is used as part of another protocol or without an explicit HTTP
+URL.  For example, clients that
 support HTTPS RRs and implement the altered WebSocket {{!WebSocket=RFC6455}}
 opening handshake from the W3C Fetch specification {{FETCH}} SHOULD use HTTPS RRs
 for the `requestURL`.
+
+When HTTP is used in a context where URLs or redirects are not applicable
+(e.g. Proxy Auto-Configuration), clients MUST implement transport security
+equivalent to the HSTS behavior specified in {{hsts}} or forego the use of
+HTTPS records.
 
 Such protocols MAY define their own SVCB mappings, which MAY
 be defined to take precedence over HTTPS RRs.


bemasc:
@davidben @ericorth Please review

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1228,12 +1228,18 @@ Section 12.1 of HSTS}}.
...
-For example, clients that
+All HTTP connections to named origins are eligible to use HTTPS RRs, even
+when HTTP is used as part of another protocol or without an explicit HTTP
+URL.  For example, clients that
 support HTTPS RRs and implement the altered WebSocket {{!WebSocket=RFC6455}}
 opening handshake from the W3C Fetch specification {{FETCH}} SHOULD use HTTPS RRs
 for the `requestURL`.
 
+When HTTP is used in a context where URLs or redirects are not applicable
+(e.g. Proxy Auto-Configuration), clients MUST implement transport security
```
davidben, line 1239: This is confusing because the PAC fetch itself is just regular URL-based http/https fetches. Redirects work naturally there.

What the bug refers to isn't the PAC fetch, but the proxy config itself. The proxy config may have been the output of a PAC script (which I mention in the bug because it's one of the few places where proxy descriptors are part of some interoperable thing, rather than some implementation-specific scheme). It's of interest because it doesn't specify a URL. But only sort of because, if you did specify a URL, redirects are still not applicable.

Perhaps just "e.g. connections to an HTTP(S) proxy".

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1228,12 +1228,18 @@ Section 12.1 of HSTS}}.
...
+equivalent to the HSTS behavior specified in {{hsts}} or forego the use of
+HTTPS records.
```
davidben, line 1241: What does it mean to forego the use of HTTPS records, since the only use of HTTPS records in these context is the upgrade itself?

That is, in the context of an HTTPS proxy, the HSTS behavior is implemented by doing nothing, because you're already on HTTPS. This clause then never applies. In the context of an HTTP proxy, you may or may not implement the HSTS behavior... but since HTTPS records are, well, for HTTPS, there isn't any other use of HTTPS records to forego.

draft-ietf-dnsop-svcb-https.md
bemasc: OK, reworded.

draft-ietf-dnsop-svcb-https.md
bemasc: Reworded

davidben (approved):

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1228,12 +1228,18 @@ Section 12.1 of HSTS}}.
...
+All HTTP connections to named origins are eligible to use HTTPS RRs, even
+when HTTP is used as part of another protocol or without an explicit HTTP
+URL.  For example, clients that
 support HTTPS RRs and implement the altered WebSocket {{!WebSocket=RFC6455}}
 opening handshake from the W3C Fetch specification {{FETCH}} SHOULD use HTTPS RRs
 for the `requestURL`.
 
+When HTTP is used in a context where URLs or redirects are not applicable
+(e.g. connections to an HTTP proxy), clients that find a corresponding HTTPS RR
+SHOULD implement a security upgrade behavior equivalent to the one specified in
```
ericorth, line 1240: I find this vague to the point of being confusing.  I've got RFC6797 open right now, and looking through it, I just don't know what this sentence is actually saying I should do.  That spec doesn't clearly describe what should be done in such cases.

Or is this one of those cases where the spec is being intentionally vague to let the clients do what they want to do as long as it's some reasonable behavior similar/same as what they do for HSTS? Maybe that's reasonable since this is a SHOULD.

bemasc: > I find this vague to the point of being confusing. I've got RFC6797 open right now, and looking through it, I just don't know what this sentence is actually saying I should do. That spec doesn't clearly describe what should be done in such cases.

This text renders as "... equivalent to the one specified in Section 9.5.", so it's not directly a reference to RFC 6797.  Does that help?

> Or is this one of those cases where the spec is being intentionally vague to let the clients do what they want to do as long as it's some reasonable behavior similar/same as what they do for HSTS? Maybe that's reasonable since this is a SHOULD.

Yes, it's deliberately vague, but if it's confusing that's still bad.

ericorth: > This text renders as "... equivalent to the one specified in Section 9.5.", so it's not directly a reference to RFC 6797. Does that help?

Ah, yes, that helps quite a bit.  I still think I would be a little bit confused on what clients should actually do if I were coming to this without all the context of our previous discussions since 9.5 mostly just talks about doing a redirect, but I don't have any ideas on how to do better without getting over-specified.  So I guess LGTM unless somebody has any better ideas.

davidben: Perhaps we should just define it for the two cases in question? (I can only think of WebSockets and proxies.) Although it's messy because HTTP doesn't do a great job of defining proxies, and the IETF and WHATWG disagree on how to define WebSockets opening. (In WHATWG and browsers, it seems WebSockets cannot redirect. Poking around Chromium, looks like we do allow WebSockets to have HSTS and handle it somewhat specially here.)
https://fetch.spec.whatwg.org/#websocket-opening-handshake

Given all the other stuff going on in this draft, another option is to explicitly leave it undefined, but leave a little note that, if in an HTTP-like-but-not-quite-HTTP context where redirects aren't allowed or meaningful (e.g. WebSockets and proxies), that protocol should define an equivalent mechanism and use it. (Although, it's unclear whether this will ever get written down. For WebSockets, we probably need a rfc6455bis to realign it with reality. HTTP proxies probably needs a whole new document to define them better.)

ericorth: > that protocol should define an equivalent mechanism and use it

Will IETF let us get away with saying protocols should define something when relevant protocols already exist that don't define it?

bemasc: @davidben I think this text is close enough to that without getting too meta.

MikeBishop (approved):

PR #381, opened 2022-04-11
Address editorial comments from Ben Kaduk

### LINKED ISSUE #369: Editorial comments from Ben Kaduk
Section 1.2

   In ServiceMode, the SvcParams of the SVCB RR provide an extensible
   data model for describing alternative endpoints that are
   authoritative for the origin, along with parameters associated with
   each of these alternative endpoints.

(editorial) Since the term "origin" has a specific meaning in the HTTP
context, and this document covers both the generic SVCB and the
HTTP-specific HTTPS records, but this section is supposed to be the
"generic" one, I wonder if we can come up with a different term than
"origin" to use here.  Would "service" or a derived term be consistent
with our definitions?

Section 2.1

   The presentation format of the record is:

   Name TTL IN SVCB SvcPriority TargetName SvcParams

   The SVCB record is defined specifically within the Internet ("IN")
   Class ([RFC1035]).

Should we say something about how "Name" and "TTL" have the same meaning
as in RFC 1035?

   alpha-lc      = %x61-7A   ;  a-z
   SvcParamKey   = 1*63(alpha-lc / DIGIT / "-")
   SvcParam      = SvcParamKey ["=" SvcParamValue]
   SvcParamValue = char-string
   value         = *OCTET

I'd suggest putting some comments in the ABNF that <char-string> is
defined in Appendix A and that <value> and <char-string> are related as
described in the prose.

Using AliasMode maintains a separation of concerns: the owner of
   [foosvc.example.net](http://foosvc.example.net/) can add or remove ServiceMode SVCB records without
   requiring a corresponding change to [example.com](http://example.com/).  Note that if
   [foosvc.example.net](http://foosvc.example.net/) promises to always publish a SVCB record, this
   AliasMode record can be replaced by a CNAME, which would likely
   improve performance.

Section 2.4.2

Using AliasMode maintains a separation of concerns: the owner of
   [foosvc.example.net](http://foosvc.example.net/) can add or remove ServiceMode SVCB records without
   requiring a corresponding change to [example.com](http://example.com/).  Note that if
   [foosvc.example.net](http://foosvc.example.net/) promises to always publish a SVCB record, this
   AliasMode record can be replaced by a CNAME, which would likely
   improve performance.

So this would be "_8080._[foo.example.com](http://foo.example.com/) CNAME [foosvc.example.net](http://foosvc.example.net/)" rather
than "[example.com](http://example.com/) CNAME [foosvc.example.net](http://foosvc.example.net/)"?  I wonder if clarifying that
it's not "[example.com](http://example.com/) CNAME [foosvc.example.net](http://foosvc.example.net/)" is worthwhile.

Section 7.1.2

   Clients SHOULD NOT attempt connection to a service endpoint whose
   SVCB ALPN set does not contain any supported protocols.  To ensure
   consistency of behavior, clients MAY reject the entire SVCB RRSet and
   fall back to basic connection establishment if all of the RRs
   indicate "no-default-alpn", even if connection could have succeeded
   using a non-default alpn.

(editorial) Just to confirm: these two directives are only related in that
they cover client ALPN handling, and there is no expectation that the "MAY
reject" is preconditioned on "SVCB ALPN set does not contain any supported
protocols"?  I might consider splitting into two paragraphs, even though
my general preference is to avoid single-sentence paragraphs.

Section 9.3

   Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
   also publish an "ech" SvcParam for any alt-authorities.  [...]

Just to confirm I'm understanding this properly: this is saying that if an
origin publishes an HTTPS RR with an "ech" SvcParam (any single RR in the
RRset suffices to trigger the condition), then the origin should also
publish HTTPS records for any alt-authorities that it would send in an
Alt-Svc header field, and those other HTTPS records should all include an
"ech" SvcParam as well?  I'm a little confused at why the RFC 7838
"alt-authorities" keyword is used here without any specific reference to
Alt-Svc itself.  (There are only three instances of the string
"alt-authorit" in this document, two of which are in this paragraph; the
third is clearly marked as applying to the Alt-Svc usage.)

Section 9.4

   Clients MUST NOT use an HTTPS RR response unless the client supports
   TLS Server Name Indication (SNI) and indicates the origin name when
   negotiating TLS.  This supports the conservation of IP addresses.

I could sort-of see a misreading of this that says you have to put the
real origin in the SNI extension and not use ECH with a separate fronting
origin.  Maybe we want to say something about "uses SNI (possibly in
conjunction with ECH) to indicate the origin name"?

Section 11.2

   queries for TargetName in advance (see Section 5).  Unless otherwise
   specified, the convention is to set TargetName to the service name
   for an initial ServiceMode record, or to "." if it is reached via an
   alias.  For foo://[foo.example.com:8080](http://foo.example.com:8080/), this might look like:

I think it would be helpful to show an example of the "set TargetName to the
service name for an initial ServiceMode record" in addition to the alias
example that is shown.

Section 11.3.3

   Suppose that svc.example's default server pool supports HTTP/2, and
   it has deployed HTTP/3 on a new server pool with a different
   configuration.  This can be expressed in the following form:

   $ORIGIN svc.example. ; A hosting provider.
   pool  7200 IN HTTPS 1 h3pool alpn=h2,h3 ech="123..."

The prose suggests that perhaps the new pool is h3-only and doesn't support
HTTP/2, but the actual example lists both alpn values.  I guess it also omits
no-default-alpn, so the new pool actually supports all three http versions...

Section 11.3.4

   *  The A, AAAA, and HTTPS resolutions are independent lookups, so
      clients may observe and follow different CNAMEs to different CDNs.
      Clients may thus find a TargetName pointing to a name other than
      the one which returned along with the A and AAAA lookups and will
      need to do an additional resolution for them.  [...]

(editorial) I think something went awry near "the one which returned along
with", as it's not even clear whether there are separate queries vs
additional-section responses in the intended scenario.

Section 17.1

Why is [DNAME] classified as normative?  I only see one reference to it,
in a list of examples of possible ways that resolvers might follow
aliases, which does not seem like a "protocol feature" of this document.

I could also perhaps see an argument that references to [HSTS] are just by
way of analogy or indicating that we provide equivalent behavior, which
would not seem to require the reader to understand [HSTS] in order to
implement this specification.

RFC 6147 is referenced only as part of a "MUST NOT perform DNS64"
construction, which seems to indicate that it is *not* a normative
reference.

bemasc:
Resolves #369

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -126,7 +126,7 @@ Additional goals specific to HTTPS RRs and the HTTP use-cases include:
 * Support non-default TCP and UDP ports
 * Enable SRV-like benefits (e.g. apex delegation, as mentioned above) for HTTP,
   where SRV {{?SRV=RFC2782}} has not been widely adopted
-* Provide an HSTS-like indication {{!HSTS=RFC6797}} signaling that the "https"
+* Provide an HSTS-like indication {{?HSTS=RFC6797}} signaling that the "https"
   scheme should be used instead of "http" for this request (see {{hsts}}).
 
 ## Overview of the SVCB RR
@@ -163,7 +163,7 @@ records in the Additional Section to responses for a SVCB query.
 
 In ServiceMode, the SvcParams of the SVCB RR
 provide an extensible data model for describing alternative
-endpoints that are authoritative for the origin, along with
+endpoints that are authoritative for a service, along with
 parameters associated with each of these alternative endpoints.
 
 For HTTP use-cases, the HTTPS RR enables many of the benefits of Alt-Svc
@@ -233,12 +233,13 @@ and formats are defined in {{keys}}.
 
 ## Zone file presentation format {#presentation}
 
-The presentation format of the record is:
+The presentation format `<RDATA>` of the record ({{!RFC1035, Section 5.1}}) has
+the form:
 
-    Name TTL IN SVCB SvcPriority TargetName SvcParams
+    SvcPriority TargetName SvcParams
 
 The SVCB record is defined specifically within
-the Internet ("IN") Class ({{!RFC1035}}).
+the Internet ("IN") Class ({{!RFC1035, Section 3.2.4}}).
 
 SvcPriority is a number in the range 0-65535,
 TargetName is a `<domain-name>` ({{!RFC1035, Section 5.1}}),
@@ -255,7 +256,7 @@ In ABNF {{!RFC5234}},
     SvcParamKey   = 1*63(alpha-lc / DIGIT / "-")
     SvcParam      = SvcParamKey ["=" SvcParamValue]
     SvcParamValue = char-string
-    value         = *OCTET
+    value         = *OCTET ; Value before key-specific parsing
 
 The SvcParamValue is parsed using the
 character-string decoding algorithm ({{decoding}}), producing a `value`.
@@ -405,7 +406,8 @@ Using AliasMode maintains a separation of concerns: the owner of
 foosvc.example.net can add or remove ServiceMode SVCB records without
 requiring a corresponding change to example.com.  Note that if
 foosvc.example.net promises to always publish a SVCB record, this AliasMode
-record can be replaced by a CNAME, which would likely improve performance.
+record can be replaced by a CNAME at the same owner name, which would likely
+improve performance.
 
 AliasMode is especially useful for SVCB-compatible RR types that do not
 require an underscore prefix, such as the HTTPS RR type.  For example,
@@ -638,7 +640,7 @@ Additional section of the response as follows:
 In this procedure, "resolve" means the resolver's ordinary recursive
 resolution procedure, as if processing a query for that RRSet.
 This includes following any aliases that the resolver would ordinarily
-follow (e.g. CNAME, DNAME {{!DNAME=RFC6672}}).  Errors or anomalies in
+follow (e.g. CNAME, DNAME {{?DNAME=RFC6672}}).  Errors or anomalies in
 obtaining additional records MAY cause this process to terminate, but
 MUST NOT themselves cause the resolver to send a failure response.
 
@@ -882,7 +884,9 @@ interfere with ALPN protocol selection.  This procedure also ensures that
 each ProtocolNameList includes at least one protocol from the SVCB ALPN set.
 
 Clients SHOULD NOT attempt connection to a service endpoint whose SVCB
-ALPN set does not contain any supported protocols.  To ensure
+ALPN set does not contain any supported protocols.
+
+To ensure
 consistency of behavior, clients MAY reject the entire SVCB RRSet and fall
 back to basic connection establishment if all of the RRs indicate
 "no-default-alpn", even if connection could have succeeded using a
@@ -959,7 +963,7 @@ approach such as Happy Eyeballs {{!HappyEyeballsV2}}.
 When only "ipv4hint" is present, IPv6-only clients may synthesize
 IPv6 addresses as specified in {{!RFC7050}} or ignore the "ipv4hint" key and
 wait for AAAA resolution ({{client-behavior}}).  Recursive resolvers MUST NOT
-perform DNS64 ({{!RFC6147}}) on parameters within a SVCB record.
+perform DNS64 ({{?RFC6147}}) on parameters within a SVCB record.
 For best performance, server operators SHOULD include an "ipv6hint" parameter
 whenever they include an "ipv4hint" parameter.
 
@@ -1167,7 +1171,8 @@ the "ech" SvcParam is present ({{ech-client-behavior}}):
 * HTTP/3 to `example.com:8443`
 
 Origins that publish an "ech" SvcParam in their HTTPS record SHOULD
-also publish an "ech" SvcParam for any alt-authorities.  Otherwise,
+also publish an HTTPS record with the "ech" SvcParam for every
+alt-authority offered in its Alt-Svc Field Values.  Otherwise,
 clients might reveal the name of the server in an unencrypted ClientHello.
 Similar consistency considerations could apply to future SvcParamKeys, so
 alt-authorities SHOULD carry the same SvcParams as the origin unless
@@ -1182,7 +1187,7 @@ connection for this origin.
 
 Clients MUST NOT use an HTTPS RR response unless the
 client supports TLS Server Name Indication (SNI) and
-indicates the origin name when negotiating TLS.
+indicates the origin name in the TLS ClientHello (which may be encrypted).
 This supports the conservation of IP addresses.
 
 Note that the TLS SNI (and also the HTTP "Host" or ":authority") will indicate
@@ -1303,16 +1308,18 @@ choose TargetName according to a predictable convention that is known
 to the client, so that clients can issue A and/or AAAA queries for TargetName
 in advance (see {{optimizations}}).  Unless otherwise specified, the
 convention is to set TargetName to the service name for an initial
-ServiceMode record, or to "." if it is reached via an alias.  For
-foo://foo.example.com:8080, this might look like:
+ServiceMode record, or to "." if it is reached via an alias.
 
     $ORIGIN example.com. ; Origin
     foo                  3600 IN CNAME foosvc.example.net.
     _8080._foo.foo       3600 IN CNAME foosvc.example.net.
+    bar                   300 IN AAAA 2001:db8::2
+    _9090._bar.bar       3600 IN SVCB 1 bar key65444=...
 
     $ORIGIN example.net. ; Service provider zone
     foosvc               3600 IN SVCB 1 . key65333=...
     foosvc                300 IN AAAA 2001:db8::1
+{: title="foo://foo.example.com:8080 is delegated to a example.net, but bar://bar.example.com:9090 is served locally.}
 
 Domain owners SHOULD avoid using a TargetName that is below a DNAME, as
 this is likely unnecessary and makes responses slower and larger.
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -959,7 +963,7 @@ approach such as Happy Eyeballs {{!HappyEyeballsV2}}.
 When only "ipv4hint" is present, IPv6-only clients may synthesize
 IPv6 addresses as specified in {{!RFC7050}} or ignore the "ipv4hint" key and
 wait for AAAA resolution ({{client-behavior}}).  Recursive resolvers MUST NOT
-perform DNS64 ({{!RFC6147}}) on parameters within a SVCB record.
+perform DNS64 ({{?RFC6147}}) on parameters within a SVCB record.
```
MikeBishop, line 966: It feels odd to use an informational reference in specifying a normative requirement.  In saying you MUST NOT take some action, we inherently make knowing what not to do necessary to complying.  (In other words, I disagree with @kaduk here, and he accepts my disagreement.)  Vote of the authors?

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1182,7 +1187,7 @@ connection for this origin.
 
 Clients MUST NOT use an HTTPS RR response unless the
 client supports TLS Server Name Indication (SNI) and
-indicates the origin name when negotiating TLS.
+indicates the origin name in the TLS ClientHello (which may be encrypted).
```
MikeBishop, line 1190:
```suggestion
indicates the origin name in the TLS ClientHello (which might be encrypted).
```

MikeBishop (approved):

draft-ietf-dnsop-svcb-https.md
bemasc: My intuition was similar to @MikeBishop's, but I view this as a process question on which @kaduk may be authoritative.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -416,7 +417,8 @@ Using AliasMode maintains a separation of concerns: the owner of
 foosvc.example.net can add or remove ServiceMode SVCB records without
 requiring a corresponding change to example.com.  Note that if
 foosvc.example.net promises to always publish a SVCB record, this AliasMode
```
enygren, line 419:
```suggestion
foosvc.example.net promises to always publish a SVCB record and has no other
records on the name, this AliasMode
```

(since this replacement with a CNAME can't happen if the ownername has other records such as NS)

draft-ietf-dnsop-svcb-https.md
enygren: Would "SHOULD NOT" be better, with the warning that doing so would break any DNSSEC signatures? 


In General Requirements we say:

> Recursive resolvers MUST be able to convey SVCB records with unrecognized SvcParamKeys, and MAY treat the entire SvcParams portion of the record as opaque, even if the contents are invalid. Alternatively, recursive resolvers MAY report an error such as SERVFAIL to avoid returning a SvcParamValue that is invalid according to the SvcParam's specification. For complex value types whose interpretation might differ between implementations or have additional future allowed values added (e.g. URIs or "alpn"), resolvers SHOULD limit validation to specified constraints.

but we don't say anything about modifying parameters.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1330,16 +1335,18 @@ choose TargetName according to a predictable convention that is known
...
     $ORIGIN example.com. ; Origin
     foo                  3600 IN CNAME foosvc.example.net.
     _8080._foo.foo       3600 IN CNAME foosvc.example.net.
+    bar                   300 IN AAAA 2001:db8::2
+    _9090._bar.bar       3600 IN SVCB 1 bar key65444=...
 
     $ORIGIN example.net. ; Service provider zone
     foosvc               3600 IN SVCB 1 . key65333=...
     foosvc                300 IN AAAA 2001:db8::1
+{: title="foo://foo.example.com:8080 is delegated to a example.net, but bar://bar.example.com:9090 is served locally.}
```
enygren, line 1349:
```suggestion
{: title="foo://foo.example.com:8080 is delegated to foosvc.example.net, but bar://bar.example.com:9090 is served locally.}
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -1498,11 +1504,11 @@ introduces a number of complexities highlighted by this example:
...
 
-* The A, AAAA, and HTTPS resolutions are independent lookups, so clients may
+* The A, AAAA, and HTTPS resolutions are independent lookups, so resolvers may
   observe and follow different CNAMEs to different CDNs.
-  Clients may thus find a TargetName pointing to a name
-  other than the one which returned along with the A and AAAA lookups
-  and will need to do an additional resolution for them.
+  Clients may thus find that the A and AAAA responses do not correspond to the
+  TargetName in the HTTPS response, and will need to perform additional
+  queries to retrieve its IP addresses.
```
enygren, line 1511:
```suggestion
  queries to retrieve the correct IP addresses.
```

(or "proper")

enygren (changes requested):
Looks generally good but some comments above.

draft-ietf-dnsop-svcb-https.md
bemasc: There is no requirement that `foosvc.example.com` not have other records, even NS.  It can certainly be a zone apex.

I think you mean "and `_8080._foo.example.com` has no other records".  Any name where this is not true would strike me as pretty bizarre, so I don't see why this needs to be mentioned, but if you think it's necessary we can add it.

draft-ietf-dnsop-svcb-https.md
bemasc: I don't think MUST vs. SHOULD is relevant to the question of whether the reference is normative.

I do think this is a MUST.  Violations of this rule will result in compatibility failures.

draft-ietf-dnsop-svcb-https.md
enygren: I misread this. Ignore my comment.  :)

draft-ietf-dnsop-svcb-https.md
enygren:
```suggestion
perform DNS64 ({{?RFC6147}}) on parameters within a SVCB record
as this would result in DNSSEC signature and other compatibility failures.
```

?  (and leave as a MUST?)
But this is odd in-general as it is saying "don't get creative here and try
and extrapolate from something informational!"

enygren (approved):
I'm fine with this, unless we want to do more with the DNS64 item.

draft-ietf-dnsop-svcb-https.md
bemasc: I'm fine with this text change, but it sounds like it doesn't quite capture your intent.  Maybe the solution is to drop the normative language and say something like

> Note that any DNS64 AAAA record synthesis ({{Section 5.1.7 of ?RFC6147}}) applies only to AAAA queries, and does not extend to IP hints conveyed in SVCB records.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -978,7 +982,7 @@ approach such as Happy Eyeballs {{!HappyEyeballsV2}}.
 When only "ipv4hint" is present, IPv6-only clients may synthesize
 IPv6 addresses as specified in {{!RFC7050}} or ignore the "ipv4hint" key and
 wait for AAAA resolution ({{client-behavior}}).  Recursive resolvers MUST NOT
```
enygren, line 984: Another potential phrasing:

```suggestion
wait for AAAA resolution ({{client-behavior}}).  Note that any DNS64 AAAA record synthesis ({{Section 5.1.7 of ?RFC6147}}) applies only to AAAA queries, and trying to extrapolate it to synthesize ipv6hint parameters would likely result in DNSSEC and other compatibility issues.
```

bemasc: After thinking about this some more, I realized that my claim that this "applies to only AAAA queries" is not really correct, because we also want AAAA synthesis to apply to the Additional records returned from a SVCB query.  That would make this a positive, and therefore unambiguously normative, reference.

I've moved this discussion into a new subsection under Resolver Behavior.  Please review!

draft-ietf-dnsop-svcb-https.md
kaduk: > I'm fine with this text change, but it sounds like it doesn't quite capture your intent. Maybe the solution is to drop the normative language and say something like
> 
> > Note that any DNS64 AAAA record synthesis ({{Section 5.1.7 of ?RFC6147}}) applies only to AAAA queries, and does not extend to IP hints conveyed in SVCB records.

This seems like a pretty good approach to me.  The specific wording might get tweaked further (the bit around "applies only to" catches my eye as something I might want to edit, but I can't put my finger on exactly why at the moment), but it seems okay to frame this statement as a factual clarification for the scope of RFC 6147 rather than a new normative statement with normative keywords.

kaduk: Oops, but there were more comments elsewhere that render this proposal obsolete anyway.  I should have read on before commenting, sorry!

draft-ietf-dnsop-svcb-https.md
kaduk: I like it, thanks for the rewrite.

draft-ietf-dnsop-svcb-https.md
```diff
@@ -664,6 +666,19 @@ to implement to mitigate loops.
 
 See {{incomplete-response}} for possible optimizations of this procedure.
 
+### DNS64
+
+DNS64 resolvers synthesize responses to AAAA queries for names that only
+have an A record ({{Section 5.1.7 of !RFC6147}}).  SVCB-aware DNS64
+resolvers SHOULD apply the same synthesis logic when resolving AAAA
+records for the TargetName (Step 2 in {{recursive-behavior}}), and MAY
+omit the Additional A records.
```
enygren, line 675:
```suggestion
resolvers SHOULD apply the same synthesis logic when resolving AAAA
records for the TargetName for inclusion as Additionals (Step 2 in
{{recursive-behavior}}), and MAY omit the Additional A records.
```

draft-ietf-dnsop-svcb-https.md
```diff
@@ -664,6 +666,19 @@ to implement to mitigate loops.
...
+
+DNS64 resolvers MUST NOT extrapolate the AAAA synthesis logic to the IP
+hints in the SvcParams ({{svcparamkeys-iphints}}).  Modifying the IP hints
+would break DNSSEC validation and would not improve performance when the
+above recommendation is implemented.
```
enygren, line 680:
```suggestion
would break DNSSEC validation for the SVCB record and would not improve 
performance when the above recommendation is implemented.
```

(DNSSEC validation for the AAAA records is broken by DNS64 already, but I don't think we need to mention that?)

enygren (changes requested):
Minor nits, but looks generally fine

enygren (approved):

PR #396, opened 2022-07-25
Fix the char-string ABNF.

### LINKED ISSUE #395: Character String ABNF too strict
First, apologies for bringing this up so late in the process, but I only noticed the issue now while I happened to work on a zone parser.

The ABNF specified in appendix A for `char-string` conflicts with the informal definition given in RFC 1035. Specifically, the ABNF requires all special characters, i.e., `"`, `;`, `(`, `)`, and `\`, to be escaped inside a quoted character string. Conversely, RFC 1035 only requires the double quote to be escaped.

This stricter definition actually is a problem in practice: DKIM and DMARC use semicolons as field separators. Both the examples in their RFCs as well as all ocurrences in wild I have seen use quoted character strings and don’t escape the semicolon.

For instance, RFC 7489 contains the following:

```
     _dmarc  IN TXT ( "v=DMARC1; p=none; "
                      "rua=mailto:dmarc-feedback@example.com" )
```

This will violate the ABNF grammar because of the unescaped semicolons in the first quoted character string.

I think this can be fixed by defining a new rule for characters allowed unescaped in quoted character strings and then redefining `quoted` through it. E.g.:

```
; VCHAR minus DQUOTE and "\"
quoted-char = WSP / "!" / %x23-5B / %x5D-7E
quoted      = DQUOTE *( quoted-char / escaped ) DQUOTE
```

This also fixes a somewhat inelegant nested repetition in the draft’s ABNF where `quoted` contains a repetition of `contiguous` which in turn is a repetition in its entirety.

partim:
This PR changes the char-string ABNF to allow all printable characters other than double quote and backslash unescaped in the quoted variant of char-string.

Fixes #395.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -1745,7 +1745,9 @@ Here we summarize the allowed input to that algorithm, using ABNF:
                   "2" ( ( %x30-34 DIGIT ) / ( "5" %x30-35 ) )
     escaped     = "\" ( non-digit / dec-octet )
     contiguous  = 1*( non-special / escaped )
-    quoted      = DQUOTE *( contiguous / ( ["\"] WSP ) ) DQUOTE
+    ; VCHAR minus DQUOTE and "\"
+    quoted-char = WSP / "!" / %x23-5B / %x5D-7E
+    quoted      = DQUOTE *( quoted-char / escaped ) DQUOTE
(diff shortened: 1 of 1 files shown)


moonshiner:
I spoke with Martin on this and working through this I concur with the PR
(and thanks for listening to me go on and on and on about ABNF)

bemasc (approved):

PR #399, opened 2022-09-08
Use fina $QNAME only when an AliasMode target

hs-viktor:
Per the title, use of $QNAME as an additional target was only ever intended when some positive number of AliasMode records were processed.


bemasc:
@wkumari Can you confirm that you are recommending that we merge this PR?

wkumari:
Yup, that's the bunny!

https://mailarchive.ietf.org/arch/msg/dnsop/LhEufm0a8IjoXZ_ck0b9RTSkDTc/ for historical record.

enygren (approved):

PR #400, opened 2022-09-22
Additional text on non-hostname values of TargetName

bemasc:
This change corrects a mistake in the introduction (asserting that TargetName is a hostname), and notes a potential operational issue that can prevent zones from choosing a TargetName that is not a hostname.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -37,6 +37,10 @@ informative:
     title: "Fetch Living Standard"
     date: May 2020
     target: "https://fetch.spec.whatwg.org/"
+  CHECK-NAMES:
+    title: "BIND 9 Configuration Reference: check-names"
+    date: September 2022
+    target: "https://bind9.readthedocs.io/en/v9_18_6/reference.html#namedconf-statement-check-names"
 --- abstract
 
 This document specifies the "SVCB" and "HTTPS" DNS resource record (RR)
@@ -193,7 +197,7 @@ hostname as the `host`.
 * The "service name" is the `host` portion of the authority.
 * The "authority endpoint" is the authority's hostname and a port number implied
   by the scheme or specified in the URI.
-* An "alternative endpoint" is a hostname, port number, and other associated
+* An "alternative endpoint" is a domain name, port number, and other associated
   instructions to the client on how to reach an instance of service.
 
 Additional DNS terminology intends to be consistent
@@ -498,6 +502,12 @@ is the effective TargetName:
     svc2.example.net. 300   IN A     192.0.2.2
     svc2.example.net. 300   IN AAAA  2001:db8::2
 
+Some authoritative DNS implementations restrict the owner names of A and AAAA
+records.  For example, BIND 9.18 requires these owner names to be hostnames
+when the `check-names` option is enabled {{CHECK-NAMES}}.  The ServiceMode
+TargetName is required to point at A or AAAA records, so any such restrictions
+may impact the choice of TargetName, and may preclude use of the value "." if
+the ServiceMode record's owner name is not permitted for A and AAAA records.
 
 
 # Client behavior {#client-behavior}


draft-ietf-dnsop-svcb-https.md
```diff
@@ -498,6 +502,12 @@ is the effective TargetName:
     svc2.example.net. 300   IN A     192.0.2.2
     svc2.example.net. 300   IN AAAA  2001:db8::2
 
+Some authoritative DNS implementations restrict the owner names of A and AAAA
+records.  For example, BIND 9.18 requires these owner names to be hostnames
+when the `check-names` option is enabled {{CHECK-NAMES}}.  The ServiceMode
+TargetName is required to point at A or AAAA records, so any such restrictions
+may impact the choice of TargetName, and may preclude use of the value "." if
+the ServiceMode record's owner name is not permitted for A and AAAA records.
```
enygren, line 510: Do we want to clarify as "not permitted by the implementation for A and AAAA records"?
(It is fairly obvious already, but it seems like we want to be crystal clear that the implementation is what is doing the permitting / not permitting.)

draft-ietf-dnsop-svcb-https.md
```diff
@@ -193,7 +197,7 @@ hostname as the `host`.
 * The "service name" is the `host` portion of the authority.
 * The "authority endpoint" is the authority's hostname and a port number implied
   by the scheme or specified in the URI.
-* An "alternative endpoint" is a hostname, port number, and other associated
+* An "alternative endpoint" is a domain name, port number, and other associated
```
wkumari, line 200: Erm. This is a larger change than we agreed to in https://mailarchive.ietf.org/arch/msg/dnsop/LhEufm0a8IjoXZ_ck0b9RTSkDTc/

draft-ietf-dnsop-svcb-https.md
wkumari: I think we should stick with Erik's originally proposed: 
"Some
implementations may not allow A or AAAA records on names starting with an
underscore due to various interpretations of RFCs. This could be an
operational issue when the TargetName contains an attrleaf label, as well
as using an TargetName of "." when the owner name contains an attrleaf
label."

While the additional text is useful background (and would have been good it we added it pre-WGLC/IETF LC) I think we should restrict changes to only those clearly agreed to in the thread https://mailarchive.ietf.org/arch/msg/dnsop/LhEufm0a8IjoXZ_ck0b9RTSkDTc/

wkumari (changes requested):
These changes are larger than what was agreed to on-list. They are probably not wrong, but for the sake of transparency/minimizing-post-approval changes we should err on the side of caution.

draft-ietf-dnsop-svcb-https.md
enygren: This is fixing this to be consistent with elsewhere in the document (so an error/errata here), so is arguably not a technical change?  (Should we ask the list about this in-particular?)
Elsewhere we have in the draft we are clear that TargetName is a domain name:

> 2. TargetName: The *domain name* of either the alias target (for
>   AliasMode) or the alternative endpoint (for ServiceMode).

> TargetName is a `<domain-name>` ({{!RFC1035, Section 5.1}}),

> In AliasMode, the TargetName will be the name of a domain that resolves to SVCB,

draft-ietf-dnsop-svcb-https.md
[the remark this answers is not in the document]
bemasc: Nit: Quotation marks, "some versions BIND".

I think "some versions of BIND" should be "BIND in some configurations", since this is an optional feature.

"may not" no longer seems appropriate when we have explicitly identified a configuration where this applies.  "do not" seems better.

"due to various interpretations of RFCs" seems too informal to appear in an RFC.  Perhaps we can remove it.

"could be an operational issue" seems both too informal and extremely vague.  I'm not even sure it's true.  The operator MUST publish the A/AAAA records before publishing the TargetName, so this issue will be addressed when choosing the owner name for those records, and the TargetName will be wherever they were ultimately placed.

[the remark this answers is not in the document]
wkumari: No.

This document was already approved and sent to the RFC Editor; I clawed it back because there were concerns expressed that there was a significant error, and, after WG discussions we agreed to the patch in #399.

Because it was purely operational advice there seemed to be general agreement that Erik's initial text was OK to add (and I got IESG agreement that we could include it).
I understand the desire to massage and improve this text, fix the nits and loose threads, etc - but it's too late for that.

So, please incorporate only the original, agreed to text and post a new revision, so that I can send it to the IESG for (re)approval, and return it to the RFC Editor queue. 
Sorry to be inflexible (and grumpy), but this is all post WGLC, IETF LC, IESG eval and IESG approval - now is not the time to be making additional changes.

Please see: https://mailarchive.ietf.org/arch/msg/dnsop/LhEufm0a8IjoXZ_ck0b9RTSkDTc/

bemasc:
Per @wkumari it seems that adjusted text on this topic is not possible at this point in the process.

PR #403, opened 2022-10-04
Minimal correction ("hostname" -> "domain name") in Intro

bemasc:


enygren (approved):
If Warren and dnsop are ok with this.

wkumari:
Once again, no. 

You have placed me in a really uncomfortable situation - the document had already been approved, and I clawed it back from the RFC Editor to address (what was stated as) a significant issue. The plan was to fix only that issue, do so quickly, and return it to the RFC Editor queue.

We worked through the issue and agreed to the changes in #399 . We also agreed (which I now regret, as perhaps it is the source of this belief that we can make other changes) to adding the operational note"`Some implementations may not allow A or AAAA records on names starting with an underscore due to various interpretations of RFCs. This could be an operational issue when the TargetName contains an attrleaf label, as well as using an TargetName of "." when the owner name contains an attrleaf label.`" (see: https://mailarchive.ietf.org/arch/msg/dnsop/LhEufm0a8IjoXZ_ck0b9RTSkDTc/ - Sept 12th (23 days ago)).

**I'm unclear how else to state this**: These are the **only** approved changes; please cut and post a new version with only those changes, and I'll take it to the rest of the IESG for approval, and then return it to the RFC Editor queue. 
If you cannot do that **soon**, I will have to return it to the RFC Editor unchanged, or possibly un-approve it and return it to the WG to restart the process.

bemasc:
Apologies, I evidently misunderstood which changes were approved here.  This change was discussed in the DNSOP thread on [Aug. 29th](https://mailarchive.ietf.org/arch/msg/dnsop/jGnsrrwJgLY8jAAhXkR-nlxOo54/), and [Sept. 29th](https://mailarchive.ietf.org/arch/msg/dnsop/zRwibFWjJggpV9BlPsRZicUHb90/).  It wasn't clear to me whether it had been reviewed by the IESG, or whether it was sufficiently trivial to include given the DNSOP discussion.

Evidently it had not been reviewed by the IESG, and is not judged to be allowable on some other basis.

PR #404, opened 2022-10-05
Some impl may not allow A/AAAA records on names starting w/ underscore

enygren:
…ith an underscore

As discussed in dnsop and approved for inclusion, note that some implementations may not allow A/AAAA records on names starting with an underscore.  See: https://mailarchive.ietf.org/arch/msg/dnsop/LhEufm0a8IjoXZ_ck0b9RTSkDTc/


wkumari:
Yes, (and thank you very much)

PR #407, opened 2022-10-20
Update draft-ietf-dnsop-svcb-https.md

systemcrash:
When discussing implications of ServiceMode for SVCB RRs, it is confusing to use an example with HTTPS RRs without SVCB RRs. Change to SVCB RRs.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -492,9 +492,9 @@ SHALL use the response's synthesized owner name as the effective TargetName.
 For example, in the following example "svc2.example.net"
 is the effective TargetName:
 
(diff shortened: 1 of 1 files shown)


bemasc:
SVCB RRs are normally used with underscore prefix owner names, so this example would probably need more changes.  However, the text of this draft is currently frozen, as it has passed to the RFC Editor, so we're not accepting new PRs at this time.

systemcrash:
Really? The example in 2.3 indicates it's OK.  
```
   svc4.example.net.  7200  IN SVCB 3 svc4.example.net. (
       alpn="bar" port="8004" ech="..." )
```

bemasc:
Yes, it is allowed, but prefixes are generally required for the "initial" owner name (i.e. the initial QNAME).  Further down the alias chain, they are not required.

PR #412, opened 2023-03-07
Move ECH-related requirements into a separate draft

bemasc:


tfpauly (approved):

draft-sbn-dnsop-svcb-ech.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Interaction with HTTP Alt-Svc
+
+HTTP clients that implement both HTTP Alt-Svc {{?RFC7838}} and the HTTPS record type {{!SVCB}} can use them together, provided that they only perform connection attempts that are "consistent" with both sets of parameters ({{Section 9.3 of !SVCB}}).  At the time of writing, there is no defined parameter related to ECH for Alt-Svc.  Accordingly, a connection attempt that uses ECH is considered "consistent" with an Alt-Svc Field Value that does not mention ECH.
+
+Origins that publish an "ech" SvcParam in their HTTPS record SHOULD also publish an HTTPS record with the "ech" SvcParam for every alt-authority offered in its Alt-Svc Field Values.  Otherwise, clients might reveal the name of the server in an unencrypted ClientHello to an alt-authority.
+
+If all HTTPS records for an alt-authority contain "ech" SvcParams, the client MUST adopt SVCB-reliant behavior (as in {{disabling-fallback}}) for that RRSet.  This precludes the use of certain connections that Alt-Svc would otherwise allow, as discussed in {{Section 9.3 of !SVCB}}.
+
```
enygren, line 78: It would be good to replicate some of the "Multi CDN" section from the SVCB draft, or at least the examples.  Which could just be dropping most or all of Multi-CDN {#multicdn}" back in here.  It's non-normative so repeating it might help provide clarity. There are some clarifications there that are important to ECH and which might otherwise be lost in subtlety.

enygren (commented):
The edits to the existing draft look good to me.  I found two other things 
we may want to also change and opened https://github.com/MikeBishop/dns-alt-svc/pull/413
(as a PR against the branch with the PR).

I haven't comprehensively reviewed sbn-dnsop-svcb-ech but it looks fine as a start.  I might re-add the MultiCDN example to it as well.

draft-sbn-dnsop-svcb-ech.md
bemasc: That seems fine to me, although duplicating the whole example in both drafts seems excessive.  Maybe we can shrink both copies to focus on the relevant aspects.

Perhaps we can land this PR and you can add/adjust the example before we publish -00?

draft-sbn-dnsop-svcb-ech.md
```diff
@@ -0,0 +1,92 @@
...
+## Security Considerations
+
+A SVCB RRSet containing some RRs with "ech" and some without is vulnerable to a downgrade attack.  This configuration is NOT RECOMMENDED. Zone owners who do use such a mixed configuration SHOULD mark the RRs with "ech" as more preferred (i.e. lower SvcPriority value) than those without, in order to maximize the likelihood that ECH will be used in the
```
chris-wood, line 81:
```suggestion
A SVCB RRSet containing some RRs with "ech" and some without is vulnerable to a downgrade attack.  In particular, an attacker can remove RRs without the "ech" parameter to hide the server's ECH capability and configuration from the client.  As such, this configuration is NOT RECOMMENDED. Zone owners who do use such a mixed configuration SHOULD mark the RRs with "ech" as more preferred (i.e. lower SvcPriority value) than those without, in order to maximize the likelihood that ECH will be used in the
```

chris-wood (approved):
Thanks, @bemasc! This looks great. I strongly suggest moving the new document to the TLS WG since the overwhelming majority of content in that document pertains to TLS and ECH. Everything else is a fine first draft. Nice work getting this together so quickly.

draft-sbn-dnsop-svcb-ech.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Bootstrapping Encrypted ClientHello with DNS Service Bindings
+abbrev: ECH in SVCB
+docname: draft-sbn-dnsop-svcb-ech-latest
```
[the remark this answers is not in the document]
tfpauly: Yeah, I agree that targeting this at the TLS WG would make sense!

sftcd:
On 07/03/2023 23:28, Tommy Pauly wrote:
> Yeah, I agree that targeting this at the TLS WG would make sense!

I agree that the TLS WG is the right place to handle
this. Could even be merged with other bits'n'pieces
related to ECH possibly.

S.

draft-sbn-dnsop-svcb-ech.md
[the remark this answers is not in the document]
bemasc: OK, done!

chris-wood (approved):
Looks great -- thanks!

moonshiner:
May I just say this is great work all and I'm very happy, and as a chair i'm hoping a new version will make it in before the cutoff so I can do the process work ASAP.

enygren (approved):
Both seem good enough to publish now.  We can refine/improve the tls-svcb-ech draft separately after we get a -00 of it out (eg, to add in a multicdn example and some other refinements), but hopefully ietf-dnsop-svcb-https can go straight to last calls?

moonshiner:
My Plan is immediate WGLC of one week; and have all prep work needed to send it right along back to the IESG.

moonshiner:
Also the ECH document in the TLS working group makes sense. DNSOP has already blessed the work.

MikeBishop (approved):
Yes, this looks solid, though I'd probably do some line-wrapping in the new draft.  Thanks for moving this forward!

bemasc:
> Yes, this looks solid, though I'd probably do some line-wrapping in the new draft. Thanks for moving this forward!

I don't believe in line wrapping :-)

PR #413, opened 2023-03-07
Updates to ECH split-out

enygren:
Found two additional places we probably want to make edits. Also added a change log entry.

### DIFF
--- a/draft-ietf-dnsop-svcb-https.md
+++ b/draft-ietf-dnsop-svcb-https.md
@@ -44,7 +44,8 @@ types to facilitate the lookup of information needed to make connections
 to network services, such as for HTTP origins.  SVCB records
 allow a service to be provided from multiple alternative endpoints,
 each with associated parameters (such as transport protocol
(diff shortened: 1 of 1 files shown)


draft-ietf-dnsop-svcb-https.md
```diff
@@ -44,7 +44,7 @@ types to facilitate the lookup of information needed to make connections
 to network services, such as for HTTP origins.  SVCB records
 allow a service to be provided from multiple alternative endpoints,
 each with associated parameters (such as transport protocol
-configuration and keys for encrypting the TLS ClientHello).  They also
+configuration and future uses such as keys for encrypting the TLS ClientHello).  They also
```
bemasc, line 47:
```suggestion
configuration), and are extensible to support future uses
such as keys for encrypting the TLS ClientHello).  They also
```

bemasc (approved):
Thanks for the close review.  I've added a suggestion.

bemasc (approved):

PR #416, opened 2023-03-19
updated two references

moonshiner:
Updated these two references

** Obsolete normative reference: RFC 7231 (Obsoleted by RFC 9110)

(changed section ref)

== Outdated reference: draft-ietf-quic-http has been published as RFC 9114

The draft-ietf-httpbis-semantics reference is in the abstract and I think it needs to be removed eventually,
but I also think you may want to include it elsewhere.


MikeBishop:
I would suggest adding this to the normative references at the top, so the display anchors aren't lost:

```
  RFC9114:
    display: HTTP/3
  RFC9110:
    display: HTTP
```

moonshiner:
Also Warren says don't worry about the outdated references until he reaches the RFC editor again.

